PR #2, opened 2022-04-07
Migrate to Dalvik patches

### LINKED ISSUE #1: Patch descriptions
Patches should have metadata like description or friendly names for the manager

### LINKED ISSUE #2430: Patch descriptions
Patches should have metadata like description or friendly names for the manager

oSumAtrIX:
This PR is an addition to https://github.com/ReVancedTeam/revanced-patcher/pull/16 and migrates the patches to Dalvik.
Closes ReVanced/revanced-patches-template#2430.


oSumAtrIX:
To make the releases build, PR https://github.com/ReVancedTeam/revanced-patcher/pull/16 has to be merged.

src/main/kotlin/app/revanced/patches/interaction/EnableSeekbarTapping.kt
```diff
@@ -39,45 +37,21 @@ class EnableSeekbarTapping : Patch("enable-seekbar-tapping") {
...
-                        pMethod.definingClass,
-                        pMethod.name,
-                        setOf("I"),
-                        "V"
-                    )
-                )
-            )
+            """
+               invoke-virtual { v12, v2 }, ${oMethod.definingClass}->${oMethod.name}(I)V
+               invoke-virtual { v12, v2 }, ${pMethod.definingClass}->${pMethod.name}(I)V
```
Lucaskyy, line 53: That `v12` looks a little arbitrary and error-prone. Could you use an offset instead? Maybe we should add helper methods to get the register used at a instruction to get an offset that way?

src/main/kotlin/app/revanced/patches/ad/VideoAdsPatch.kt
```diff
@@ -0,0 +1,103 @@
...
+                methodMetadata = MethodMetadata(
+                    definingClass = "zai",
+                    name = "<init>",
+                ),
+                patternScanMethod = PatternScanMethod.Fuzzy(2),// FIXME: Test this threshold and find the best value.
+                compatiblePackages = compatiblePackages,
+                description = """Signature for the constructor of some class.
+                This signature is being used to find another method in the parent class
+                and was discovered in the YouTube version v17.03.38""",
+                version = "0.0.1"
```
Lucaskyy, line 38: This is not what I initially meant with the `version` parameter for `MethodSignature`.
Even though this is fine, we now don't have any context for `methodMetadata`. How should we know which version that metadata is made for?

src/main/kotlin/app/revanced/patches/ad/VideoAdsPatch.kt
```diff
@@ -0,0 +1,103 @@
...
+    signatures = listOf(
+        MethodSignature(
+            methodSignatureMetadata = MethodSignatureMetadata(
+                name = "show-video-ads-constructor",
+                methodMetadata = MethodMetadata(
+                    definingClass = "zai",
+                    name = "<init>",
+                ),
+                patternScanMethod = PatternScanMethod.Fuzzy(2),// FIXME: Test this threshold and find the best value.
+                compatiblePackages = compatiblePackages,
```
Lucaskyy, line 34: What is this used for? It doesn't seem to have a use right now.

src/main/kotlin/app/revanced/patches/ad/VideoAdsPatch.kt
```diff
@@ -0,0 +1,103 @@
...
+                description = """Signature for the constructor of some class.
+                This signature is being used to find another method in the parent class
+                and was discovered in the YouTube version v17.03.38""",
```
Lucaskyy, line 37: Make sure to trim the indents (tabs) everywhere you do this.
```suggestion
                and was discovered in the YouTube version v17.03.38""".trimIndent(),
```

src/main/kotlin/app/revanced/patches/ad/VideoAdsPatch.kt
```diff
@@ -0,0 +1,103 @@
...
+import app.revanced.patcher.patch.*
+import app.revanced.patcher.signature.MethodMetadata
+import app.revanced.patcher.signature.MethodSignature
+import app.revanced.patcher.signature.MethodSignatureMetadata
+import app.revanced.patcher.signature.PatternScanMethod
+import app.revanced.patcher.smali.asInstructions
+import org.jf.dexlib2.AccessFlags
+import org.jf.dexlib2.Opcode
+
+private val compatiblePackages = arrayOf("com.google.android.youtube")
```
Lucaskyy, line 15: Maybe put this is in a `Shared.kt` instead and make it `internal`?

Sculas (changes requested):
Most of these apply to all patches, but I only reviewed `VideoAdsPatch` because they are all the same.

src/main/kotlin/app/revanced/patches/ad/VideoAdsPatch.kt
oSumAtrIX: The `version` parameter is just metadata for the user. So later it will be displayed in the manager and should be used to version signatures or patches. This way someone can down or upgrade patches and signatures. The `description` parameter or an additional one can be used to tell, which base version a signature and thus `MethodMetadata` has been generated for.

src/main/kotlin/app/revanced/patches/ad/VideoAdsPatch.kt
oSumAtrIX: The `compatible packages` parameter will be useful to filter for patches. There might be global patches, which edit a package's icon or resources, or generic patches Microg, which can be applied for multiple packages like YouTube Music or YouTube. A patch might also be compatible with one package only. The manager can use this metadata, to apply correct patches for each package accordingly.

src/main/kotlin/app/revanced/patches/ad/VideoAdsPatch.kt
oSumAtrIX: As suggested in https://github.com/ReVancedTeam/revanced-patches/pull/2#discussion_r850279985, this parameter should be unique to each patch. It will be used to determine which patch should be allowed for what package.

src/main/kotlin/app/revanced/patches/interaction/EnableSeekbarTapping.kt
Lucaskyy: @oSumAtrIX ?

oSumAtrIX: IIRC two instructions afterwards, there are instructions referring to the same register, we can use them

oSumAtrIX: Fixed in 0924ca2ad30ae865dcc0fd484cb0da517e827352

PR #6, opened 2022-04-25
Enable Background & Download Settings

dylangerdaly:
Enables Background and Download within Settings

### DIFF
--- a/src/main/kotlin/app/revanced/patches/layout/ForceSettingsPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/layout/ForceSettingsPatch.kt
@@ -1,0 +1,68 @@
+package app.revanced.patches.layout
+
+import app.revanced.patcher.PatcherData
+import app.revanced.patcher.extensions.addInstructions
+import app.revanced.patcher.extensions.or
+import app.revanced.patcher.patch.*
+import app.revanced.patcher.signature.MethodMetadata
+import app.revanced.patcher.signature.MethodSignature
+import app.revanced.patcher.signature.MethodSignatureMetadata
+import app.revanced.patcher.signature.PatternScanMethod
+import app.revanced.patcher.smali.toInstructions
+import org.jf.dexlib2.AccessFlags
+import org.jf.dexlib2.Opcode
+
+private val compatiblePackages = listOf(
+    PackageMetadata(
+        "com.google.android.youtube",
+        listOf("17.14.35")
+    )
+)
+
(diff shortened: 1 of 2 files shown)


src/main/kotlin/app/revanced/patches/layout/ForceSettingsPatch.kt
```diff
@@ -0,0 +1,68 @@
...
+import org.jf.dexlib2.Opcode
+
+private val compatiblePackages = listOf(
+    PackageMetadata(
+        "com.google.android.youtube",
+        listOf("17.14.35")
+    )
+)
+
+class ForceSettingsPatch : Patch(
```
oSumAtrIX, line 22: This patch should have a different name. Something like "DownloadPatch" or "DownloadAndBackgroundPlaybackPatch".

src/main/kotlin/app/revanced/patches/layout/ForceSettingsPatch.kt
```diff
@@ -0,0 +1,68 @@
...
+                PatternScanMethod.Fuzzy(0), // FIXME: Test this threshold and find the best value.
+                compatiblePackages,
+                "Signature for the method required to be patched.",
+                "0.0.1"
+            ),
+            "Z",
+            AccessFlags.PUBLIC or AccessFlags.STATIC,
+            listOf("L", "L"),
+            listOf(
+                Opcode.IF_EQZ,
```
oSumAtrIX, line 44: Are you sure this pattern is unique?

src/main/kotlin/app/revanced/patches/layout/ForceSettingsPatch.kt
```diff
@@ -0,0 +1,68 @@
...
+        "Force Download and Background Playback Settings Patch",
+        "Enable Download and Background Playback within Settings.",
+        compatiblePackages,
+        "0.0.1"
+    ),
+    listOf(
+        MethodSignature(
+            MethodSignatureMetadata(
+                "force-settings-patch-manager",
+                MethodMetadata(null, null), // unknown
```
oSumAtrIX, line 34: Please leave a name of the method it matches. The name and class should be from the app version, in which this signature is valid/ has been created from

oSumAtrIX (changes requested):
There might also be a conflicting patch for background playback, you should remove either one with this PR.

dylangerdaly:
Let me get Download working first, then I'll redo this

PR #7, opened 2022-04-26
feat: `YouTube Music` patches

oSumAtrIX:


epicsampler:
> Explain what taste builder means

grafik

it only wastes space on the home page, never goes away and same thing can be triggered from settings anyway

epicsampler:
@oSumAtrIX 

> Explain what taste builder means



it only wastes space on the home page, never goes away and same thing can be triggered from settings anyway

oSumAtrIX:
The description is meant to be long. It should be used to explain the patch, maby add this information there

PR #15, opened 2022-06-04
Update release.yml

josesilveiraa:
Upload pre-built binaries so it's easier to build the patch app.


josesilveiraa:
talking about actions btw

oSumAtrIX:
Packages are already being generated, what's the necessity of this PR

Sculas:
`semantic-release` already publishes releases.

PR #22, opened 2022-06-15
feat: `music-microg-support` patch

oSumAtrIX:


src/main/kotlin/app/revanced/patches/music/misc/microg/patch/bytecode/MusicMicroGBytecodePatch.kt
```diff
@@ -0,0 +1,28 @@
...
+import app.revanced.patcher.patch.annotations.Patch
+import app.revanced.patcher.patch.implementation.BytecodePatch
+import app.revanced.patcher.patch.implementation.misc.PatchResult
+import app.revanced.patcher.patch.implementation.misc.PatchResultSuccess
+import app.revanced.patches.music.misc.microg.annotations.MusicMicroGPatchCompatibility
+import app.revanced.patches.youtube.misc.microg.patch.resource.MicroGResourcePatch
+
+@Patch(include = false)
+@Dependencies(dependencies = [MicroGResourcePatch::class])
+@Name("music-microg-support")
```
Sculas, line 17:
```suggestion
@Name("microg-support")
```
Consistency with the other MicroG Support patch. The manager filters this correctly, not sure about the CLI. If not, that should be fixed on the CLI.

Sculas (changes requested):

src/main/kotlin/app/revanced/patches/music/misc/microg/patch/bytecode/MusicMicroGBytecodePatch.kt
oSumAtrIX: The cli uses the patch name as the key to add the patch, it has to be distinct

Sculas: > The manager filters this correctly, not sure about the CLI. If not, that should be fixed on the CLI.

As I said above, that needs to be properly filtered on the CLI. So that's a CLI issue.

PR #23, opened 2022-06-15
feat: `hide-cast-button` patch

ghost:
Note: Signature is not working yet

### DIFF
--- a/src/main/kotlin/app/revanced/patches/youtube/layout/castbutton/annotations/CastPatchCompatibility.kt
+++ b/src/main/kotlin/app/revanced/patches/youtube/layout/castbutton/annotations/CastPatchCompatibility.kt
@@ -5,7 +5,7 @@ import app.revanced.patcher.annotation.Package
 
 @Compatibility(
     [Package(
-        "com.google.android.youtube", arrayOf("17.14.35", "17.17.34", "17.19.36", "17.20.37", "17.22.36") // only 17.20.37 is tested
+        "com.google.android.youtube", arrayOf("17.20.37")
     )]
 )
 @Target(AnnotationTarget.CLASS)


src/main/kotlin/app/revanced/patches/youtube/layout/castbutton/annotations/CastPatchCompatibility.kt
```diff
@@ -0,0 +1,13 @@
+package app.revanced.patches.youtube.layout.castbutton.annotations
+
+import app.revanced.patcher.annotation.Compatibility
+import app.revanced.patcher.annotation.Package
+
+@Compatibility(
+    [Package(
+        "com.google.android.youtube", arrayOf("17.14.35", "17.17.34", "17.19.36", "17.20.37", "17.22.36") // only 17.20.37 is tested
```
Sculas, line 8:
```suggestion
        "com.google.android.youtube", arrayOf("17.20.37")
```
If only `17.20.37` is tested, you should only put that in the compatibility list.
Only when a patch is tested for a certain version, may it be added to the compatibility list.

src/main/kotlin/app/revanced/patches/youtube/layout/castbutton/signatures/CastButtonSignature.kt
```diff
@@ -0,0 +1,29 @@
...
+
+@Name("cast-button-signature")
+@MatchingMethod(
+    "Landroidx/mediarouter/app/MediaRouteButton", "setVisibility" // first one is prolly causing the issue since not sure what to write there
+)
+@FuzzyPatternScanMethod(2) // FIXME: Test this threshold and find the best value.
+@CastButtonCompatibility
+@Version("0.0.1")
+object CastButtonSignature : MethodSignature(
+    "V", // this might be broken as well since idk what is this
```
Sculas, line 21: This is a `void` type. See [here](https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html#jvms-4.3.2) for more information about types in bytecode.

src/main/kotlin/app/revanced/patches/youtube/layout/castbutton/signatures/CastButtonSignature.kt
```diff
@@ -0,0 +1,29 @@
...
+    AccessFlags.PUBLIC or AccessFlags.FINAL,
+    listOf("L","L"), // this too
```
Sculas, line 23: Like above, this is a `class` type. We do not specify what kind of class/object it is, we just say "it's a class/object".

src/main/kotlin/app/revanced/patches/youtube/layout/castbutton/signatures/CastButtonSignature.kt
```diff
@@ -0,0 +1,29 @@
...
+import app.revanced.patcher.signature.implementation.method.MethodSignature
+import app.revanced.patcher.signature.implementation.method.annotation.FuzzyPatternScanMethod
+import app.revanced.patcher.signature.implementation.method.annotation.MatchingMethod
+import app.revanced.patches.youtube.layout.castbutton.annotations.CastButtonCompatibility
+import org.jf.dexlib2.AccessFlags
+import org.jf.dexlib2.Opcode
+
+@Name("cast-button-signature")
+@MatchingMethod(
+    "Landroidx/mediarouter/app/MediaRouteButton", "setVisibility" // first one is prolly causing the issue since not sure what to write there
```
Sculas, line 15: Why are you hooking `MediaRouteButton#setVisibility`? What is it you're trying to do with this patch?

Sculas (changes requested):

src/main/kotlin/app/revanced/patches/youtube/layout/castbutton/signatures/CastButtonSignature.kt
Canny1913: 👍 
are both of these correct or should i change them?

src/main/kotlin/app/revanced/patches/youtube/layout/castbutton/signatures/CastButtonSignature.kt
Canny1913: ^

i basically tried to do what vanced did in their code

Canny1913:
also i just saw that in xGlobals, castButtonShown is set to true

PR #53, opened 2022-06-23
feat: `materialyou` patch 

datnerdguy:
For now exclude amoled patch while patching (I don't know how to exclude a patch when another is selected so if someone could help me out I would be very much thankful)


oSumAtrIX:
This patch should be combined with the amoled patch to a themer patch. Otherwise they conflict each other

datnerdguy:
How should I visualise the selection ui then?

oSumAtrIX:
YouTube has a setting to switch between light and dark mode, this would fit there.

datnerdguy:
Will reopen this pr wen I am ready.

PR #54, opened 2022-06-23
Update release.yml

Itroublve:


Itroublve:
mb, was testing and mistankely pushed a pr.

PR #55, opened 2022-06-23
fix: Change environment from 20.04 to 18.04

Itroublve:
Fixes the error with `disable-create-button` patch using prebuilt.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,7 +12,7 @@ on:
 jobs:
   release:
     name: Release
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
     steps:
       - name: Checkout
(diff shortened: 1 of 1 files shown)


oSumAtrIX:
ReVanced should be buildable on the latest version from Ubuntu. Did GitHub update it?

Sculas:
Ubuntu 20 has some breaking changes from 18 that don't work well. Figuring out how to get Ubuntu 20 to work is not priority and out of scope. I think this should be merged.

oSumAtrIX:
It would be necessary to at least first understand what causes this issue.

Sculas:
This is an issue people are having right now and it needs to be fixed ASAP. We need a workaround for now, and try to apply a better fix when we've found one.

PR #58, opened 2022-06-23
Removed hide-cast-button as a microg dependency.

Aunali321:
Closes issue ReVanced/revanced-patches-template#2409


oSumAtrIX:
This will cause battery overheating on some devices due to microg

TheJeterLP:
> This will cause battery overheating on some devices due to microg

So? Imo let the user decide what he wants? It is included by default, if someone really wants to exclude it why not let them do that? The user should k is what he's doing then.

PR #67, opened 2022-06-26
feat: add youtube version 17.24.35

Aunali321:
Tested every patch in root and non-root version. Everything works as expected.


TheJeterLP:
Thank you!

PR #77, opened 2022-06-29
feat: make resource mapping patch aware of types

bogadana:


src/main/kotlin/app/revanced/patches/youtube/misc/mapping/patch/ResourceIdMappingProviderResourcePatch.kt
```diff
@@ -15,22 +15,30 @@ import org.w3c.dom.Element
...
             }
         }
 
         return PatchResultSuccess()
     }
-}
\ No newline at end of file
+}
+
+class ResourceElement(type: String, name: String, id: Long) {
```
oSumAtrIX, line 40: Lookup `data class` for this.

PR #79, opened 2022-06-30
feat: `disable-autoplay-button` patch

bogadana:


TheJeterLP:
Shouldn't there be a setting in integrations to be checked?

PR #80, opened 2022-06-30
fix: change fingerprint to work on latest youtube

bogadana:


TheJeterLP:
You should add the YouTube version to compatibility then

bogadana:
betas aren't added afaik

TheJeterLP:
Oh it's a beta alright. Wasn't aware of that

PR #91, opened 2022-06-30
feat: `autorepeat-by-default` patch

TheJeterLP:
This can be merged after https://github.com/revanced/revanced-cli/issues/75 is worked out and after 
https://github.com/revanced/revanced-patcher/pull/52 has been merged

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,17 @@
+# [2.5.0](https://github.com/revanced/revanced-patches/compare/v2.4.0...v2.5.0) (2022-06-30)
+
+
+### Features
+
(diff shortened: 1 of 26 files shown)


TheJeterLP:
> This patch should be improved, to be based on already included repeat video feature.
> 
> Just set the aforementioned to true, without add any new methods or whatever.

There are no new methods added. For now we do the implementation the same way Vanced did. We will change stuff after every patch is done.

OxrxL:
For now we do the implementation the same way Vanced did.

Yeah...I know. Indeed Vanced add a method call inside another one, to create the video loop.

Anyway yes...this is a work to start after core features implementation.

TheJeterLP:
> For now we do the implementation the same way Vanced did.
> 
> Yeah...I know. Indeed Vanced add a method call inside another one, to create the video loop.
> 
> Anyway yes...this is a work to start after core features implementation.

Yes that we do here too for now

PR #110, opened 2022-07-03
fix: Update the Watermark patch description

CnC-Robert:


TheJeterLP:
Thank you :)

oSumAtrIX:
He should've committed the suggestion since it's grammatically wrong.

PR #116, opened 2022-07-04
feat: Enable-Debugging Patch

shadow578:
small patch to set [android:debuggable](https://developer.android.com/guide/topics/manifest/application-element#debug) attribute in manifest. 
This can be useful when developing new patches because it enables the use of some Android studio tools, like
- package filter on logcat
- [Layout Inspector](https://developer.android.com/studio/debug/layout-inspector#layout-inspector)
- Debugging revanced-integrations code (breakpoints and all that good stuff)
- ...


bogadana:
compatibility versions can be removed

PR #117, opened 2022-07-05
fix: remove unnecessary version constraints in debugging patch

bogadana:
this patch is also compatible with every android app but there doesnt seem to be a good way to specify that

### DIFF
--- a/src/main/kotlin/app/revanced/patches/youtube/misc/enabledebugging/annotations/EnableDebuggingCompatibility.kt
+++ b/src/main/kotlin/app/revanced/patches/youtube/misc/enabledebugging/annotations/EnableDebuggingCompatibility.kt
@@ -5,19 +5,7 @@ import app.revanced.patcher.annotation.Package
 
 @Compatibility(
     [Package(
-        "com.google.android.youtube",
-        arrayOf(
-            "17.14.35",
-            "17.17.34",
-            "17.19.36",
-            "17.20.37",
-            "17.22.36",
-            "17.23.35",
-            "17.23.36",
-            "17.24.34",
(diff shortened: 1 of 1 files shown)


oSumAtrIX:
The package name constraint can also be removed if necessary: https://github.com/revanced/revanced-cli/blob/387b6bd0163d660cebd31fc2126c0c57fbae7470/src/main/kotlin/app/revanced/utils/patcher/Patcher.kt#L38

bogadana:
still i think generating warnings when nothing is wrong is undesirable

oSumAtrIX:
You could remove the package constraint as previously said, this will produce an expected warning though.

bogadana:
I can't because `Null can not be a value of a non-null type Array<Package>`

oSumAtrIX:
About the nullability of the annotation properties, if possible, could you PR a proper implementation? Sometimes you have to use the empty array initializer when not required.

oSumAtrIX:
> I can't because `Null can not be a value of a non-null type Array<Package>`

You have to use an empty array

oSumAtrIX:
> and also `An annotation argument must be a compile-time constant`

Iirc you can use []

bogadana:
> could you PR a proper implementation?

Annotation parameters cannot have nullable types, because the JVM does not support storing null as a value of an annotation attribute. (from https://kotlinlang.org/docs/annotations.html)

bogadana:
oh yeah that i can do

bogadana:
empty array results in 
`WARNING: Skipping enable-debugging: Incompatible with com.google.android.youtube. This patch is only compatible with`

PR #118, opened 2022-07-05
fix: make minimized playback patch enable settings

bogadana:


src/main/kotlin/app/revanced/patches/youtube/layout/minimizedplayback/fingerprints/MinimizedPlaybackSettingsFingerprint.kt
```diff
@@ -0,0 +1,39 @@
...
+    listOf(
+        Opcode.INVOKE_VIRTUAL,
+        Opcode.MOVE_RESULT,
+        Opcode.INVOKE_VIRTUAL,
+        Opcode.MOVE_RESULT,
+        Opcode.IF_EQZ,
+    ),
+    customFingerprint = {
+        it.implementation!!.instructions.any {
+            (it as? WideLiteralInstruction)?.wideLiteral == ResourceIdMappingProviderResourcePatch.resourceMappings
```
oSumAtrIX, line 35: This will do a lookup in the map every time. Instead, create a constant for the id you search for and use it in the custom fingerprint.

src/main/kotlin/app/revanced/patches/youtube/layout/minimizedplayback/patch/MinimizedPlaybackPatch.kt
```diff
@@ -32,6 +40,22 @@ class MinimizedPlaybackPatch : BytecodePatch(
                 return v0
                 """
         )
+
+        val method = MinimizedPlaybackSettingsFingerprint.result!!.mutableMethod
+        val booleanCalls = method.implementation!!.instructions.withIndex()
+            .filter { ((it.value as? ReferenceInstruction)?.reference as? MethodReference)?.returnType == "Z" }
```
oSumAtrIX, line 46: use `.single` here. `.filter` should be used if you expect multiple results.

src/main/kotlin/app/revanced/patches/youtube/layout/minimizedplayback/fingerprints/MinimizedPlaybackSettingsFingerprint.kt
```diff
@@ -0,0 +1,40 @@
...
+        Opcode.IF_EQZ,
+    ),
+    customFingerprint = {
+        it.implementation!!.instructions.any {
+            (it as? WideLiteralInstruction)?.wideLiteral == resourceId 
+        }
+    }
+)
+
+val resourceId = ResourceIdMappingProviderResourcePatch.resourceMappings.first { it.type == "string" && it.name == "pref_background_category" }.id
```
oSumAtrIX, line 40: A bit out of place but whatever :)

bogadana: wasn't sure where to put it, what's a better place?

oSumAtrIX: I would've created a `Constants` class. This in general is a dangerous operation tho, since `ResourceIdMappingProviderResourcePatch` has been executed by luck already before this fingerprints class is being instantiated resulting in the `resourceId` being set.

bogadana: Well can fingerprints have dependencies?

bogadana: oh I see

oSumAtrIX: > Well can fingerprints have dependencies?

No, fingerprints should not depend on patches such as `ResourceIdMappingProviderResourcePatch`.

PR #120, opened 2022-07-05
fix: change minimized playback fingerprint

bogadana:
no more constants issues


src/main/kotlin/app/revanced/patches/youtube/layout/minimizedplayback/fingerprints/MinimizedPlaybackSettingsFingerprint.kt
```diff
@@ -20,7 +20,7 @@ import org.jf.dexlib2.iface.instruction.WideLiteralInstruction
 @MinimizedPlaybackCompatibility
 @Version("0.0.1")
 object MinimizedPlaybackSettingsFingerprint : MethodFingerprint(
-    "L",
+    "Ljava/lang/String;",
```
oSumAtrIX, line 23:
```suggestion
    "L",
```

bogadana: I'm not sure this will work but I'll check

xemulat:
new patch?

Issue #132, opened 2022-07-07
Autogenerated list of patches for `README.md`


bogadana:
I intend to create a method which would generate a table like this one

|        Name        |                                       Description                                      | Latest supported version |
|:------------------:|:--------------------------------------------------------------------------------------:|:------------------------:|
| Microg support     | Patch to allow YouTube ReVanced to run without root and under a different package name | 17.25.34                 |
| Old quality layout | Enable the original quality flyout menu.                                               | 17.25.34                 |

But:
1. Where to put it?
2. How to run it? (gradle task?)
3. How to get information about the patches, PatchBundle class from patcher requires an existing jar file
4. Modifications to the table?

oSumAtrIX:
1. The README should consist out of the standard sections of a heading, description, build instructions and meta such as your table. A good generator is [readme.so](https://readme.so). We will need to be consistent in the README across the organization.
2.  A gradle task can generate the new table on publish.
3. I am not sure yet.
4.
| 💊 Patch | 📜 Description | 🎯 Target package | 🏹 Target package version |
|:-:|:-:|:-:|:-:|
| Microg Support | Patch to allow YouTube ReVanced to run without root and under a different package name. | `com.google.android.youtube` | 17.25.34
| Old Quality Layout | Enable the original quality flyout menu. | `com.google.android.youtube` | 17.25.34

jobhax25:
Would the wiki page be good for this? and the readme would send you to check out the page?

oSumAtrIX:
The wiki page should not contain anything specific for a patch

PR #133, opened 2022-07-08
docs: auto generate list of patches in `README.md`

### LINKED ISSUE #2378: Autogenerated list of patches for `README.md`
I intend to create a method which would generate a table like this one

|        Name        |                                       Description                                      | Latest supported version |
|:------------------:|:--------------------------------------------------------------------------------------:|:------------------------:|
| Microg support     | Patch to allow YouTube ReVanced to run without root and under a different package name | 17.25.34                 |
| Old quality layout | Enable the original quality flyout menu.                                               | 17.25.34                 |

But:
1. Where to put it?
2. How to run it? (gradle task?)
3. How to get information about the patches, PatchBundle class from patcher requires an existing jar file
4. Modifications to the table?

bogadana:
Implements https://github.com/revanced/revanced-patches/issues/132

Is this way of getting patch data ok? (with proper path finding)


TheJeterLP:
Can we integrate this into actions somehow instead of doing it in Gradle? This will make many merge conflicts which have to be manually resolved in the future this way

TheJeterLP:
So we should only run this when the Gradle task publish is being run (only used by semantic-release) and then somehow let actions commit the README and push it to main repo. That way this will happen automatically

bogadana:
yes that's the point but actions can't run java code by themselves afaik

TheJeterLP:
> yes that's the point but actions can't run java code by themselves afaik

That's why I said run this on grades publish task

.github/workflows/release.yml
```diff
@@ -42,3 +42,7 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: npx semantic-release
+      - name: Update readme
+        uses: stefanzweifel/git-auto-commit-action@v4
+        with:
+          commit_message: "Update `README.md`"
```
oSumAtrIX, line 48: Does not follow conventional commits or skip CI.

README.md
```diff
@@ -1,2 +1,32 @@
...
+|Premium Heading|Show the premium branding on the the YouTube home screen.|`com.google.android.youtube`|all|
+|Old Quality Layout|Enable the original quality flyout menu.|`com.google.android.youtube`|17.25.34|
+|Hide Shorts Button|Hide the shorts button.|`com.google.android.youtube`|17.25.34|
+|Hide Cast Button|Patch to hide the cast button.|`com.google.android.youtube`|all|
+|Minimized Playback|Enable minimized and background playback.|`com.google.android.youtube`|17.25.34|
+|Disable Fullscreen Panels|Disable comments panel in fullscreen view.|`com.google.android.youtube`|17.25.34|
+|Disable Create Button|Disable the create button.|`com.google.android.youtube`|17.25.34|
+|Hide Watermark|Hide the creator's watermark on video's.|`com.google.android.youtube`|17.25.34|
+|Amoled|Enables pure black theme.|`com.google.android.youtube`|17.25.34|
+|Hide Autoplay Button|Disable the autoplay button.|`com.google.android.youtube`|17.25.34|
```
oSumAtrIX, line 32: This should not be pushed by anyone. The workflow would trigger the task which would trigger a commit which skips CI.

.github/workflows/release.yml
TheJeterLP: Also the actual pushing to main is missing here. It only commits it, or am I missing something?

TheJeterLP: You should also set `branch: main` so it pushes to main. Otherwise it will default to the branch where the job is running on. Also only push the readme when the job is running on the main branch. can be done with an if statement

bogadana: what's wrong with updating the readme on the dev branch?

TheJeterLP: Dev branch is not a problem. but other branches might be

bogadana: this whole workflow only runs on dev and main

TheJeterLP: oh you're right. Forgot about that xD

README.md
Sculas: Fixed.

.github/workflows/release.yml
Sculas: Removed.

src/main/kotlin/app/revanced/meta/ReadmeGenerator.kt
```diff
@@ -0,0 +1,55 @@
...
+                """    
+            # ReVanced Patches
+            🧩 Official patches by ReVanced
+
+            # Patch list
+            """.trimIndent()
+
+            val tableHeader =
+                """
+            |💊Patch|📜Description|🎯Target Package|🏹Target Version|
```
oSumAtrIX, line 26:
```suggestion
            | 💊 Patch | 📜 Description | 🎯 Target Package | 🏹 Target Version |
```

Sculas:
Should be all fixed now. Ready for merging.
The current solution is not perfect, feel free to improve it in the future.

oSumAtrIX:
LGTM.

src/main/kotlin/app/revanced/meta/ReadmeGenerator.kt
Sculas: It won't affect anything. But for our eyes, I approve of this change.

oSumAtrIX:
Future improvements can include converting the `README` to a template, and the task replaces the placeholders.

Sculas (approved):

oSumAtrIX (approved):

PR #134, opened 2022-07-08
fix: patch description consistency

bogadana:
Chose this format because I think its more user friendly than commit message style


TheJeterLP:
Thanks!

PR #147, opened 2022-07-11
fix: listing of wrong fingerprint class

ponces:


TheJeterLP:
That is not the wrong fingerprint?

TheJeterLP:
Thanks for that

ponces:
I still need to fix this patch but at least is not throwing an exception when applying

ponces:
> I still need to fix this patch but at least is not throwing an exception when applying

Actually, this patch (apparently) doesn't make sense as it is patching the return statement by replacing void by a value in a function which states void to be the return type. Right? I'm still not to familar with this type of development work but I guess I am interpreting this right...

@oSumAtrIX you were the one to push the original code. Was this based on some previous patches from Vanced team?

ponces:
> > I still need to fix this patch but at least is not throwing an exception when applying
> 
> Actually, this patch (apparently) doesn't make sense as it is patching the return statement by replacing void by a value in a function which states void to be the return type. Right? I'm still not too familar with this type of development work but I guess I am interpreting this right...
> 
> @oSumAtrIX you were the one that pushed the original code on b60c9d33b611bb4d5b55bb419652bc14b0309792.
> Was this based on some previous patches from Vanced team?
> 

Patch is being applied on wrong method. I'll push a new PR with the working (already tested) fix for this patch

ponces:
> Which patch is not working?

Sorry, i was not clear.

I meant exclusive audio patch

PR #156, opened 2022-07-11
docs: better semver comparison for readme

bogadana:


src/main/kotlin/app/revanced/meta/readme/Semver.kt
```diff
@@ -0,0 +1,24 @@
+data class Semver(val major: Int, val minor: Int, val patch: Int) {
+    companion object {
+        fun fromString(version: String): Semver {
+            var parts = version.split(".")
+
+            if (parts.count() != 3) throw IllegalArgumentException("semver must have 3 parts")
```
oSumAtrIX, line 6:
```suggestion
            if (parts.count() != 3) throw IllegalArgumentException("Invalid semantic version")
```

src/main/kotlin/app/revanced/meta/readme/Semver.kt
```diff
@@ -0,0 +1,24 @@
...
+
+            val intParts = parts.map { it.toInt() }
+            return Semver(intParts[0], intParts[1], intParts[2])
```
oSumAtrIX, line 9:
```suggestion
            val versionNumbers = parts.map { it.toInt() }
            return SemanticVersion(versionNumbers[0], versionNumbers[1], versionNumbers[2])```

File: src/main/kotlin/app/revanced/meta/readme/Semver.kt [hunk 3]
```diff
@@ -0,0 +1,24 @@
+data class Semver(val major: Int, val minor: Int, val patch: Int) {
```
oSumAtrIX, line 1:
```suggestion
data class SemanticVersion(val major: Int, val minor: Int, val patch: Int) {
```

src/main/kotlin/app/revanced/meta/readme/Semver.kt
```diff
@@ -0,0 +1,24 @@
...
+    companion object {
+        fun fromString(version: String): Semver {
```
oSumAtrIX, line 3:
```suggestion
        fun fromString(version: String): SemanticVersion {
```

src/main/kotlin/app/revanced/meta/readme/Semver.kt
```diff
@@ -0,0 +1,24 @@
...
+
+            val intParts = parts.map { it.toInt() }
+            return Semver(intParts[0], intParts[1], intParts[2])
+        }
+    }
+
+    override fun toString(): String = "$major.$minor.$patch"
+}
+
+class SemverComparator {
```
oSumAtrIX, line 16:
```suggestion
class SemanticVersionComparator {
```

src/main/kotlin/app/revanced/meta/readme/Semver.kt
```diff
@@ -0,0 +1,24 @@
...
+    companion object : Comparator<Semver> {
```
oSumAtrIX, line 17: Instead of using a companion object here, simply convert the class to an `object`. It will be compiled into a singleton, where a companion object is a singleton inside the class.

src/main/kotlin/app/revanced/meta/readme/Semver.kt
bogadana: why the backticks?

oSumAtrIX: Newline was missing, fixed.

bogadana:
It is done

oSumAtrIX (approved):

PR #170, opened 2022-07-13
fix: `codecs-unlock` patch and update Music patches to `5.14.53`

ponces:
Versions older than v5.03.50 need a different `codecs-unlock` fingerprint.


oSumAtrIX:
Is this tested?

ponces:
> Is this tested?

If by "tested" you meant, if this is actually patching, then yes.
If you instead meant if I found any evidence of the patch working, then no.
The same happens with the old version when the patch is applied...

As you were the one who pushed this patch, can you guide me on how to do it?
I already enabled "Show stats for nerds" and up until now I didn't find a videos that doesn't have "140 mp4a" in the "Audio format" parameter... :/

oSumAtrIX:
Newer versions changed some methods for which the codec patch has been created, this PR does not consider this change. The patches are not yet compatible with the latest versions. Try patching the current compatible version and check if that works.

ponces:
You are right, current compatible version works.
Mine does not with the latest version.
I will update the patch and open a new PR later.
Thanks for the guidance!

PR #172, opened 2022-07-13
feat: `music-video-ads`

ponces:
This is heavily based on the video-ads patch from YouTube.
Is anyone able to test this PR?
It appears to be working on my side as no ad is being triggered for some time but still...

Edit: been using for quite some time now and indeed no ad appears 👍

### DIFF
--- a/src/main/kotlin/app/revanced/patches/music/ad/video/fingerprints/ShowMusicVideoAdsConstructorFingerprint.kt
+++ b/src/main/kotlin/app/revanced/patches/music/ad/video/fingerprints/ShowMusicVideoAdsConstructorFingerprint.kt
@@ -1,0 +1,39 @@
+package app.revanced.patches.music.ad.video.fingerprints
+
+import app.revanced.patcher.annotation.Name
+import app.revanced.patcher.annotation.Version
+import app.revanced.patcher.extensions.or
+import app.revanced.patcher.fingerprint.method.impl.MethodFingerprint
+import app.revanced.patcher.fingerprint.method.annotation.FuzzyPatternScanMethod
+import app.revanced.patcher.fingerprint.method.annotation.MatchingMethod
+import app.revanced.patches.music.ad.video.annotations.MusicVideoAdsCompatibility
+import org.jf.dexlib2.AccessFlags
+import org.jf.dexlib2.Opcode
+
+@Name("show-video-ads-constructor-fingerprint")
+@MatchingMethod(
+    "Laghd;",
+    "<init>",
+)
+@FuzzyPatternScanMethod(2) // FIXME: Test this threshold and find the best value.
+@MusicVideoAdsCompatibility
+@Version("0.0.1")
+object ShowMusicVideoAdsConstructorFingerprint : MethodFingerprint(
+    "V", AccessFlags.PUBLIC or AccessFlags.CONSTRUCTOR, listOf("L", "L", "L"), listOf(
+        Opcode.INVOKE_DIRECT,
+        Opcode.NEW_INSTANCE,
+        Opcode.INVOKE_DIRECT,
+        Opcode.IPUT_OBJECT,
+        Opcode.NEW_INSTANCE,
+        Opcode.INVOKE_DIRECT,
+        Opcode.IPUT_OBJECT,
+        Opcode.NEW_INSTANCE,
+        Opcode.INVOKE_DIRECT,
+        Opcode.IPUT_OBJECT,
+        Opcode.IPUT_OBJECT,
+        Opcode.IPUT_OBJECT,
+        Opcode.CONST_4,
+        Opcode.IPUT_BOOLEAN,
+        Opcode.RETURN_VOID
+    )
+)
--- a/src/main/kotlin/app/revanced/patches/music/ad/video/fingerprints/ShowMusicVideoAdsFingerprint.kt
+++ b/src/main/kotlin/app/revanced/patches/music/ad/video/fingerprints/ShowMusicVideoAdsFingerprint.kt
@@ -1,0 +1,26 @@
+package app.revanced.patches.music.ad.video.fingerprints
+
+import app.revanced.patcher.annotation.Name
+import app.revanced.patcher.annotation.Version
+import app.revanced.patcher.extensions.or
+import app.revanced.patcher.fingerprint.method.annotation.FuzzyPatternScanMethod
+import app.revanced.patcher.fingerprint.method.annotation.MatchingMethod
+import app.revanced.patcher.fingerprint.method.impl.MethodFingerprint
+import app.revanced.patches.music.ad.video.annotations.MusicVideoAdsCompatibility
+import org.jf.dexlib2.AccessFlags
+import org.jf.dexlib2.Opcode
+
+@Name("show-video-ads-method-fingerprint")
+@MatchingMethod(
+    "Laghd;", "d"
+)
+@FuzzyPatternScanMethod(2) // FIXME: Test this threshold and find the best value.
+@MusicVideoAdsCompatibility
+@Version("0.0.1")
+object ShowMusicVideoAdsFingerprint : MethodFingerprint(
+    "V", AccessFlags.PUBLIC or AccessFlags.FINAL, listOf("Z"), listOf(
+        Opcode.IPUT_BOOLEAN,
+        Opcode.INVOKE_VIRTUAL,
+        Opcode.RETURN_VOID
+    )
+)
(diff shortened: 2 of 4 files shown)


ponces:
@oSumAtrIX @TheJeterLP 
Could you please test this?

TheJeterLP:
Yes, as soon as I have time I can try to test this. Will take a few days tho

src/main/kotlin/app/revanced/patches/music/ad/video/fingerprints/ShowMusicVideoAdsFingerprint.kt
```diff
@@ -0,0 +1,21 @@
...
+
+@Name("show-video-ads-method-fingerprint")
+@MatchingMethod(
+    definingClass = "zai"
+)
+@DirectPatternScanMethod
+@MusicVideoAdsCompatibility
+@Version("0.0.1")
+object ShowMusicVideoAdsFingerprint : MethodFingerprint(
+    "V", AccessFlags.PUBLIC or AccessFlags.FINAL, listOf("Z"), null
```
oSumAtrIX, line 20: This fingerprint looks like it will not match to the desired method.

oSumAtrIX (changes requested):
For YT music I don't think we should use the integrations yet. The plan is later that patches are bundled with their own integrations and also can make use of a shared one (like libraries basically). Simply inject the set visibility and call yourself. Also, I can not test the patches yet because the root variant does not work? How did you test it?

src/main/kotlin/app/revanced/patches/music/ad/video/fingerprints/ShowMusicVideoAdsFingerprint.kt
ponces: Fixed this too

ponces:
> For YT music I don't think we should use the integrations yet. The plan is later that patches are bundled with their own integrations and also can make use of a shared one (like libraries basically). Simply inject the set visibility and call yourself. Also, I can not test the patches yet because the root variant does not work? How did you test it?

Ok! Current version of this PR no longer depends on integrations.
It seems to be working just fine! Tomorrow I'll make a longer test just to make sure there's not ad popping up...
Could you please take a look at it @oSumAtrIX ?
Regarding my testing method, I can already patch the old and the latest Music version available but I have to do a little gimmick on the mount script to include the mount of a lib which Music depends on and is currently available in the APK. Somehow the Vanced team had no issue with this as I can't see in the Vanced Manager archived sources an approach to this. So I assume they had no issue with this lib. The issue is correctly described in https://github.com/revanced/revanced-cli/issues/35#issuecomment-1159479181 and the fix too in https://github.com/revanced/revanced-cli/issues/35#issuecomment-1159659757. You can take a look at my current script to build a Magisk module for Music [here](https://gist.github.com/ponces/379956e67c87664975a51025608df0a2).

src/main/kotlin/app/revanced/patches/music/ad/video/fingerprints/ShowMusicVideoAdsFingerprint.kt
```diff
@@ -0,0 +1,26 @@
...
+@FuzzyPatternScanMethod(2) // FIXME: Test this threshold and find the best value.
+@MusicVideoAdsCompatibility
+@Version("0.0.1")
+object ShowMusicVideoAdsFingerprint : MethodFingerprint(
+    "V", AccessFlags.PUBLIC or AccessFlags.FINAL, listOf("Z"), listOf(
+        Opcode.IPUT_BOOLEAN,
+        Opcode.INVOKE_VIRTUAL,
+        Opcode.RETURN_VOID
+    )
+)
```
oSumAtrIX, line 26: Is this really finding the method you need?

ponces: This is the method that it targets: 
image
How should I do it then?

ponces: Today I can confirm with a higher level of certainty  that it's working 😄

Canny1913: Try using jadx and find the code.

ponces: Thank you but no luck too :(

ponces: If there is someone who can test this just to make sure I'm not wrong, it would be great. I can even provide the binary that I'm using by PM

Canny1913: Maybe it's patching to the wrong location?

Canny1913: I will test it once i get back to PC and maybe change some things about fingerprint.

Canny1913: image
yeah it patches

ponces: Yeah I know I got it with my latest push using the constructor fingerprint!
But it's crashing that is why I didn't tell earlier about this push. I'll try to fix it still today and let you know.
Thank you for the test!!
image

src/main/kotlin/app/revanced/patches/music/ad/video/fingerprints/ShowMusicVideoAdsConstructorFingerprint.kt
```diff
@@ -0,0 +1,39 @@
...
+@MatchingMethod(
+    "Laghd",
+    "<init>",
+)
+@FuzzyPatternScanMethod(2) // FIXME: Test this threshold and find the best value.
+@MusicVideoAdsCompatibility
+@Version("0.0.1")
+object ShowMusicVideoAdsConstructorFingerprint : MethodFingerprint(
+    "V", AccessFlags.PUBLIC or AccessFlags.CONSTRUCTOR, listOf("L", "L", "L"), listOf(
+        Opcode.INVOKE_DIRECT,
```
Canny1913, line 23: too many opcodes, i recommend you using less so it has a higher chance of surviving across new versions.

Canny1913:
image
doesn't work

Canny1913:
i will when i can.

src/main/kotlin/app/revanced/patches/music/ad/video/fingerprints/ShowMusicVideoAdsConstructorFingerprint.kt
```diff
@@ -0,0 +1,39 @@
...
+import app.revanced.patcher.fingerprint.method.impl.MethodFingerprint
+import app.revanced.patcher.fingerprint.method.annotation.FuzzyPatternScanMethod
+import app.revanced.patcher.fingerprint.method.annotation.MatchingMethod
+import app.revanced.patches.music.ad.video.annotations.MusicVideoAdsCompatibility
+import org.jf.dexlib2.AccessFlags
+import org.jf.dexlib2.Opcode
+
+@Name("show-video-ads-constructor-fingerprint")
+@MatchingMethod(
+    "Laghd",
```
oSumAtrIX, line 15:
```suggestion
    "Laghd;",
```

ponces: No need for the setVisibility anymore. The patch became even simpler and more close to what the Vanced team did 😄 
Style suggestion fixed.
Are you able to merge it now?

src/main/kotlin/app/revanced/patches/music/ad/video/fingerprints/ShowMusicVideoAdsFingerprint.kt
ponces: > The reason why it was crashing: It is crashing because you try to call set visibility on the instance object inside register p1 but p1 is a boolean and not an instance of view. Replace the first argument of the invoke call with p0 which equals to the this instance as the method is an instance method.

No need for the setVisibility anymore. The patch became even simpler and more close to what the Vanced team did 😄
Are you able to merge it now?

ponces:
Are you able to merge it now?

src/main/kotlin/app/revanced/patches/music/ad/video/fingerprints/ShowMusicVideoAdsConstructorFingerprint.kt
oSumAtrIX: I suppose for now this suffices. It is not too big.

PR #175, opened 2022-07-14
feat: return youtube dislikes

j4k0xb:
related discussion: https://github.com/orgs/revanced/discussions/237, integrations PR: https://github.com/revanced/revanced-integrations/pull/81

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,26 @@
+## [2.14.1](https://github.com/revanced/revanced-patches/compare/v2.14.0...v2.14.1) (2022-07-15)
+
+
+### Bug Fixes
+
+* `codecs-unlock` patch and update Music patches to `5.14.53` ([2437d30](https://github.com/revanced/revanced-patches/commit/2437d3070f6a630d353619f642cefebd47abee20))
+
+# [2.14.0](https://github.com/revanced/revanced-patches/compare/v2.13.6...v2.14.0) (2022-07-15)
+
+
+### Features
+
+* `music-video-ads` patch ([#172](https://github.com/revanced/revanced-patches/issues/172)) ([cbb71b5](https://github.com/revanced/revanced-patches/commit/cbb71b5faf1386ed31dd9e0387f6cf0509d92d31))
+
+## [2.13.6](https://github.com/revanced/revanced-patches/compare/v2.13.5...v2.13.6) (2022-07-14)
+
+## [2.13.5](https://github.com/revanced/revanced-patches/compare/v2.13.4...v2.13.5) (2022-07-14)
+
+
+### Bug Fixes
+
+* `swipe-controls` with active engagement panel ([#177](https://github.com/revanced/revanced-patches/issues/177)) ([000ec6d](https://github.com/revanced/revanced-patches/commit/000ec6d8f6ecbb910a06ec852564ee9e5f03dcf8))
+
 ## [2.13.4](https://github.com/revanced/revanced-patches/compare/v2.13.3...v2.13.4) (2022-07-13)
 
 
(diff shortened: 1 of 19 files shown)


src/main/kotlin/app/revanced/patches/youtube/layout/returnyoutubedislikes/patch/RYDPatch.kt
```diff
@@ -0,0 +1,80 @@
...
+            """
+            const/4 v0, 0
+            invoke-static {v0}, Lapp/revanced/integrations/patches/ReturnYouTubeDislikesPatch;->sendVote(I)V
+            """
+        )
+
+        VideoIdPatch.injectCall("Lapp/revanced/integrations/patches/ReturnYouTubeDislikesPatch;->newVideoLoaded(Ljava/lang/String;)V")
+
+        val parentResult = TextComponentSpecParentFingerprint.result!!
+        val classDef = parentResult.classDef
```
oSumAtrIX, line 59: You can use the fingerprints mutable hosting class, using `data.proxy(classDef)` is effectively the same:

https://github.com/revanced/revanced-patcher/blob/21b540418029c0c25baec897c414debbc11ead3a/src/main/kotlin/app/revanced/patcher/fingerprint/method/impl/MethodFingerprint.kt#L59

src/main/kotlin/app/revanced/patches/youtube/layout/returnyoutubedislikes/patch/RYDPatch.kt
```diff
@@ -0,0 +1,80 @@
...
+                    .all { param -> param == "Ljava/util/concurrent/atomic/AtomicReference;" }
+            }
+            ?: return PatchResultError("TextComponentSpec.createComponent not found")
+
+        val textRefParam = createComponentMethod.parameters.size - 2
+
+        createComponentMethod.addInstructions(
+            0,
+            """
+            move-object/from16 v0, p5
```
oSumAtrIX, line 72: Did you account for the magic register numbers? Sometimes creating a named variable and assigning it the register value explains what the register is used for. Additionally you can dynamically get the register, for the case it changes in the future. Not necessary but a QoL idea.

src/main/kotlin/app/revanced/patches/youtube/misc/videoid/patch/VideoIdPatch.kt
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    companion object {
+        private var offset = 2
+
+        /**
+         * Adds an invoke-static instruction, called with the new id when the video changes
+         * @param methodDescriptor which method to call. Params have to be `Ljava/lang/String;`
+         */
+        fun injectCall(
```
oSumAtrIX, line 40: Is this meant to be public? For the scope of this patch, this should be private.

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/youtube/misc/videoid/patch/VideoIdPatch.kt
j4k0xb: Yes, that way patches that also need it can add their own methods like:
https://github.com/revanced/revanced-patches/pull/175/files#diff-3897157b7579c9e0f9618a8b8fe3c1e0bc754afb7aba449fd7d96aa9ff8cae8bR56

src/main/kotlin/app/revanced/patches/youtube/layout/returnyoutubedislikes/patch/RYDPatch.kt
j4k0xb: Good idea with getting them dynamically, but the params havent rly changed since v17.14.35 so I just hardcoded them

oSumAtrIX:
Alright, looks good so far, mark the PRs as ready to review and we can get them merged

oSumAtrIX (approved):

PR #195, opened 2022-07-17
feat: `hide-get-premium` patch

ponces:
Before (original):

After partial patch (hidden bottom panel):

After successful patching:

### DIFF
--- a/src/main/kotlin/app/revanced/patches/music/layout/premium/fingerprints/HideGetPremiumFingerprint.kt
+++ b/src/main/kotlin/app/revanced/patches/music/layout/premium/fingerprints/HideGetPremiumFingerprint.kt
@@ -1,0 +1,25 @@
+package app.revanced.patches.music.layout.premium.fingerprints
+
+import app.revanced.patcher.annotation.Name
+import app.revanced.patcher.annotation.Version
+import app.revanced.patcher.extensions.or
+import app.revanced.patcher.fingerprint.method.impl.MethodFingerprint
+import app.revanced.patcher.fingerprint.method.annotation.MatchingMethod
+import app.revanced.patches.music.layout.premium.annotations.HideGetPremiumCompatibility
+import org.jf.dexlib2.AccessFlags
+import org.jf.dexlib2.Opcode
+
+@Name("hide-get-premium-fingerprint")
+@MatchingMethod(
+    "Lktn;", "k"
+)
+@HideGetPremiumCompatibility
+@Version("0.0.1")
+object HideGetPremiumFingerprint : MethodFingerprint(
+    "V",
+    AccessFlags.PUBLIC or AccessFlags.FINAL,
+    listOf(),
+    null,
+    listOf("FEmusic_history"),
+    null
+)
--- a/src/main/kotlin/app/revanced/patches/music/layout/premium/fingerprints/HideGetPremiumFingerprint.kt
+++ b/src/main/kotlin/app/revanced/patches/music/layout/premium/fingerprints/HideGetPremiumFingerprint.kt
@@ -16,10 +16,11 @@ import org.jf.dexlib2.Opcode
 @HideGetPremiumCompatibility
 @Version("0.0.1")
 object HideGetPremiumFingerprint : MethodFingerprint(
-    "V",
-    AccessFlags.PUBLIC or AccessFlags.FINAL,
-    listOf(),
-    null,
-    listOf("FEmusic_history"),
-    null
+    "V", AccessFlags.PUBLIC or AccessFlags.FINAL, listOf(), listOf(
+        Opcode.IF_NEZ,
+        Opcode.CONST_16,
+        Opcode.GOTO,
+        Opcode.NOP,
+        Opcode.INVOKE_VIRTUAL
+    )
 )
--- a/src/main/kotlin/app/revanced/patches/music/layout/premium/fingerprints/HideGetPremiumParentFingerprint.kt
+++ b/src/main/kotlin/app/revanced/patches/music/layout/premium/fingerprints/HideGetPremiumParentFingerprint.kt
@@ -11,15 +11,19 @@ import org.jf.dexlib2.Opcode
 
 @Name("hide-get-premium-parent-fingerprint")
 @MatchingMethod(
-    "Lktn;", "i"
+    "Lktn;", "k"
 )
 @HideGetPremiumCompatibility
 @Version("0.0.1")
 object HideGetPremiumParentFingerprint : MethodFingerprint(
-    "L",
-    AccessFlags.PUBLIC or AccessFlags.FINAL,
-    listOf(),
-    null,
-    listOf("avatar_menu_activate_switch_account"),
-    null
+    "V", AccessFlags.PUBLIC or AccessFlags.FINAL, listOf(), listOf(
+        Opcode.IGET_BOOLEAN,
+        Opcode.CONST_4,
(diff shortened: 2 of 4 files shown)


src/main/kotlin/app/revanced/patches/music/layout/premium/fingerprints/HideGetPremiumFingerprint.kt
```diff
@@ -0,0 +1,27 @@
...
+import app.revanced.patcher.fingerprint.method.annotation.MatchingMethod
+import app.revanced.patches.music.layout.premium.annotations.HideGetPremiumCompatibility
+import org.jf.dexlib2.AccessFlags
+import org.jf.dexlib2.Opcode
+
+@Name("hide-get-premium-fingerprint")
+@MatchingMethod(
+    "Lktn;", "k"
+)
+@FuzzyPatternScanMethod(2) // FIXME: Test this threshold and find the best value.
```
oSumAtrIX, line 17:
```suggestion

```

src/main/kotlin/app/revanced/patches/music/layout/premium/fingerprints/HideGetPremiumParentFingerprint.kt
```diff
@@ -0,0 +1,27 @@
...
+import app.revanced.patcher.fingerprint.method.annotation.MatchingMethod
+import app.revanced.patches.music.layout.premium.annotations.HideGetPremiumCompatibility
+import org.jf.dexlib2.AccessFlags
+import org.jf.dexlib2.Opcode
+
+@Name("hide-get-premium-parent-fingerprint")
+@MatchingMethod(
+    "Lktn;", "i"
+)
+@FuzzyPatternScanMethod(2) // FIXME: Test this threshold and find the best value.
```
oSumAtrIX, line 17:
```suggestion

```

src/main/kotlin/app/revanced/patches/music/layout/premium/patch/HideGetPremiumPatch.kt
```diff
@@ -0,0 +1,55 @@
...
+        val parentResult = HideGetPremiumParentFingerprint.result!!
+        HideGetPremiumFingerprint.resolve(data, parentResult.classDef)
+
+        val method = HideGetPremiumFingerprint.result!!.mutableMethod
+
+        val firstIndex = method.implementation!!.instructions.indexOfFirst {
+            ((it as? BuilderInstruction35c)?.reference.toString() == "Landroid/text/TextUtils;->isEmpty(Ljava/lang/CharSequence;)Z")
+        }
+        method.replaceInstruction(
+            firstIndex - 6, """
```
oSumAtrIX, line 39: Is -6 not a bit too far? In the fingerprint `HideGetPremiumFingerprint` add a 1-x opcodes which match to the instructions you need to change.

src/main/kotlin/app/revanced/patches/music/layout/premium/patch/HideGetPremiumPatch.kt
```diff
@@ -0,0 +1,55 @@
...
+            const/4 v1, 0x0
+        """
+        )
+
+        val secondIndex = method.implementation!!.instructions.indexOfFirst {
+            ((it as? BuilderInstruction35c)?.reference.toString() == "Landroid/view/View;->setVisibility(I)V")
+        }
+        method.addInstructions(
+            secondIndex - 4, """
```
oSumAtrIX, line 48: Same here

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/music/layout/premium/patch/HideGetPremiumPatch.kt
ponces: I'm not using opcodes on the fingerprint, I'm using a string... how can I achieve what you're asking?

oSumAtrIX: Check how other fingerprints are made with both an opcode pattern and strings

ponces: I have to get the index of two places in this (quite huge) method so I'll split into two fingerprint classes to get each index. Is this the right thing to do?

TheJeterLP: Theres multiple ways to do what you want to do

oSumAtrIX: You'd need to simply create a fingerprint to find the method, then a second one with just the opcode pattern and resolve it to the method. The second fingerprint will then contain the information for the scan result.

ponces: Ok. But then I should have a third fingerprint with another opcode pattern to get the second place in the method which I want to retrieve. Does this make sense?

oSumAtrIX: No, that would be done by the first fingerprint already.

ponces: Ok, so the first fingerprint should find the method and the first index which I want to change. The second fingerprint should find the second index alone, using the first fingerprint to resolve/find the method. Is this ok now?

ponces: It's working with https://github.com/revanced/revanced-patches/pull/195/commits/f0f5450ea8080069b68fb61041018fd0dd89abfe! Thanks for the help!
Always great to learn from you both :)

src/main/kotlin/app/revanced/patches/music/layout/premium/patch/HideGetPremiumPatch.kt
ponces: Fixed on https://github.com/revanced/revanced-patches/pull/195/commits/f0f5450ea8080069b68fb61041018fd0dd89abfe

TheJeterLP:
Looks good to me 👍🏻

TheJeterLP:
I don't know how the target method looks like, but instead of creating two fingerprints which do exactly the same (dupe code) you can make one fingerprint, put all the opcodes in there from your first instruction you want to change until the last instruction you want to change and just do startIndex and endIndex.

ponces:
> I don't know how the target method looks like, but instead of creating two fingerprints which do exactly the same (dupe code) you can make one fingerprint, put all the opcodes in there from your first instruction you want to change until the last instruction you want to change and just do startIndex and endIndex.

Can't do that as the target method is quite huge and the opcodes list would have a length too big. Which in the future can lead to patching failures easily...

TheJeterLP:
> > I don't know how the target method looks like, but instead of creating two fingerprints which do exactly the same (dupe code) you can make one fingerprint, put all the opcodes in there from your first instruction you want to change until the last instruction you want to change and just do startIndex and endIndex.
> 
> Can't do that as the target method is quite huge and the opcodes list would have a length too big. Which in the future can lead to patching failures easily...

Yeah okay then that is not an option 👍🏻

PR #208, opened 2022-07-19
feat: complete `music-microg-support` patch

Canny1913:
Still needs to be dedupelicated.


Canny1913:
Yes it worked for me with and without the manifest patch.
Check the apk alignment using zipalign.

ponces:
I'm using the root version. Do I even need it?

Canny1913:
Manifest patch will not work on root due to how root version is installed.

ponces:
Yeah I know that but the weird part is that Vanced did not had any workaround for this lib issue apparently. Manager installed the stock version and used a mount of the patched version on it. No lib fix or another mount script handling the lib folder (as I'm doing as a workaround)... Is this libsjc.so a recent requirement for Music?

Actually, on Vanced even the root version has the extractNativeLibs flag enabled although I don't believe it has any impact on the patched version...

oSumAtrIX:
Vanced had this issue with their last supported versions of Music too. It was said, that certain architectures failed to work via the root variant.

ponces:
I have it working on v4.64.51. Is this their last?

Canny1913:
> Yeah I know that but the weird part is that Vanced did not had any workaround for this lib issue apparently. Manager installed the stock version and used a mount of the patched version on it. No lib fix or another mount script handling the lib folder (as I'm doing as a workaround)... Is this libsjc.so a recent requirement for Music?
> 
> Actually, on Vanced even the root version has the extractNativeLibs flag enabled although I don't believe it has any impact on the patched version...

That library is being used for quite a while and yes nativeLibs will not impact the rooted version because it only extracts libs if the apk is installed, not mounted.


> Vanced had this issue with their last supported versions of Music too. It was said, that certain architectures failed to work via the root variant.

zipalign shows that arm and x86 libraries were "compressed" while arm64 is not "compressed". Maybe that's why it doesn't work on some archs.

oSumAtrIX:
@Canny1913 can you try aligning them with zipalign and check if that works?

Canny1913:
> @Canny1913 can you try aligning them with zipalign and check if that works?

The "compressed" binaries?

oSumAtrIX:
The entire application

Canny1913:
Sure, one sec

Canny1913:
By the way apparently 4.39.50 crashes with the same error so imma inspect that

src/main/kotlin/app/revanced/patches/music/misc/microg/fingerprints/GooglePlayUtilityFingerprint.kt
```diff
@@ -0,0 +1,21 @@
...
+import app.revanced.patcher.fingerprint.method.annotation.FuzzyPatternScanMethod
+import app.revanced.patcher.fingerprint.method.annotation.MatchingMethod
+import app.revanced.patches.music.misc.microg.annotations.MusicMicroGPatchCompatibility
+import org.jf.dexlib2.AccessFlags
+
+@Name("google-play-utility-fingerprint")
+@MatchingMethod(
+    "Lnuv;", "b"
+)
+@FuzzyPatternScanMethod(2) // FIXME: Test this threshold and find the best value.
```
oSumAtrIX, line 16:
```suggestion
import app.revanced.patcher.fingerprint.method.annotation.DirectPatternScanMethod
import app.revanced.patcher.fingerprint.method.annotation.MatchingMethod
import app.revanced.patches.music.misc.microg.annotations.MusicMicroGPatchCompatibility
import org.jf.dexlib2.AccessFlags

@Name("google-play-utility-fingerprint")
@MatchingMethod(
    "Lnuv;", "b"
)
@DirectPatternScanMethod
```

src/main/kotlin/app/revanced/patches/music/misc/microg/fingerprints/ServiceCheckFingerprint.kt
```diff
@@ -0,0 +1,22 @@
...
+
+@Name("google-play-service-checker-fingerprint")
+@MatchingMethod(
+    "Lnuv;", "d"
+)
+@FuzzyPatternScanMethod(2) // FIXME: Test this threshold and find the best value.
+@MusicMicroGPatchCompatibility
+@Version("0.0.1")
+object ServiceCheckFingerprint : MethodFingerprint(
+    "V", AccessFlags.PUBLIC or AccessFlags.STATIC, listOf("L", "I"), listOf(null,
```
oSumAtrIX, line 20: Why do you have a listOf(null) here?

src/main/kotlin/app/revanced/patches/music/misc/microg/patch/bytecode/MusicMicroGBytecodePatch.kt
```diff
@@ -0,0 +1,164 @@
...
+import org.jf.dexlib2.iface.instruction.formats.Instruction21c
+import org.jf.dexlib2.iface.reference.StringReference
+import org.jf.dexlib2.immutable.reference.ImmutableStringReference
+
+@Patch
+@Dependencies(
+    dependencies = [
+        MusicMicroGResourcePatch::class,
+    ]
+)
```
oSumAtrIX, line 34:
```suggestion
@Dependencies([MusicMicroGResourcePatch::class])
```

src/main/kotlin/app/revanced/patches/music/misc/microg/patch/bytecode/MusicMicroGBytecodePatch.kt
```diff
@@ -0,0 +1,164 @@
+package app.revanced.patches.music.misc.microg.patch.bytecode
```
oSumAtrIX, line 1: This file is odly similar to the YouTube MicroG patch. Consider creating a shared dependency and using them in the YouTube as well as YouTube Music patch.

src/main/kotlin/app/revanced/patches/music/misc/microg/patch/resource/MusicMicroGResourcePatch.kt
```diff
@@ -0,0 +1,44 @@
+package app.revanced.patches.music.misc.microg.patch.resource
```
oSumAtrIX, line 1: This file seems also very similar.

src/main/kotlin/app/revanced/patches/youtube/misc/microg/shared/Constants.kt
```diff
@@ -4,4 +4,6 @@ object Constants {
     internal const val BASE_MICROG_PACKAGE_NAME = "com.mgoogle"
     internal const val REVANCED_PACKAGE_NAME = "app.revanced.android.youtube"
     internal const val REVANCED_APP_NAME = "YouTube ReVanced"
+    internal const val REVANCED_MUSIC_APP_NAME = "YT Music ReVanced"
```
oSumAtrIX, line 7: The constant is in the wrong package, it should be moved to the MicroG patch package

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/music/misc/microg/fingerprints/ServiceCheckFingerprint.kt
Canny1913: the string was enough and i decided to not use opcodes

TheJeterLP: Then use null instead of listof(null)

Canny1913: done

src/main/kotlin/app/revanced/patches/music/misc/microg/patch/resource/MusicMicroGResourcePatch.kt
Canny1913: dont know how to do both currently but will figure it out a way 👍

src/main/kotlin/app/revanced/patches/youtube/misc/microg/shared/Constants.kt
Canny1913: yeah makes no sense mb

Canny1913: done

oSumAtrIX:
> Still needs to be dedupelicated.

Do not forget to open a tracking issue for this.

Canny1913:
> > Still needs to be dedupelicated.
> 
> Do not forget to open a tracking issue for this.

I was just about to hit enter :D
But yeah this can be merged for now.

src/main/kotlin/app/revanced/patches/music/misc/microg/shared/Constants.kt
```diff
@@ -0,0 +1,7 @@
+package app.revanced.patches.music.misc.microg.shared
+
+object Constants {
+    internal const val BASE_MICROG_PACKAGE_NAME = "com.mgoogle"
+    internal const val REVANCED_MUSIC_APP_NAME = "YT Music ReVanced"
```
oSumAtrIX, line 5: The brand name should not be morphed:
```suggestion
    internal const val REVANCED_MUSIC_APP_NAME = "YouTube Music ReVanced"
```

oSumAtrIX (changes requested):

oSumAtrIX:
Thanks!

PR #225, opened 2022-07-23
fix: bump youtube version to 17.28.34

ponces:


TheJeterLP:
I guess that you tested if all the patches work now ? :)

ponces:
> I guess that you tested if all the patches work now ? :) 

Yes with the non-split version 😅

PR #236, opened 2022-07-26
fix: bump youtube version to 17.29.34

Canny1913:
(yes i wrote the commit name wrong again)

PR #239, opened 2022-07-26
readd some microg opcode patterns

ghost:
shorter opcode patterns for wider compatibility that I made while porting microg patch to 17.29.xx

just didn't notice that @Canny1913 was faster https://github.com/revanced/revanced-patches/pull/236

however canny completely removed the opcodes

these could be added back

### DIFF
--- a/src/main/kotlin/app/revanced/patches/youtube/misc/microg/fingerprints/GooglePlayUtilityFingerprint.kt
+++ b/src/main/kotlin/app/revanced/patches/youtube/misc/microg/fingerprints/GooglePlayUtilityFingerprint.kt
@@ -8,6 +8,7 @@ import app.revanced.patcher.fingerprint.method.annotation.MatchingMethod
 import app.revanced.patcher.fingerprint.method.impl.MethodFingerprint
 import app.revanced.patches.youtube.misc.microg.annotations.MicroGPatchCompatibility
 import org.jf.dexlib2.AccessFlags
+import org.jf.dexlib2.Opcode
 
 @Name("google-play-utility-fingerprint")
 @MatchingMethod(
@@ -17,5 +18,22 @@ import org.jf.dexlib2.AccessFlags
 @MicroGPatchCompatibility
 @Version("0.0.1")
 object GooglePlayUtilityFingerprint : MethodFingerprint(
-    "I", AccessFlags.PUBLIC or AccessFlags.STATIC, listOf("L", "L"), null, listOf("This should never happen.", "MetadataValueReader", "com.google.android.gms")
+    "I", AccessFlags.PUBLIC or AccessFlags.STATIC, listOf("L", "L"), listOf(
+        Opcode.MOVE_RESULT_OBJECT,
+        Opcode.INVOKE_STATIC,
+        Opcode.GOTO,
+        Opcode.IGET_BOOLEAN,
+        Opcode.IF_NEZ,
+        Opcode.CONST_4,
+        Opcode.GOTO,
+        Opcode.RETURN,
+        Opcode.CONST_STRING,
+        Opcode.INVOKE_STATIC,
+        Opcode.MOVE_RESULT_OBJECT,
+        Opcode.CONST_STRING,
+        Opcode.INVOKE_VIRTUAL,
+        Opcode.MOVE_RESULT_OBJECT,
+        Opcode.INVOKE_STATIC,
+        Opcode.RETURN
+    ), listOf("This should never happen.", "MetadataValueReader", "com.google.android.gms")
 )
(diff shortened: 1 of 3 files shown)


Canny1913:
why readd those back
strings already work great

epicsampler:
> why readd those back strings already work great

more unique fingerprints, possibly prevents wrong methods being found
also had those ready anyway and they work fine

comes down to personal preference how "safe"/"unique" a fingerprint should be
this is just a suggestion
feel free to (not) merge

oSumAtrIX:
> comes down to personal preference how "safe"/"unique" a fingerprint should be this is just a suggestion feel free to (not) merge

The combination of the current strings for each patch are very likely to not be found anywhere else ever. I suppose its fine to close this for that reason?

epicsampler:
> The combination of the current strings for each fingerprint are very likely to not be found anywhere else ever. I suppose its fine to close this for that reason?

currently. you never know about the future.

my idea of fingerprints is always to make them very "safe" so that they intentionally break if something changes and not find some unrelated methods

since these opcode patterns do not have downside currently, I vote for keeping them

oSumAtrIX:
> since these opcodes do not have downside currently I vote for keeping them

but you said:
> currently. you never know about the future.

For that reason, the current strings also "currently" do not have a downside.

The fingerprint is being used without a pattern scan result because no pattern is included. This means even if the pattern changes the patch has the potential to succeed due to the patch working with it dynamically. Making it intentionally fail is counter-intuitive.

TheJeterLP:
Only using Strings is probably the safest method to use. I'm fine with not using opcodes at all.

oSumAtrIX:
Thanks regardless.

epicsampler:
> In the future, make sure to open an issue and discuss something like this so your work does not get rejected.

sure just in this case it was done already anyway

TheJeterLP:
> > For that reason, the current strings also "currently" do not have a downside.
> 
> best of both worlds opcodes currently have no downside and work but help making the fingerprints more safe for the future
> 
> > Making it intentionally fail is counter-intuitive.
> 
> Depends. Having less safe fingerprints increases the chance of possibly finding and changing the wrong methods in future versions. This can lead to the patch succeeding but the app crashing at runtime. (Patch result will become less trustworthy) Not really the behaviour I like. But again, personal preference.

opcodes dont make the fingerprint more safe. they make it less safe

epicsampler:
> opcodes dont make the fingerprint more safe. they make it less safe

they make it more safe because they decrease the chance of wrong methods being found (in case the current strings will be used in other places in future versions)

oSumAtrIX:
> they make i more safe 

They also make it less safe because the fingerprint can possibly not resolve.

epicsampler:
> They also make it less safe because the fingerprint can possibly not resolve.

this intentional breakage is exactly what i mean by "more safe"

your definition of "more safe": having a fingerprint that has a high chance of resolving in future versions (even if something changes and possibly breaks the app at runtime)

my definition of "more safe": having fingerprints that intentionally break if something chances to decrease the chance of patching in the wrong places but actively forcing someone to look into it to make sure everything is still correct (decreases the chance of patch saying it was successfully while failing at runtime -> patch results are more trustworthy)

oSumAtrIX:
> my definition of "more safe": having fingerprints that intentionally break if something chances

In that case, the opcode pattern has to capture what has to be modified. It is not the fingerprint's responsibility to safely patch the app, its the patches responsibility, note, that the strings do not return a scan result, so the patch is dynamically doing its thing in the method, adding a pattern means, that the pattern could change, but the patch does still work, which lead you to fix the fingerprint even though the patch still works.

epicsampler:
> It is not the fingerprint's responsibility to safely patch the app, its the patches responsibility.

They work together, no reason fingerprint should find possibly unrelated stuff

> so the patch is dynamically doing its thing in the method

right and having more safe fingerprints is one way to make sure that "the method" is the correct one and not something else that also matched the fingerprint by accident

> adding a pattern means, that the pattern could change, but the patch does still work, which lead you to fix the fingerprint even though the patch still works.

i like this - at least for now since patcher does (afaik) not scan for duplicates
forces devs to make sure everything is still correct and unique
but also can be annoying

---

another, better approach could be to let the patcher look for duplicate results of fingerprints and fail if they are not unique
this allows making fingerprints less "safe" while still making sure they find only one unique thing

oSumAtrIX:
> duplicate results of fingerprints

This shouldn't run at the patching process, but at development. So the dev can check for dupes but in production its fast.

epicsampler:
> This shouldn't run at the patching process, but at development. So the dev can check for dupes but in production its fast.

makes sense

patcher should support scanning for duplicates
but make it optional and disabled in production mode by default

cli could have this duplicate scanning disabled when patching supported versions
but have duplicate scanning enabled when `--experimental` is used

oSumAtrIX:
> cli could have this duplicate scanning disabled when patching supported versions but have duplicate scanning enabled when `--experimental` is used

How about a `--warn` switch? The option `--experimental` is for disabling the version checks. Please open an issue for this so we can track it.

bogadana:
I half made this but It's hard to get fingerprint results from the cli

bogadana:
Either the logs would have to come from patcher which is apparently not allowed or the whole system would have to be changed

oSumAtrIX:
Actually, the utility method which scans the classes for the fingerprinted method can simply return a result of the scan, which then the patcher can yield to the logger.

bogadana:
then the patcher is logging which is 'not its job'

oSumAtrIX:
The patcher instance is holding a patcher options instance with a reference to a logger for this purpose. It is also being used by the patcher.

bogadana:
last time you told me it doesnt belong in patcher 🤷

oSumAtrIX:
> last time you told me it doesnt belong in patcher 🤷

No, but this is offtopic.

PR #246, opened 2022-07-28
refactor: do not account for order in `VideoIdPatch.injectCall`

oSumAtrIX:
## Description

This PR gets rid of the offset when injecting calls. This means that calls will be inserted in reverse order. 

## Motivation

Patches do not depend on each other when they use the VideoId patch as a dependency, hence the order of these calls can be removed. This should be reviewed by @j4k0xb.

### DIFF
--- a/src/main/kotlin/app/revanced/patches/youtube/misc/videoid/patch/VideoIdPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/youtube/misc/videoid/patch/VideoIdPatch.kt
@@ -39,10 +39,11 @@ class VideoIdPatch : BytecodePatch(
     }
 
     companion object {
-        private lateinit var result: MethodFingerprintResult
+        // move-result-object offset
+        private const val offset = 2
         private var videoIdRegister: Int = 0
+        private lateinit var result: MethodFingerprintResult
         private lateinit var insertMethod: MutableMethod
-        private var offset = 2
 
         /**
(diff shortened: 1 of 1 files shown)


Sculas:
@j4k0xb:
> e.g. A and B depend on video id without offset adjusts its gonna inject
> 
> invoke-static B.newVideo
> invoke-static A.newVideo
> invoke-static VideoInformation.setNewVideo
> 
> now A accesses VideoInformation.channelName or smth but its wrong because not updated yet
> 
> order doesnt matter between patches (A/B) but VideoInformation should be called first

Closing.

src/main/kotlin/app/revanced/patches/youtube/misc/videoid/patch/VideoIdPatch.kt
```diff
@@ -35,12 +35,14 @@ class VideoIdPatch : BytecodePatch(
...
 
+        // offset so setCurrentVideoId is called before any injected call
+        offset++
+
         return PatchResultSuccess()
     }
 
     companion object {
-        // move-result-object offset
-        private const val offset = 2
```
Sculas: Want to add this comment back?

oSumAtrIX: No, this has been moved to the correct location somewhere below.

PR #251, opened 2022-07-30
fix: missing intent crashing

Canny1913:


src/main/kotlin/app/revanced/patches/youtube/misc/integrations/patch/IntegrationsPatch.kt
```diff
@@ -60,6 +73,15 @@ class IntegrationsPatch : BytecodePatch(
                 )
             ).toMutable()
         )
+
+        val serviceIndex = serviceMethod.implementation!!.instructions.lastIndex
+        serviceMethod.addInstructions(
+            serviceIndex, """
+               invoke-static {}, Lcom/google/android/apps/youtube/app/YouTubeTikTokRoot_Application;->getAppContext()Landroid/content/Context;
```
TheJeterLP, line 80: That makes no sense here. ReVancedUtils->getContext() returns getAppContext() so I don't see how this would work?

Canny1913: uh this instruction returns the context itself so i can  set it in ReVancedUtils

TheJeterLP: Oh ya nvm, I'm dumb lol

src/main/kotlin/app/revanced/patches/youtube/misc/integrations/patch/IntegrationsPatch.kt
```diff
@@ -60,6 +73,15 @@ class IntegrationsPatch : BytecodePatch(
...
+               move-result-object p0
```
Sculas, line 81: Is `p0` safe for use here? Just making sure :)

Canny1913: its not a "fresh" register but since its being overwritten it should be fine
i should've got the register dynamically tho

Sculas: You don't need to get it dynamically per see. I was asking if you're not overriding a register that's being used after without setting a value first, indirectly causing a use-after-free.

Canny1913: keep in mind that PR is still a draft tho

Sculas: I'm aware.

PR #278, opened 2022-08-02
feat: custom-branding-music patch

Lyceris-chan:
I changed the custom-branding patch to work with YouTube music using the same icons currently used for YouTube, I also moved the icons used for the custom-branding patch for YouTube to its own folder so we can more easily support custom-branding for other applications in the future if need be.

Tested on:
YouTube Music 5.16.51
YouTube 17.29.34


oSumAtrIX (changes requested):
The branding patch for YouTube is the same, you should create a shared dependency patch that works for both patches to get rid of duplicate code.

Lyceris-chan:
@oSumAtrIX can you give an example, I've personally never really worked on anything like this.

oSumAtrIX:
There's two patches that have basically identical code, you can create a shared object class, then use its functions in both patches.

Lyceris-chan:
I have no experience in kotlin so someone else would have to do so

oSumAtrIX:
Please create a new PR when you are ready. Thanks!

PR #285, opened 2022-08-03
feat: `swipe controls` override volume button behaviour

### LINKED ISSUE #279: feat: Unlock Cricbuzz Premium
### Application

Cricbuzz ( https://play.google.com/store/apps/details?id=com.cricbuzz.android )

### Issue

Unlock Cricbuzz Premium 

### Patch

To unlock the cricbuzz premium features. ( Removes ads + Unlock premium articles ) 

And also to add option to change theme as per system settings. 

### Motivation

Cricbuzz is one of the most popular sports app for Cricket. But they have started bombarding ads. Most of the time misleading and to some mild NSFW ads. Not only it ruins the whole experience but it makes the app less friendly for younger cricket enthusiasts.  

Their premium subscription isn't value for money and merely removes the ads and unlock few other articles to read. 

Also, though you can switch to dark mode in the app settings but there is not option of theme changing as per system settings. 

### Acknowledgements

shadow578:
closes https://github.com/revanced/revanced-patches/issues/279

needs to be merged together with https://github.com/revanced/revanced-integrations/pull/114

### DIFF
--- a/src/main/kotlin/app/revanced/extensions/Extensions.kt
+++ b/src/main/kotlin/app/revanced/extensions/Extensions.kt
@@ -37,7 +37,7 @@ internal fun MutableMethodImplementation.injectHideCall(
  */
 fun BytecodeData.traverseClassHierarchy(targetClass: MutableClass, callback: MutableClass.() -> Unit) {
     callback(targetClass)
-    this.findClass(targetClass.superclass ?: "")?.resolve()?.let {
+    this.findClass(targetClass.superclass ?: return)?.resolve()?.let {
         traverseClassHierarchy(it, callback)
     }
 }
--- a/src/main/kotlin/app/revanced/extensions/Extensions.kt
+++ b/src/main/kotlin/app/revanced/extensions/Extensions.kt
@@ -1,8 +1,10 @@
 package app.revanced.extensions
 
+import app.revanced.patcher.data.impl.BytecodeData
 import app.revanced.patcher.data.impl.ResourceData
 import app.revanced.patcher.extensions.addInstructions
 import app.revanced.patcher.patch.PatchResultError
+import app.revanced.patcher.util.proxy.mutableTypes.MutableClass
 import app.revanced.patcher.util.proxy.mutableTypes.MutableMethod
 import app.revanced.patcher.util.smali.toInstruction
 import org.jf.dexlib2.AccessFlags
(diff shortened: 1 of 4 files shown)


src/main/kotlin/app/revanced/extensions/Extensions.kt
```diff
@@ -27,6 +29,30 @@ internal fun MutableMethodImplementation.injectHideCall(
...
 
+/**
+ * traverse the class hierarchy starting from the given root class
+ *
+ * @param targetClass the class to start traversing the class hierarchy from
+ * @param callback function that is called for every class in the hierarchy
+ */
+fun BytecodeData.traverseClassHierarchy(targetClass: MutableClass, callback: MutableClass.() -> Unit) {
+    callback(targetClass)
+    this.findClass(targetClass.superclass ?: "")?.resolve()?.let {
```
oSumAtrIX, line 40: Is resolving the proxy required? Resolve creates a mutable clone of ImmutableClass. You should rely on the immutable type as long as you do not need mutability.

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/extensions/Extensions.kt
shadow578: idea behind this is that, if you're traversing the class hierarchy, you'll probably also want to mutate the classes in some way. 

Changing the function to 
```kotlin
fun BytecodeData.traverseClassHierarchy(targetClass: ClassDef, callback: ClassDef.() -> Unit) {
    targetClass.callback()
    this.findClass(targetClass.superclass ?: "")?.immutableClass?.let {
        traverseClassHierarchy(it, callback)
    }
}
```
and resolving in the callback does not work, as the changes done to the mutable class (superclass is changed) are not reflected in `immutableClass`.

oSumAtrIX: @Sculas, please check if this is really necessary.

shadow578:
while trying to implement the changes requested, i found that it didn't actually work quite right (final flag on methods wasn't cleared correctly for whatever reason).

---

What the `traverseClassHierarchy` is used for is to remove the `final` modifier on all classes (and their methods) in all superclasses of a class, so that the injected class can overwrite them.
Maybe there is a better way to achive this, but i couldn't find one...

src/main/kotlin/app/revanced/extensions/Extensions.kt
Sculas, line 40: Finding a class named `""` is going to get you nowhere.
```suggestion
    this.findClass(targetClass.superclass ?: return)?.resolve()?.let {
```

Sculas (changes requested):

src/main/kotlin/app/revanced/extensions/Extensions.kt
oSumAtrIX: @Sculas Please also review this

Sculas: I'll look into it.

oSumAtrIX:
Thanks

PR #287, opened 2022-08-03
feat: remove `force-vp9-codec` patch

epicsampler:
This functionality was useful some years ago but time goes on.

Stock has following logic: it uses whatever the best video codec is that your device has a hardware decoder for. Years ago some devices did not have a vp9 decoder so this feature helped to get those to play vp9 anyway (software decoded).

However nowadays basically every phone can decode vp9 in hardware. So vp9 gets used anyway, even without the patch.

Btw. the patch emulates a Samsung Galaxy S6 to fake a vp9 decoder.
This phone is from 2015!

So vp9 decoders came up ~7 years ago.
Means this patch only makes sense for cheaper very old phones.
But realistically speaking: who still uses such old phones?

Even if someone does:
Decoding vp9 on those old devices is not fun. It has to be done in software (on cpu) which drains the battery very fast (who would want that).

After all that considered this patch should be safe to remove.


oSumAtrIX:
Thanks, it is indeed useless and caused more issues than fixed.

epicsampler:
btw. the setting for this is still in `revanced_prefs.xml`. I intentionally not removed it to not conflict with new settings framework pr so this will solve itself

PR #291, opened 2022-08-04
fix swipe control defaults

LoV432:
swipe controls are on by default [here](https://github.com/revanced/revanced-integrations/blob/2a6d401df5ebfabba8fee3966d66e8905500652a/app/src/main/java/app/revanced/integrations/settings/SettingsEnum.java)


oSumAtrIX:
This is expected, if you include the patch you rather want it to be enabled than having to additionally enable it.

LoV432:
my bad....should have explained it completely....the issue is that the UI shows it as off by default and that is what i actually fixed

oSumAtrIX:
Please open a PR in the `settings-framework` branch as this will replace the current file.

PR #307, opened 2022-08-05
feat: music patches for 5.17.51

ghost:
happy weekend

### DIFF
--- a/src/main/kotlin/app/revanced/patches/music/audio/codecs/patch/CodecsUnlockPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/music/audio/codecs/patch/CodecsUnlockPatch.kt
@@ -26,12 +26,12 @@ class CodecsUnlockPatch : BytecodePatch(
     )
 ) {
     override fun execute(data: BytecodeData): PatchResult {
-        var result = CodecsLockFingerprint.result!!
+        val codecsLockResult = CodecsLockFingerprint.result!!
 
-        val implementation = result.mutableMethod.implementation!!
+        val implementation = codecsLockResult.mutableMethod.implementation!!
 
-        val scanResultStartIndex = result.patternScanResult!!.startIndex
-        val instructionIndex = scanResultStartIndex  +
+        val scanResultStartIndex = codecsLockResult.patternScanResult!!.startIndex
+        val instructionIndex = scanResultStartIndex +
                 if (implementation.instructions[scanResultStartIndex - 1].opcode == Opcode.CHECK_CAST) {
                     // for 5.16.xx and lower
                     -3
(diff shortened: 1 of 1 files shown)


src/main/kotlin/app/revanced/patches/music/audio/codecs/fingerprints/CodecsLockFingerprint.kt
```diff
@@ -12,17 +12,14 @@ import org.jf.dexlib2.Opcode
...
 )
 @FuzzyPatternScanMethod(2) // FIXME: Test this threshold and find the best value.
 @CodecsUnlockCompatibility
 @Version("0.0.1")
 object CodecsLockFingerprint : MethodFingerprint(
-    "L", AccessFlags.PUBLIC or AccessFlags.STATIC, listOf("L", "L", "L", "L"), listOf(
-        Opcode.INVOKE_VIRTUAL,
-        Opcode.MOVE_RESULT_OBJECT,
-        Opcode.CHECK_CAST,
+    "L", AccessFlags.PUBLIC or AccessFlags.STATIC, null, listOf(
```
Sculas, line 22: Does this fingerprint still work on older versions? Just making sure.
If not, remove compatibility for older versions or make it a new/separate fingerprint.

epicsampler: > Does this fingerprint still work on older versions
yes thats why parameters and some opcodes are removed so it works in both

Sculas: Alright!

Sculas (approved):

src/main/kotlin/app/revanced/patches/music/audio/codecs/patch/CodecsUnlockPatch.kt
```diff
@@ -29,7 +30,13 @@ class CodecsUnlockPatch : BytecodePatch(
 
         val implementation = result.mutableMethod.implementation!!
 
-        val instructionIndex = result.patternScanResult!!.startIndex
+        val instructionIndex = if (implementation.instructions[result.patternScanResult!!.startIndex - 1].opcode == Opcode.CHECK_CAST) {
```
oSumAtrIX, line 33: To avoid duplicate code, use the if case later

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/music/audio/codecs/patch/CodecsUnlockPatch.kt
```diff
@@ -29,7 +30,14 @@ class CodecsUnlockPatch : BytecodePatch(
 
         val implementation = result.mutableMethod.implementation!!
 
-        val instructionIndex = result.patternScanResult!!.startIndex
+        val instructionIndex = result.patternScanResult!!.startIndex +
+                if (implementation.instructions[result.patternScanResult!!.startIndex - 1].opcode == Opcode.CHECK_CAST) {
```
oSumAtrIX, line 34:
```suggestion
        val scanResultStartIndex = result.patternScanResult!!.startIndex
        val instructionIndex = scanResultStartIndex  +
                if (implementation.instructions[scanResultStartIndex - 1].opcode == Opcode.CHECK_CAST) {
```

oSumAtrIX: Also below at line 44 this can be changed too

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/music/audio/codecs/patch/CodecsUnlockPatch.kt
epicsampler: makes no sense for line 44 since this is on another scan result and only used once
however refactor is coming to make this more clear

PR #315, opened 2022-08-07
fix: WarnWetter failing to register push registration

### LINKED ISSUE #314: bug: WarnWetter doesn't send push notifications
### Type

Error at runtime

### Bug description

After opening the app and going through the setup process, a popup tells you that *Push registration failed*... it's results in the same issue without any patches applied.

### Steps to reproduce

Patch the app without mounting.

### Relevant log output

```shell
W/Firebase-Installations: Error when communicating with the Firebase Installations server API. HTTP response: [403 Bad Request: {
  "error": {
    "code": 403,
    "message": "Request from this Android client application de.dwd.warnapp are blocked.",
    "status": "PERMISSION_DENIED",
    "details": [
      {
        "@type": "type.googleapis.com/google.rpc.ErrorInfo",
        "reason": "API_KEY_ANDROID_APP_BLOCKED",
        "domain": "googleapis.com",
        "metadata": [
          {
                ...
          }
        ]
      }
    ]
  }
}
]
```

### Screenshots or videos

image

### Solution

Spoof the keystore hash to always send back the hash of the original app.

### Additional context

I'll be fixing this, this is just an issue for discussion and for acknowledgement of bug.

Itroublve:
Closes #314, spoofs the `X-Android-Cert` header in order to allow push notifications to be sent to the patched app.

### DIFF
--- a/src/main/kotlin/app/revanced/patches/warnwetter/misc/firebasegetcert/patch/FirebaseGetCertPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/warnwetter/misc/firebasegetcert/patch/FirebaseGetCertPatch.kt
@@ -1,0 +1,45 @@
+package app.revanced.patches.warnwetter.misc.firebasegetcert.patch
+
+import app.revanced.patcher.annotation.Description
+import app.revanced.patcher.annotation.Name
+import app.revanced.patcher.annotation.Version
+import app.revanced.patcher.data.impl.BytecodeData
+import app.revanced.patcher.extensions.addInstructions
+import app.revanced.patcher.patch.PatchResult
+import app.revanced.patcher.patch.PatchResultSuccess
+import app.revanced.patcher.patch.impl.BytecodePatch
+import app.revanced.patches.warnwetter.misc.firebasegetcert.fingerprints.*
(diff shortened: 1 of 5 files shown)


src/main/kotlin/app/revanced/patches/warnwetter/misc/firebasegetcert/patch/FirebaseGetCertPatch.kt
```diff
@@ -0,0 +1,45 @@
...
+import app.revanced.patcher.data.impl.BytecodeData
+import app.revanced.patcher.extensions.addInstructions
+import app.revanced.patcher.patch.PatchResult
+import app.revanced.patcher.patch.PatchResultSuccess
+import app.revanced.patcher.patch.impl.BytecodePatch
+import app.revanced.patches.warnwetter.misc.firebasegetcert.fingerprints.*
+import app.revanced.patches.warnwetter.misc.firebasegetcert.annotations.FirebaseGetCertPatchCompatibility
+
+@Name("spoof-cert-patch")
+@Description("Spoofs the X-Android-Cert header")
```
oSumAtrIX, line 15:
```suggestion
@Description("Spoofs the X-Android-Cert header.")
```

oSumAtrIX (changes requested):

oSumAtrIX (approved):

PR #331, opened 2022-08-13
fix: gitignore .vscode/

Technikte:
Add .vscode/ folder to gitignore for all vscode user


oSumAtrIX:
Do you actually not want to commit the vsc folder?

Technikte:
I dont think this is a good idea since the work env for everyone can differ quite a bit..

oSumAtrIX:
What about run or debug configurations?

PR #332, opened 2022-08-13
feat: add pflotsh ECMWF app patch

Technikte:
Add a patch for pflotsh ecmwf app, after which no intercepts for subscription should occur. It worked before but a the moment I can not really test the patch live..(see discord for reason) because of that for now only a draft.
At some point I will try to make the fingerprint a bit more robust..
https://pflotsh.com/de/ecmwf.html


Technikte:
I have tested on my device and patch seems to work like intended. Ready for review.

src/main/kotlin/app/revanced/patches/ecmwf/misc/subscription/fingerprints/SubscriptionUnlockFingerprint.kt
```diff
@@ -0,0 +1,24 @@
...
+@SubscriptionUnlockCompatibility
+@Version("0.0.1")
+object SubscriptionUnlockFingerprint : MethodFingerprint(
+    "Z",
+    null,
+    null,
+    null,
+    null,
+    { methodDef ->
+        methodDef.definingClass.endsWith("MapsActivity;") && methodDef.name == "t0"
```
oSumAtrIX, line 22: You constrained the patch compatibility to one version. Hence `t0` would always work, but you should rather make this dynamic with a proper pattern or different fingerprint so the patch works with future versions too.

src/main/kotlin/app/revanced/patches/ecmwf/misc/subscription/patch/SubscriptionUnlockPatch.kt
```diff
@@ -0,0 +1,47 @@
...
+        val method = result.mutableMethod
+
+        val index = method.implementation!!.instructions.size
+
+        // remove R() at 10212
+        method.removeInstruction(index - 3)
+        // remove R() at 10206
+        method.removeInstruction(index - 5)
+
+        val insertIndex = method.implementation!!.instructions.count()
```
oSumAtrIX, line 37:
```suggestion
        val insertIndex = index
```

src/main/kotlin/app/revanced/patches/ecmwf/misc/subscription/patch/SubscriptionUnlockPatch.kt
```diff
@@ -0,0 +1,47 @@
...
+
+        method.addInstructions(
+            (insertIndex - 1),
```
oSumAtrIX, line 40:
```suggestion
            insertIndex - 1,
```

src/main/kotlin/app/revanced/patches/ecmwf/misc/subscription/fingerprints/SubscriptionUnlockFingerprint.kt
Technikte: Of course I will make the fingerprint in the future more robust, since the app dont get often updates (1-2 per year) it sould be fine for now.

src/main/kotlin/app/revanced/patches/ecmwf/misc/subscription/patch/SubscriptionUnlockPatch.kt
Technikte: done

src/main/kotlin/app/revanced/patches/ecmwf/misc/subscription/patch/SubscriptionUnlockPatch.kt
Technikte: done

src/main/kotlin/app/revanced/patches/ecmwf/misc/subscription/patch/SubscriptionUnlockPatch.kt
```diff
@@ -0,0 +1,47 @@
...
+import app.revanced.patcher.extensions.removeInstruction
+import app.revanced.patcher.patch.annotations.Patch
+import app.revanced.patcher.patch.impl.BytecodePatch
+import app.revanced.patcher.patch.PatchResult
+import app.revanced.patcher.patch.PatchResultSuccess
+import app.revanced.patches.ecmwf.misc.subscription.annotations.SubscriptionUnlockCompatibility
+import app.revanced.patches.ecmwf.misc.subscription.fingerprints.SubscriptionUnlockFingerprint
+
+@Patch
+@Name("subscription-unlock")
```
oSumAtrIX, line 17:
```suggestion
@Name("pflotsh-ecmwf-subscription-unlock")
```

oSumAtrIX (changes requested):

PR #359, opened 2022-08-19
feat: bundle dex file into jar file

ponces:
This should work as a cross-platform solution both for desktop and mobile!

It basically inserts the dex file inside the jar file so that it can be used on desktop with the JarPatchBundle (as it is now) and for mobile with the DexPatchBundle. Using only the dex file in mobile was not possible due to missing resources on the dex file.

Merging both is the solution :)

Semantic release file was also updated as after merging this, the jar asset is enough for both platforms.

### DIFF
--- a/.releaserc
+++ b/.releaserc
@@ -31,9 +31,6 @@
         "assets": [
           {
             "path": "build/libs/*.jar"
-          },
-          {
-            "path": "build/libs/*.dex"
           }
         ]
       }
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -27,15 +27,14 @@ dependencies {
 }
 
 tasks {
-    register<DefaultTask>("generateDex") {
-        description = "Generate dex files from build"
+    register<DefaultTask>("generateBundle") {
+        description = "Generate dex files from build and bundle them in the jar file"
         dependsOn(build)
 
         doLast {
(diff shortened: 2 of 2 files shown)


Sculas:
Even though this is out of scope for this PR, we should really make a Gradle plugin for this so it can be reused in other projects in the near future.

Sculas (approved):

oSumAtrIX:
> Even though this is out of scope for this PR, we should really make a Gradle plugin for this so it can be reused in other projects in the near future.

What do you mean in other projects?

Sculas:
Other developers that may want to make their own patches repository. Since you can add repositories in the manager, we should account for that.

oSumAtrIX:
I was thinking of creating a template repository like [this](https://github.com/Aliucord/plugins-template). That way people could "fork" the template and simply create their own patches, for now this is fine though.

Sculas:
That could also work! But I still think making it a proper Gradle plugin that is not platform agnostic is much better. (remind you, the current one **only** works on the GitHub CI and Linux with the Android SDK!)

Anyway, thanks @ponces for your contribution! ❤️

oSumAtrIX:
> That could also work! But I still think making it a proper Gradle plugin that is not platform agnostic is much better. 

Yes, both, a template and the plugin should be made.

PR #366, opened 2022-08-21
fix: make `custom-branding` cross-platform

ponces:
InputStream.readAllBytes was only introduced [on Android in API 33](https://developer.android.com/reference/java/io/InputStream#readAllBytes()) so we need an implementation for a wider adoption :)


Sculas:
Could this be made an extension function instead? (in a separate PR, of course)

ponces:
> Could this be made an extension function instead? (in a separate PR, of course)

You mean all this code, including `getIconStream`?

Sculas:
Not the entire function, just the part for `readAllBytes`. So if we need to `readAllBytes`, we can reuse that code.

ponces:
> Not the entire function, just the part for `readAllBytes`. So if we need to `readAllBytes` anywhere else, we can reuse that code.

Ah, understood! Great idea! Will do!

PR #401, opened 2022-08-29
[seekbar-tapping] Fix always effect although be disabled in settings.

d4rkk3y:


Canny1913:
Does this fix break the patch for old versions?

d4rkk3y:
> Does this fix break the patch for old versions?

which version should I check? I checked the 17.32.35

Canny1913:
Check 17.29.34.

d4rkk3y:
Tested in 17.19.36,  17.29.34, 17.32.35. It work.

src/main/kotlin/app/revanced/patches/youtube/interaction/seekbar/patch/EnableSeekbarTappingPatch.kt
```diff
@@ -85,6 +85,7 @@ class EnableSeekbarTappingPatch : BytecodePatch(
         val register = (instruction as Instruction35c).registerC
 
         // the instructions are written in reverse order.
+        val elseLabel = implementation.newLabelForIndex(result.patternScanResult!!.endIndex + 1)
```
oSumAtrIX, line 88: Please fix the comment, it should be below the instruction

oSumAtrIX (changes requested):

PR #433, opened 2022-09-05
feat: make use of tags annotation.

Aunali321:
Uses `tags` annotation introduced [here](https://github.com/revanced/revanced-patcher/pull/96).


oSumAtrIX:
Tags are not categories, which they are currently used as in this PR.

oSumAtrIX:
I don't see where you said it is okay to push to the README file?

PR #438, opened 2022-09-07
Renamed "YouTube Music Revanced" to "YT Music Revanced"

decipher3114:
Nothing special, just renamed that..


oSumAtrIX:
The name should not be abbreviated. A patch option will be able to change the default name.

PR #434, opened 2022-09-06
update the `download` icon to fit in with the rest

Ushie:
Screenshot_20220906-115940_YouTube

PR #464, opened 2022-09-11
fix: change download icon to match other icons

afnzmn:


oSumAtrIX:
Whats the difference to #434?

afnzmn:
> Whats the difference to #434?

The PR isn't opened by Ushie.
Also, I remade the icon so its slightly different from that of YouTube's.

oSumAtrIX:
Why should this PR be merged instead of #464?

afnzmn:
In #434 you question whether the icon Ushie took from YouTube complies with GPLv3 license.

PR #476, opened 2022-09-14
chore: remove ReVanced from YouTube Music label as the name is too long already

AMSAMSAMSAMS:
remove ReVanced from YouTube Music as the name is too long already


oSumAtrIX:
Then there's no difference between the patched and unpatched app.

PR #478, opened 2022-09-14
Prettified json

decipher3114:


oSumAtrIX:
The JSON is auto-generated and will be overwritten when the workflow runs and builds the patches. Instead, a json serialization option should exist and be used to pretty-print the JSON.

decipher3114:
> The JSON is auto-generated and will be overwritten when the workflow runs and builds the patches. Instead, a json serialization option should exist and be used to pretty-print the JSON.

I tried doing that in Generator.kt file inside Meta.

I added setPrettyPrinting() in the val gson. but no success after build.

PR #487, opened 2022-09-15
feat: `MaterialYou` Patch

datnerdguy:


oSumAtrIX:
Theres no need to create a separate patch for new themes, simply add them to the Themes enum.

decipher3114:
Add this beside the amoled patch and add its option in theme

datnerdguy:
> Theres no need to create a separate patch for new themes, simply add them to the Themes enum.

Ok sir

PR #489, opened 2022-09-15
Specify latest working Twitter target in readme

krokofant:
Related ReVanced/revanced-patches-template#2245

The twitter patch seems to have been broken for a while and having `target: all` in the readme is confusing.


oSumAtrIX:
The file is autogenerated, also the patch targets all versions.

krokofant:
@oSumAtrIX Alright! 😅

PR #501, opened 2022-09-16
feat(tiktok): `tiktok-feed-filter` , `tiktok-settings` and `tiktok-force-login`

### LINKED ISSUE #417: feat: Vanced MicroG for Discord
### Application

Discord

### Issue

This patch would solve the issue that people who don't have Google Play Services cannot receive any Discord notifications.

### Patch

A patch like the one for YouTube, where MicroG integration is added so users can receive notification, loggning in with Google is not needed on Discord.

However, I do not know if Vanced MicroG would work with any other apps...

### Motivation

There are a lot of people with for example LineageOS or Huawei phones that don't have Google services, Discord only provides notifications from Google, there also are no Discord app alternatives for Android that are fully featured

### Acknowledgements

d4rkk3y:
I don't know if it is able to implement settings like youtube for tiktok since tiktok resource can not decode and rebuild with apktool. So in this PR I use a trivial ways to enable hideLive filter.
This PR also possibly fix https://github.com/revanced/revanced-patches/issues/417

RE: https://github.com/revanced/revanced-integrations/pull/142

### DIFF
--- a/src/main/kotlin/app/revanced/patches/tiktok/feedfilter/annotations/FeedFilterCompatibility.kt
+++ b/src/main/kotlin/app/revanced/patches/tiktok/feedfilter/annotations/FeedFilterCompatibility.kt
@@ -1,0 +1,14 @@
+package app.revanced.patches.tiktok.feedfilter.annotations
+
+import app.revanced.patcher.annotation.Compatibility
+import app.revanced.patcher.annotation.Package
+
+@Compatibility(
+    [
+        Package("com.ss.android.ugc.trill"),
+        Package("com.zhiliaoapp.musically")
+    ]
+)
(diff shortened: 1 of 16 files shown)


oSumAtrIX:
Technically, resources don't have to be necessarily modified. UI can be changed programmatically too (notice the SponsorBlock od RYD settings screen). Though it's better to fix the issue with decoding and encoding resources.

d4rkk3y:
> Technically, resources don't have to be necessarily modified. UI can be changed programmatically too (notice the SponsorBlock od RYD settings screen). Though it's better to fix the issue with decoding and encoding resources.

I don't know if I can implement that but I will try.

oSumAtrIX (approved):

PR #568, opened 2022-09-21
feat: Rename YT Music

GeekCornerGH:
The current name is too long. This PR aims to fix it just like the official YT Music app does.


oSumAtrIX:
YT is an abbreviation and should not be used as the name.

GeekCornerGH:
> YT is an abbreviation and should not be used as the name.
@oSumAtrIX 
Official app is actually named "YT Music"

PR #579, opened 2022-09-21
fix: sponsorblock shorts blocker doesn't work properly

OxrxL:


src/main/kotlin/app/revanced/patches/youtube/layout/sponsorblock/bytecode/fingerprints/ShortsPlayerConstructorFingerprint.kt
```diff
@@ -33,6 +36,16 @@ object ShortsPlayerConstructorFingerprint : MethodFingerprint(
...
+        Opcode.IPUT_OBJECT,
+        Opcode.IPUT_OBJECT,
+        Opcode.CONST_4
+    ),
+    customFingerprint = { methodDef ->
+        methodDef.implementation?.instructions?.any { instruction ->
+            (instruction as? WideLiteralInstruction)?.wideLiteral ==
+                    ResourceIdMappingProviderResourcePatch.resourceMappings.single {
+                        it.type == "id" && it.name == "reel_persistent_edu_button_group"
+                    }.id
```
oSumAtrIX, line 48: Keep in mind this will call `.single` every time the `customFingerprint´ function is executed. Please move it outside of the function.

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/youtube/layout/sponsorblock/bytecode/fingerprints/ShortsPlayerConstructorFingerprint.kt
```diff
@@ -33,6 +36,17 @@ object ShortsPlayerConstructorFingerprint : MethodFingerprint(
...
-        Opcode.IPUT_OBJECT
-    )
+        Opcode.IPUT_OBJECT,
+        Opcode.IPUT_OBJECT,
+        Opcode.CONST_4
+    ),
+    customFingerprint = { methodDef ->
+        val targetResourceId = ResourceIdMappingProviderResourcePatch.resourceMappings.single {
+            it.type == "id" && it.name == "reel_persistent_edu_button_group"
+        }.id
```
oSumAtrIX, line 46: This will still call `.single` every time the customFingerprint is being invoked. Move this outside of customFingerprint.

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/youtube/layout/sponsorblock/bytecode/fingerprints/ShortsPlayerConstructorFingerprint.kt
OxrxL: > This will still call `.single` every time the customFingerprint is being invoked. Move this outside of customFingerprint.

Please...When you make a request try to be clearer, because otherwise I do what you say but it doesn't work.

OxrxL: Ok...trying here and there blindly I managed to do what you asked for. Hope it's okay now :)

oSumAtrIX: Note, I previously said that every time the `customFingerprint´ function is executed `.single` will be called and asked you to move it outside of that particular function.

Since the patch dependencies are being executed first, then the fingerprints are resolved, and lastly, the execute method of the patch is being called, it is necessary to previously get the resource id. This can be done by creating a dependency patch that holds the value. I will push and fix it so you can see how it needs to be done.

src/main/kotlin/app/revanced/patches/youtube/layout/sponsorblock/bytecode/fingerprints/ShortsPlayerConstructorFingerprint.kt
```diff
@@ -33,6 +36,18 @@ object ShortsPlayerConstructorFingerprint : MethodFingerprint(
...
+        methodDef.implementation?.instructions?.any { instruction ->
+            (instruction as? WideLiteralInstruction)?.wideLiteral ==
+                    targetResourceId
+        } == true
+    }
+)
+
+val targetResourceId = ResourceIdMappingProviderResourcePatch.resourceMappings.single {
+    it.type == "id" && it.name == "reel_persistent_edu_button_group"
+}.id
```
oSumAtrIX, line 53: This is dangerous, since it can not be guaranteed, that the `ResourceIdMappingProviderResourcePatch` has been executed before this code runs.

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/youtube/layout/sponsorblock/bytecode/fingerprints/ShortsPlayerConstructorFingerprint.kt
OxrxL: > Note, I previously said that every time the `customFingerprint´ function is executed `.single` will be called and asked you to move it outside of that particular function.
> 
> Since the patch dependencies are being executed first, then the fingerprints are resolved, and lastly, the execute method of the patch is being called, it is necessary to previously get the resource id. This can be done by creating a dependency patch that holds the value. I will push and fix it so you can see how it needs to be done.

Thanks...I'll be waiting for your changes so. Sometimes how the patcher works it's dark to me. 👍

PR #585, opened 2022-09-22
fix(sponsorblock): fix wording of "privateID"

mchangrh:
The current wording of the "privateID/ userID" is causing revanced users to constantly leak their privateIDs in the SponsorBlock discord.

The point of this wording to change is to explicitly reinforce that this is not the public component, this is the private component.

### DIFF
--- a/src/main/resources/sponsorblock/host/values/strings.xml
+++ b/src/main/resources/sponsorblock/host/values/strings.xml
@@ -14,7 +14,7 @@
     <string name="general_adjusting_sum">This is the number of milliseconds you can move when you use the time adjustment buttons while adding new segment</string>
     <string name="general_min_duration">Minimum segment duration</string>
     <string name="general_min_duration_sum">Segments shorter than the set value (in seconds) will not be skipped or shown in the player</string>
-    <string name="general_uuid">Your Private UserID</string>
+    <string name="general_uuid">Your private user id</string>
     <string name="general_uuid_sum">This should be kept private. This is like a password and should not be shared with anyone. If someone has this, they can impersonate you</string>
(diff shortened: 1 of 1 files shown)


oSumAtrIX:
I remember SponsorBlock referring to it as public id somewhere. Is it really private?

Canny1913:
> I remember SponsorBlock referring to it as public id somewhere. Is it really private?

I don't remember that being referred anywhere.

oSumAtrIX:
image

German for "Public".

mchangrh:
> I remember SponsorBlock referring to it as public id somewhere. Is it really private?

The "UUID" that the string describes is the privateID components

image

the public component is not displayed in reVanced except in the username

mchangrh:
the publicID is derived from the publicID, if you copy your "UserID" from the extension options and paste it in that field, you'll get the same "username" which defaults to publicID unless explicitly set - if you instead copy your "username" you'll get a completely different "username"

oSumAtrIX:
I think it makes more sense to keep this the user id, the same as SponsorBlock refers to, but add the same notice to the UI.

mchangrh:
> I think it makes more sense to keep this the user id, the same as SponsorBlock refers to, but add the same notice to the UI.

The issue is that people are mistaking it for the public component, which can be shared freely. People are not reading past the title and I don't think an additional toast will suffice

oSumAtrIX:
A toast would not suffice it would need to be a notice just like SponsorBlock's.

mchangrh:
https://github.com/ajayyy/SponsorBlock/pull/1490 merged upstream in browser extension

oSumAtrIX:
it still refers to the id as private `userId`, not `privateId`.

PR #591, opened 2022-09-23
fix(hide-premium-navbar): remove the correct instructions

juztim:
LGTM


src/main/kotlin/app/revanced/patches/spotify/premium_navbar_tab/patch/PremiumNavbarTabPatch.kt
```diff
@@ -13,40 +13,48 @@ import app.revanced.patcher.patch.annotations.Patch
...
+        var removeAmount = 2
+        // 2nd const remove method
         for ((i, instruction) in methodInstructions.asReversed().withIndex()) {
             if (instruction.opcode.ordinal != Opcode.CONST.ordinal) continue
-            if ((instruction as OneRegisterInstruction).registerA != premiumTabId) continue
-            val constIdx = methodInstructions.size - i
-            val methodIdx = constIdx + 8
-            method.removeInstruction(methodIdx)
-            break
+            if ((instruction as WideLiteralInstruction).wideLiteral != premiumTabId) continue
```
Sculas, line 48: This cast might fail.
```suggestion
            if ((instruction as? WideLiteralInstruction)?.wideLiteral != premiumTabId) continue
```

src/main/kotlin/app/revanced/patches/spotify/premium_navbar_tab/patch/PremiumNavbarTabPatch.kt
```diff
@@ -13,40 +13,48 @@ import app.revanced.patcher.patch.annotations.Patch
...
+
+            val findThreshold = 10
+            val constIndex = lastInstructionIdx - i
+            val invokeInstruction = methodInstructions.subList(constIndex, constIndex + findThreshold).first {
+                it.opcode.ordinal == Opcode.INVOKE_VIRTUAL_RANGE.ordinal
```
Sculas, line 53: I don't think this is needed.
```suggestion
                it.opcode == Opcode.INVOKE_VIRTUAL_RANGE
```

src/main/kotlin/app/revanced/patches/spotify/premium_navbar_tab/patch/PremiumNavbarTabPatch.kt
```diff
@@ -13,40 +13,48 @@ import app.revanced.patcher.patch.annotations.Patch
...
+            }
+            method.removeInstruction(methodInstructions.indexOf(invokeInstruction))
```
Sculas, line 55: Why are you doing `indexOf`? Can't you just remove using the index?

Sculas (changes requested):
This PR was merged without a proper review. If someone has the time, please resolve these changes in a new PR.

juztim:
@Sculas I will make a new PR tomorrow with the changes you suggested

src/main/kotlin/app/revanced/patches/spotify/premium_navbar_tab/patch/PremiumNavbarTabPatch.kt
oSumAtrIX: No, removal has to be done via the implementation, not via the instruction list.

src/main/kotlin/app/revanced/patches/spotify/premium_navbar_tab/patch/PremiumNavbarTabPatch.kt
oSumAtrIX: Comparing the ordinal should be faster rather than comparing the entire enum object.

src/main/kotlin/app/revanced/patches/spotify/premium_navbar_tab/patch/PremiumNavbarTabPatch.kt
oSumAtrIX: The cast will never fail. Previous line of code checks for the opcode of the instruction.

PR #602, opened 2022-09-24
fix: improved hdr brightness fingerprint compatibility

OxrxL:


OxrxL:
Tried with 17.33.42 and latest 17.36.37.

src/main/kotlin/app/revanced/patches/youtube/misc/hdrbrightness/patch/HDRBrightnessPatch.kt
```diff
@@ -40,7 +41,10 @@ class HDRBrightnessPatch : BytecodePatch(
             )
         )
 
-        val method = HDRBrightnessFingerprint.result!!.mutableMethod
+        var result = HDRBrightnessFingerprint.result
+        if (result == null) result = HDRBrightnessLegacyFingerprint.result
```
oSumAtrIX, line 45: Is it possible to only rely on a single fingerprint which works for both versions?

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/youtube/misc/hdrbrightness/patch/HDRBrightnessPatch.kt
OxrxL: Unfortunately no. On versions prior to 17.33.42 the structure of the method changes completely.

src/main/kotlin/app/revanced/patches/youtube/misc/hdrbrightness/fingerprints/HDRBrightnessFingerprint.kt
```diff
@@ -14,5 +18,18 @@ import app.revanced.patches.youtube.misc.hdrbrightness.annotations.HDRBrightness
 @Version("0.0.1")
 object HDRBrightnessFingerprint : MethodFingerprint(
     "V",
-    strings = listOf("c.SettingNotFound;", "screen_brightness", "android.mediaview", "lux"),
+    customFingerprint = { methodDef ->
```
oSumAtrIX, line 21: The custom fingerprint is very complex, instead a parent fingerprint should be used as previously mentioned.

oSumAtrIX (changes requested):

OxrxL:
Fixed to reduce the unnecessary wide compatibility (which now it start from stable version 17.33.42)

PR #605, opened 2022-09-24
added input type in patch options

decipher3114:
I added type in patch options.
I also saw that patcher is dependent on it, But I don't have wnough knowledge to edit those, so please update that too.


oSumAtrIX:
The type is automatically infereable from the options toml file.

decipher3114:
Great then 👍

PR #644, opened 2022-09-27
fix: youtube crash when play a video.

d4rkk3y:
This PR fixes the issue since that make youtube crash when play a video.


oSumAtrIX:
Why is this change required suddenly?

d4rkk3y:
> Why is this change required suddenly?

Because this patch 
https://github.com/revanced/revanced-patches/blob/336a887b5523d57dd1b0be9715e6b6ac0da6fa45/src/main/kotlin/app/revanced/patches/youtube/layout/hidetimeandseekbar/patch/HideTimeAndSeekbarPatch.kt#L43-L55
and this patch
https://github.com/revanced/revanced-patches/blob/336a887b5523d57dd1b0be9715e6b6ac0da6fa45/src/main/kotlin/app/revanced/patches/youtube/layout/sponsorblock/bytecode/patch/SponsorBlockBytecodePatch.kt#L95-L105
If HideTimeAndSeekbar be executed before SponsorBlock. Then SponsorBlock will be patched at wrong position.

PR #654, opened 2022-09-28
fix: hide email crash

OxrxL:


oSumAtrIX (approved):

Sculas:
Has this been tested with older versions of YT? It seems that everyone forgets about the compatibility annotations.
I can guarantee you that tons of patches don't work with all the versions they advertise they work on.

OxrxL:
> I can guarantee you that tons of patches don't work with all the versions they advertise they work on.

Yeah...I know, and I can't assure the compatibility with too older versions.

So...I planned to increase minimum version to 17.26 or 17.30 for all patches.

oSumAtrIX:
The initial PRs offset was wrong, the current offsets should work til the first version the patch advertises compatibility for. The same can be guaranteed for other patches except for a couple, which either have to be removed or fixed.

OxrxL:
> The initial PRs offset was wrong, the current offsets should work til the first version the patch advertises compatibility for. The same can be guaranteed for other patches except for a couple, which either have to be removed or fixed.

Was wrong due to a my mistake...that's all.

But we really should reduce the compatible versions after a while, otherwise 10,000 versions to test drive us crazy ahahaha.

oSumAtrIX:
Testing shouldn't be necessary until a newer version breaks a patch, if they do, either drop support for older versions, or implement the patch to support the newer version in a dynamic wa, if possible (such as if an offset was hardcoded like in the current example).

OxrxL:
> Testing shouldn't be necessary until a newer version breaks a patch, if they do, either drop support for older versions, or implement the patch to support the newer version in a dynamic wa, if possible (such as if an offset was hardcoded like in the current example).

Also...yes. But I suggest to cut compatibility to versions older than 17.26.

Issue #301, opened 2022-08-05
bug: finishing general ads todo


OxrxL:
### Type

Other

### Bug description

General ads patch it is still unfinished, due to the presence of ToDo comments.

P.S: `promoted_video_item_full_bleed` patch still does not exist.

### Steps to reproduce

...

### Relevant log output

```shell
...
```

### Additional context

https://github.com/revanced/revanced-patches/blob/main/src/main/kotlin/app/revanced/patches/youtube/ad/general/bytecode/patch/GeneralBytecodeAdsPatch.kt

OxrxL:
https://github.com/revanced/revanced-patches/pull/641

OxrxL:
https://github.com/revanced/revanced-patches/commit/a04aeb598e807d550ffb624db8458dc0473247cf

Issue #295, opened 2022-08-04
bug: <ads toggles>


OxrxL:
### Type

Other

### Bug description

Some general ads patches don't have a settings toggle, but just a direct setvisibility reference.

### Steps to reproduce

...

### Relevant log output

```shell
...
```

### Solution

Change setvisibility with a reference to a dedicated method inside settings.

### Additional context

https://github.com/revanced/revanced-patches/blob/main/src/main/kotlin/app/revanced/patches/youtube/ad/general/bytecode/patch/GeneralBytecodeAdsPatch.kt

oSumAtrIX:
Please list all the patches which miss the toggle.

OxrxL:
> Please list all the patches which miss the toggle.

First comment updated with screenshot

OxrxL:
https://github.com/revanced/revanced-patches/pull/641

OxrxL:
https://github.com/revanced/revanced-patches/pull/641

Issue #167, opened 2022-07-12
problem: `old ads patches`


OxrxL:
## 🐞 Issue

From the moment that Google moved (and still moving) so many stuff to litho, seems that old ads patches with id "ad_attribution" are not useful anymore.

## ⚙ Reproduce

Excluding them from patcher shouldn't change anything.

## 🛠 Solution

Deprecate patches.

OxrxL:
https://github.com/revanced/revanced-patches/blob/main/src/main/kotlin/app/revanced/patches/youtube/ad/general/bytecode/patch/GeneralBytecodeAdsPatch.kt

```
 // list of resource names to get the id of
    private val resourceIds = arrayOf(
        "ad_attribution",
```

OxrxL:
https://github.com/revanced/revanced-patches/pull/641

Issue #108, opened 2022-07-03
Further ad patches cleanup


OxrxL:
Reel reference can be removed from this patch: https://github.com/revanced/revanced-patches/blob/main/src/main/kotlin/app/revanced/patches/youtube/ad/general/bytecode/patch/GeneralBytecodeAdsPatch.kt

because has been moved to litho.

OxrxL:
@oSumAtrIX

TheJeterLP:
> @oSumAtrIX 

No need to ping Team members, we will see it :)

OxrxL:
> > @oSumAtrIX
> 
> No need to ping Team members, we will see it :)

Sorry. I did this as a result of a private conversation (related to the the issue) with the aformentioned team member.

OxrxL:
https://github.com/revanced/revanced-patches/pull/641

OxrxL:
Closed because the patch is still useful (the component is still visible, and can be hidden).

PR #687, opened 2022-10-01
fix: refactor general ads patch

### LINKED ISSUE #326: feat: extended litho elements block
### Type

Cosmetic

### Issue

The current litho view blockage is incomplete, and based solely on container-type templates.

To be more specific:

Each litho view is composed of multiple elements. For example: a `buttons bar` is a container, inside which there will be other containers of type `button`, which in turn will contains other elements of type `image icon` and `textbox`.

Well...following the aforementioned example, we are currently able to block just the `buttons bar` and not the child elements (entire button or its icon/textbox).

### Feature

The same bytecode method related to ads blocking, re-fill the SAME bytebuffer and template name variables with other informations, before performing a `return-object`.

These informations are related to every child elements of every main container.

### Motivation

To block further elements without hide the entire parent object.

### Additional context

Add the following bytecode patch before the last `if-nez`, followed by `return-object`:

```
move-object/from16 v1, p2
invoke-virtual {v1}, Ljava/lang/Object;->toString()Ljava/lang/String;
move-result-object v1
move-object/from16 v2, p3
iget-object v2, v2, {byteBufferClassRegister};->b:Ljava/nio/ByteBuffer;
invoke-static {v1, v2}, YourCopiedLithoADSBlockClass;->YourCopiedLithoADSBlockMethod(Ljava/lang/String;Ljava/nio/ByteBuffer;)Z
move-result v2
if-eqz v2, :cond_removeactionbarbutton
const/4 v0, 0x0
:cond_removeactionbarbutton
```

ghost:
Closes #326, #301, #108, #167, #295.


OxrxL:
Closed due to required rewritten from scratch.

PR #723, opened 2022-10-05
feat(youtube): Select Preferred Video quality like Youtube Vanced

johnconner122:
Depends on:

 - https://github.com/revanced/revanced-integrations/pull/170


oSumAtrIX:
If we are at it, might also put settings related to this patch in its own subscreen instead of having one switch in the Miscellaneous screen?

johnconner122:
> If we are at it, might also put settings related to this patch in its own subscreen instead of having one switch in the Miscellaneous screen?

ok, and I don't think we need remember-video-quality patch after quality preferences.

PR #759, opened 2022-10-10
fix(disable-capture-restriction): find instruction positions dynamically instead of hard-coding them

ThePBone:
A user pointed out that this Spotify patch failed when using the xManager Spotify Premium Mod as a base APK. They seem to inject random non-sense `const/4` instructions for obfuscation, which broke my patch because it had the position of an instruction that needs to be modified hard-coded.


oSumAtrIX:
While you find the index programmatically, maybe you can capture the index in a fingerprint via an opcode pattern?

ThePBone:
I can't do that because these filler opcodes are randomized with each release of their mod. They are in different positions and differ in quantity between versions, so I would have to add one opcode fingerprint per version.

PR #762, opened 2022-10-10
fix(theme): include coloring the comment action bar

OxrxL:


OxrxL:
@inotia00 👍

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/CommentsFilterBarPatch.kt
```diff
@@ -0,0 +1,40 @@
...
+@Description("Apply theme to comments filter action bar.")
+@ThemeCompatibility
+@Version("0.0.1")
+class CommentsFilterBarPatch : BytecodePatch(
+    listOf(
+        CommentsFilterBarFingerprint
+    )
+) {
+    override fun execute(context: BytecodeContext): PatchResult {
+
```
oSumAtrIX, line 26: Fix the formatting here

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/CommentsFilterBarPatch.kt
```diff
@@ -0,0 +1,40 @@
...
+        val result = CommentsFilterBarFingerprint.result!!
+        val method = CommentsFilterBarFingerprint.result!!.mutableMethod
```
oSumAtrIX, line 28: Avoid duplicate code here.

oSumAtrIX (changes requested):
Does this patch consider the capability of the theme patch to chise any color?

OxrxL:
> Does this patch consider the capability of the theme patch to chise any color?

I tried to apply the same patch of this PR on a fully "red colored" package, and the color of filter bar followed that of the other (red) elements.

So...I assume it works correctly.

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/CommentsFilterBarPatch.kt
```diff
@@ -0,0 +1,39 @@
...
+import app.revanced.patcher.extensions.instruction
+import app.revanced.patcher.patch.BytecodePatch
+import app.revanced.patcher.patch.PatchResult
+import app.revanced.patcher.patch.PatchResultSuccess
+import app.revanced.patcher.util.smali.ExternalLabel
+import app.revanced.patches.youtube.layout.theme.annotations.ThemeCompatibility
+import app.revanced.patches.youtube.layout.theme.fingerprints.CommentsFilterBarFingerprint
+
+@Name("comment-filter-bar-theme")
+@Description("Apply theme to comments filter action bar.")
```
oSumAtrIX, line 17:
```suggestion
@Description("Applies custom theming to comments filter action bar.")
```

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/CommentsFilterBarPatch.kt
```diff
@@ -0,0 +1,39 @@
...
+    )
+) {
+    override fun execute(context: BytecodeContext): PatchResult {
+        val result = CommentsFilterBarFingerprint.result!!
+        val method = result.mutableMethod
+        val patchIndex = result.scanResult.patternScanResult!!.endIndex - 1
+
+        method.addInstructions(
+            patchIndex, """
+                const v1, -0xdededf
```
oSumAtrIX, line 32: What is this magic constant? Can you explain it in a code comment?

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/CommentsFilterBarPatch.kt
OxrxL: The constant is based on a color value (in this case, a tint of gray) from register p1, returned by a method which manages a bytebuffer (another one, not the same used by ads blocking).

The aformentioned method return also different other color values (for this reason I added a if-condition), with which you can change the color of other elements of the bar (for example: the colors of the pressed buttons).

oSumAtrIX (approved):

PR #765, opened 2022-10-11
fix(youtube/general-ads): Add back general ad resource patch

TheNoFace:
From comments of issue ReVanced/revanced-patches-template#1955 

> > the are not randomly or rarely like people say in the issue ReVanced/revanced-patches-template#2343, the ads are always there
> 
> It looks like this issue is only on tablet landscape mode, no matter what brand the table or phone is. My samsung table also has ads showing always on main and player ui unless reverting resource patch commit.
> 
> However devs are no more using resource patch method, so just wait patiently until they come across another solution. Personally, I use my own fork of patch that use resource patch for hiding cosmetic ads. Try this one: [`revanced-patches`](https://github.com/TheNoFace/revanced-patches)

### DIFF
--- a/src/main/kotlin/app/revanced/patches/youtube/ad/general/resource/patch/GeneralResourceAdsPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/youtube/ad/general/resource/patch/GeneralResourceAdsPatch.kt
@@ -1,0 +1,60 @@
+package app.revanced.patches.youtube.ad.general.resource.patch
+
+import app.revanced.extensions.doRecursively
+import app.revanced.extensions.startsWithAny
+import app.revanced.patcher.annotation.Description
+import app.revanced.patcher.annotation.Name
+import app.revanced.patcher.annotation.Version
+import app.revanced.patcher.data.impl.ResourceData
+import app.revanced.patcher.patch.PatchResult
+import app.revanced.patcher.patch.PatchResultSuccess
+import app.revanced.patcher.patch.annotations.Dependencies
+import app.revanced.patcher.patch.impl.ResourcePatch
+import app.revanced.patches.youtube.ad.general.annotation.GeneralAdsCompatibility
+import app.revanced.patches.youtube.misc.manifest.patch.FixLocaleConfigErrorPatch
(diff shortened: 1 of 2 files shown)


src/main/kotlin/app/revanced/patches/youtube/ad/general/resource/patch/GeneralResourceAdsPatch.kt
```diff
@@ -0,0 +1,60 @@
...
+
+@DependsOn(dependencies = [FixLocaleConfigErrorPatch::class])
+@Name("general-resource-ads")
+@Description("Patch to remove general ads in resources.")
+@GeneralAdsCompatibility
+@Version("0.0.1")
+class GeneralResourceAdsPatch : ResourcePatch {
+    // list of resource file names which need to be hidden
+    private val resourceFileNames = arrayOf(
+        "compact_promoted_video_item.xml",
```
oSumAtrIX, line 25: Are all those resources required to be patched?

TheNoFace: Removed unnecessary resources.

src/main/kotlin/app/revanced/patches/youtube/ad/general/resource/patch/GeneralResourceAdsPatch.kt
```diff
@@ -22,13 +22,8 @@ import org.w3c.dom.Element
...
     private val resourceFileNames = arrayOf(
-        "compact_promoted_video_item.xml",
-        "inline_muted_metadata_swap.xml",
-        "interstitial_promo_view.xml",
-        "pip_ad_overlay.xml",
-        "promoted_",
-        "watch_metadata_companion_cards.xml",
-        //"watch_while_activity.xml" // FIXME: find out why patching this resource fails
+        "compact_promoted_",
+        "promoted_video_",
```
TheNoFace, line 26: Removed unnecessary resources.

oSumAtrIX (approved):

PR #766, opened 2022-10-11
Update ThemePatch.kt

LatroM69:


Ushie:
Can you tell us why this change is needed?

PR #781, opened 2022-10-13
Dynamic Color Patch (Youtube)

chowder908:
Made a patch that Replaces the light theme and the UI buttons of the dark theme with the user's phone color similar to how the Twitter patch works.


oSumAtrIX:
This PR gets rid of the capabilities of the existing `theme` patch. Instead extend on the existing code or keep the same functionality.

PR #818, opened 2022-10-19
fix(youtube/video-ads): restore video ads player block

OxrxL:


src/main/kotlin/app/revanced/patches/youtube/ad/video/fingerprints/LoadAdsParentFingerprint.kt
```diff
@@ -0,0 +1,21 @@
...
+
+import app.revanced.patcher.fingerprint.method.impl.MethodFingerprint
+import app.revanced.patches.youtube.ad.video.annotations.VideoAdsCompatibility
+import org.jf.dexlib2.AccessFlags
+import org.jf.dexlib2.Opcode
+
+@Name("load-ads-parent-fingerprint")
+@VideoAdsCompatibility
+@Version("0.0.1")
+object LoadAdsParentFingerprint : MethodFingerprint(
```
oSumAtrIX, line 15: By current conventions this fingerprint shouldn't be called parent as it is directly used to resolve to the method required to patch.

oSumAtrIX (changes requested):

oSumAtrIX (approved):

PR #831, opened 2022-10-22
feat(patch): Fix metaNav Scaling

halotroop2288:
RFC: This is my first time creating a patch. Does this code seem correct? If not, why?

Here is the intended diff (original on the right):
file difference


oSumAtrIX (approved):
Looks good, to get the svgNode element you probably do not need to first get the body node element, try omitting that.

oSumAtrIX:
Also do not commit to the patches json file, it is autogenerated.

oSumAtrIX (approved):

PR #873, opened 2022-10-27
fix(YouTube/theme): patched new gray value

ghost:
photo_2022-10-27_15-20-01


mexon86:
black color is nice but having all black is not. i cant see where is the opened window, it overlap and not look good if we have slightly bright gray to show the window, isnt it will be much good ?

PR #877, opened 2022-10-27
fix(YouTube/shorts button remover): improved fingerprint

OxrxL:


oSumAtrIX:
Why is this required?

Canny1913:
> Why is this required?

Fingerprint fails on latest versions.

OxrxL:
> > Why is this required?
> 
> Fingerprint fails on latest versions.

Exactly.

oSumAtrIX:
The latest versions are not supported. If you intend to, please do so in a single PR.

PR #973, opened 2022-11-04
feat(HexEditor): disable-ads

johnconner122:


src/main/kotlin/app/revanced/patches/hexeditor/ad/fingerprints/HexEditorAdsFingerprint.kt
```diff
@@ -0,0 +1,15 @@
...
+
+import app.revanced.patcher.annotation.Name
+import app.revanced.patcher.annotation.Version
+import app.revanced.patcher.fingerprint.method.impl.MethodFingerprint
+import app.revanced.patches.hexeditor.ad.annotations.HexEditorAdsCompatibility
+
+@Name("ads-fingerprint-primary")
+@HexEditorAdsCompatibility
+@Version("0.0.1")
+object HexEditorAdsFingerprint : MethodFingerprint(
```
oSumAtrIX, line 11: Maybe name this `PrimaryAdsFingerprint`.

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/hexeditor/ad/patch/HexEditorAdsPatch.kt
```diff
@@ -0,0 +1,43 @@
...
+        
+        method.replaceInstructions(
+            0,
+            """
+                const/4 v0, 0x1
+                return v0
+            """
+        )
+
+        for(i in 0..4)
```
Anajrim01, line 38: Why this for loop if it isn't used anywhere?

johnconner122: To remove useless instructions, as we are already returning from function

Anajrim01: You have hardcoded 2, the for loop is never used :).

johnconner122: Because there are 5 instructions after returning from function, as we are returning on index 1, so we'll remove index 2 five times

Anajrim01: Right, got it. It all LGTM then

src/main/kotlin/app/revanced/patches/hexeditor/ad/patch/HexEditorAdsPatch.kt
```diff
@@ -0,0 +1,43 @@
...
+            method.removeInstruction(2)
```
oSumAtrIX, line 39:
```suggestion
```

Can be removed. The second instruction in this method will return the method anyways.

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/hexeditor/ad/patch/HexEditorAdsPatch.kt
```diff
@@ -0,0 +1,40 @@
...
+import app.revanced.patcher.patch.annotations.Patch
+import app.revanced.patches.hexeditor.ad.annotations.HexEditorAdsCompatibility
+import app.revanced.patches.hexeditor.ad.fingerprints.PrimaryAdsFingerprint
+
+@Patch
+@Name("disable-ads")
+@Description("Disables ads in HexEditor.")
+@HexEditorAdsCompatibility
+@Version("0.0.1")
+
```
oSumAtrIX, line 21:
```suggestion
```

oSumAtrIX (changes requested):

PR #977, opened 2022-11-04
feat: short app name

kazimmt:
Current name is too long that it doesn't even fit on launcher. Another thing is, main YouTube music also use YT Music on Launcher

PR #979, opened 2022-11-05
feat: change patched YouTube name

Nan123721:
This changes the default YouTube name to be more compact, and more importantly, actually fit into the launcher. Similar to #977, but for YouTube


oSumAtrIX:
ReVanced patches multiple apps. It is not specific for YouTube.

NeeRaj-2401:
> actually fit into the launcher

@Nan123721 alternativerly u should have changed it **"YT Revanced"** for this purpose

oSumAtrIX:
YT is an unofficial abbreviation. You can do it via the patch options yourself.

PR #1005, opened 2022-11-08
refactor(youtube/general-ads-patch): removed old video ads references

OxrxL:


OxrxL:
@oSumAtrIX This reference is still needed?

PR #1006, opened 2022-11-09
refactor(youtube/general-ads-patch): squash extensions classes

OxrxL:


OxrxL:
@oSumAtrIX It's worth keeping unique methods of general ads patch, in separate classes?

oSumAtrIX (approved):

PR #1024, opened 2022-11-12
Update PivotBarFingerprint.kt

nibir34:
This should fix `hide-create-button` & `hide-shorts-button` patch on YouTube version 17.44.34


oSumAtrIX:
Does this still guarantee compatibility with existing versions?

PR #1041, opened 2022-11-13
feat(twitch/block-audio-ads): Blocks audio ads in streams and VODs

ThePBone:
This patch blocks [audio ads](https://help.twitch.tv/s/article/audio-ads?language=en_US) in the Twitch app (= stream is visible, but an ad takes over the audio).

I didn't even know these ads existed until I went through the decompilation, but luckily, Twitch's debug mode had the option to force show them in every stream so that I could test my blocking method properly.


src/main/kotlin/app/revanced/patches/twitch/ad/audio/fingerprints/AudioAdsPresenterPlayFingerprint.kt
```diff
@@ -0,0 +1,17 @@
+package app.revanced.patches.twitch.ad.audio.fingerprints
+
+import app.revanced.patcher.annotation.Name
+import app.revanced.patcher.annotation.Version
+
+import app.revanced.patcher.fingerprint.method.impl.MethodFingerprint
+import app.revanced.patches.twitch.ad.audio.annotations.AudioAdsCompatibility
+
+@Name("audio-ads-presenter-play-fingerprint")
+
```
oSumAtrIX, line 10:
```suggestion
```

oSumAtrIX (approved):

PR #1047, opened 2022-11-14
feat(iconpackstudio): unlock pro features

FineFindus:
Adds a patch that unlocks the pro mode in [IconPackStudio](https://play.google.com/store/apps/details?id=ginlemon.iconpackstudio).


src/main/kotlin/app/revanced/patches/iconpackstudio/misc/pro/fingerprints/CheckProFingerprint.kt
```diff
@@ -0,0 +1,15 @@
...
+import app.revanced.patcher.annotation.Version
+import app.revanced.patcher.fingerprint.method.impl.MethodFingerprint
+import app.revanced.patches.iconpackstudio.misc.pro.annotations.UnlockProCompatibility
+
+@Name("check-pro-fingerprint")
+@UnlockProCompatibility
+@Version("0.0.1")
+object CheckProFingerprint : MethodFingerprint(
+    customFingerprint = { methodDef ->
+        methodDef.definingClass.endsWith("IPSPurchaseRepository;") && methodDef.name == "b"
```
oSumAtrIX, line 13: Method name "b" can change in future versions, either constrain the package version or find a different fingerprint.

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/iconpackstudio/misc/pro/fingerprints/CheckProFingerprint.kt
FineFindus: How large should the fingerprint be? Currently, there is only one method with a boolean return in the class, would that be enough? Or should it also include the signature and implementation?

oSumAtrIX: It always depends on the context, in yours, since the class only contains one method with a boolean return value, it would be enough to simply create a fingerprint which asks for the class name and the return value of boolean (Z)

oSumAtrIX (approved):

PR #1057, opened 2022-11-16
fix(tiktok): add compatibility for new version

d4rkk3y:
https://github.com/orgs/revanced/discussions/734

### DIFF
--- a/src/main/kotlin/app/revanced/patches/tiktok/misc/settings/patch/TikTokSettingsPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/tiktok/misc/settings/patch/TikTokSettingsPatch.kt
@@ -29,7 +29,7 @@ import org.jf.dexlib2.iface.reference.TypeReference
 @Patch
 @DependsOn([TikTokIntegrationsPatch::class])
 @Name("settings")
-@Description("Adds settings to TikTok.")
+@Description("Adds settings for ReVanced to TikTok.")
 @TikTokSettingsCompatibility
 @Version("0.0.1")
 class TikTokSettingsPatch : BytecodePatch(
--- a/src/main/kotlin/app/revanced/patches/tiktok/misc/settings/patch/TikTokSettingsPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/tiktok/misc/settings/patch/TikTokSettingsPatch.kt
@@ -14,6 +14,7 @@ import app.revanced.patcher.patch.annotations.DependsOn
 import app.revanced.patcher.patch.annotations.Patch
 import app.revanced.patches.tiktok.misc.integrations.patch.TikTokIntegrationsPatch
 import app.revanced.patches.tiktok.misc.settings.annotations.TikTokSettingsCompatibility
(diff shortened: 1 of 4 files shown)


oSumAtrIX:
Does this still guarantee compatibility with current versions?

src/main/kotlin/app/revanced/patches/tiktok/misc/settings/patch/TikTokSettingsPatch.kt
```diff
@@ -79,12 +83,39 @@ class TikTokSettingsPatch : BytecodePatch(
             )
             break
         }
+        //Change onClick of About to start revanced settings
+        val method2 = AboutOnClickMethodFingerprint.result!!.mutableMethod
```
oSumAtrIX, line 87: Maybe chose a more describing name rather than "method2".

src/main/kotlin/app/revanced/patches/tiktok/misc/settings/patch/TikTokSettingsPatch.kt
```diff
@@ -79,12 +83,39 @@ class TikTokSettingsPatch : BytecodePatch(
...
+        method2.addInstructions(
+            0,
+            """
+                    invoke-static {}, Lapp/revanced/tiktok/settingsmenu/SettingsMenu;->startSettingsActivity()V
+                    return-void
+                """
+        )
+        //Replace string `About` to 'Revanced Settings` in TikTok settings.
+        for ((index, instruction) in implementation1.instructions.withIndex()) {
```
oSumAtrIX, line 96: Instead of searching for the instruction index, either use an opcode pattern or use the string scan result from the fingerprint and a fixed offset.

oSumAtrIX (changes requested):

d4rkk3y:
> Does this still guarantee compatibility with current versions?
I tested with lasted version. It worked.

src/main/kotlin/app/revanced/patches/tiktok/misc/settings/patch/TikTokSettingsPatch.kt
Sculas: You could try and match the first one and do some heuristic offsetting from there (try to find your instruction from the matching point, don't directly index unless you can guarantee it'll work across versions).

oSumAtrIX:
Made some changes. Please test if everything works as intended. Also if possible create methods for each kind of patch you do there. That way the execute method won't be this large.

src/main/kotlin/app/revanced/patches/tiktok/misc/settings/patch/TikTokSettingsPatch.kt
d4rkk3y: @Sculas There are more bytecode which have same opcode pattern in target method. The pic only shows one of them.

d4rkk3y:
> Made some changes. Please test if everything works as intended. Also if possible create methods for each kind of patch you do there. That way the execute method won't be this large.

Tested your changes with tiktok 26.9.4 and everything worked.

d4rkk3y:
@oSumAtrIX are there anything need to be improved to make this PR be merged?

PR #1058, opened 2022-11-17
Rebase feat/split

JavierFlores09:
This PR is just to rebase feat/split branch, because it was way behind main


Sculas:
Don't think this is necessary, since my (upcoming) PR for the split branch already did this.

oSumAtrIX:
> Don't think this is necessary, since my (upcoming) PR for the split branch already did this.

To avoid situations like this, it is adviced to create an issue to track that it is already being worked on so that a faster PR than yours does not obstruct yours.

JavierFlores09:
> Don't think this is necessary, since my (upcoming) PR for the split branch already did this.

I just did this because I also planned to work on the branch, and if I just rebased and did my changes on top, then PR'd, we'd be again at merge-conflict hell when it came to your PR or any other change from team (as rebasing itself caused conflicts).

As osum said, a heads up with an issue or in #development would've been nice, either way if you believe you are indeed going to PR your changes then I can just close this

Sculas:
I stopped working on my PR due to certain issues, so this PR can be merged if wanted.

oSumAtrIX:
I am not sure how merging this PR will behave, will it actually rebase or add all those commits on top?

JavierFlores09:
> I am not sure how merging this PR will behave, will it actually rebase or add all those commits on top?

I believe the latter since the rebased commits are are newer, if you want them to appear as they're newer I can do that though

oSumAtrIX:
Rebasing is not possible, to rebase it is required to force push to this branch which is why this PR should be closed for now.

PR #1052, opened 2022-11-15
refactor: merge integrations code

Sculas:
This PR merges the integrations code between TikTok and YouTube since they're basically the same.

PR #1074, opened 2022-11-19
feat(twitch/integrations): Add integrations support for Twitch

ThePBone:
Waiting for #1052, so I can add the new integrations patch abstraction.

Depends on: https://github.com/revanced/revanced-integrations/pull/216

### DIFF
--- a/src/main/kotlin/app/revanced/patches/twitch/misc/integrations/patch/TwitchIntegrationsPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/twitch/misc/integrations/patch/TwitchIntegrationsPatch.kt
@@ -1,39 +1,13 @@
 package app.revanced.patches.twitch.misc.integrations.patch
 
-import app.revanced.patcher.annotation.Description
 import app.revanced.patcher.annotation.Name
-import app.revanced.patcher.annotation.Version
-import app.revanced.patcher.data.BytecodeContext
-import app.revanced.patcher.extensions.addInstruction
-import app.revanced.patcher.patch.BytecodePatch
-import app.revanced.patcher.patch.PatchResult
-import app.revanced.patcher.patch.PatchResultError
-import app.revanced.patcher.patch.PatchResultSuccess
-import app.revanced.patcher.patch.annotations.Patch
 import app.revanced.patches.twitch.misc.integrations.fingerprints.InitFingerprint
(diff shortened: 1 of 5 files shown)


ThePBone:
I rebased #1052 onto this PR and updated it to support the abstract integrations patch. Should be ready to go now.

oSumAtrIX:
Seems like rebasing caused issues since the other PR you rebased on was squashed before merge. Can you revert the rebase/rebase onto main?

ThePBone:
I had to force-push to overwrite the old commits; it is up-to-date now.

oSumAtrIX (approved):

src/main/kotlin/app/revanced/patches/twitch/misc/integrations/patch/TwitchIntegrationsPatch.kt
```diff
@@ -0,0 +1,13 @@
+package app.revanced.patches.twitch.misc.integrations.patch
+
+import app.revanced.patcher.annotation.Name
+import app.revanced.patches.twitch.misc.integrations.fingerprints.InitFingerprint
+import app.revanced.patches.twitch.misc.integrations.annotations.TwitchIntegrationsCompatibility
+import app.revanced.shared.patches.AbstractIntegrationsPatch
+
+@Name("twitch-integrations")
```
oSumAtrIX, line 8: The application name can be omitted here because patches are already scoped to packages implicitely.

```suggestion
@Name("integrations")
```

src/main/kotlin/app/revanced/patches/twitch/misc/integrations/patch/TwitchIntegrationsPatch.kt
```diff
@@ -0,0 +1,13 @@
...
+@TwitchIntegrationsCompatibility
+class TwitchIntegrationsPatch : AbstractIntegrationsPatch(
```
oSumAtrIX, line 10: For the same reason as the other suggestion, this class should be named Integrations, as there is no reason to name the app due to the class being already in the correct package for twitch patches. Reflect this change with the compatibility annotation class as well.

oSumAtrIX (changes requested):

oSumAtrIX (approved):

PR #1075, opened 2022-11-19
feat(twitch/settings): Add settings menu patch

ThePBone:
The current patches in this repo don't have support for the settings yet, so the menu is still technically blank. I'll create separate PRs for the updated patches later on.

Depends on:
* https://github.com/revanced/revanced-patches/pull/1074
* https://github.com/revanced/revanced-integrations/pull/216

screenshot


oSumAtrIX:
Currently, this PR introduces a lot of duplicate code. Instead, the current code should be abstracted, so this PR can use the existing settings patch and reduce duplicate code which would otherwise cause problems with maintenance in the future.

ThePBone:
Changes tested and ready to merge

src/main/kotlin/app/revanced/patches/twitch/misc/settings/components/CustomPreferenceCategory.kt
```diff
@@ -0,0 +1,20 @@
+package app.revanced.patches.twitch.misc.settings.components
+
+import app.revanced.patches.shared.settings.BasePreference
+import app.revanced.patches.shared.settings.impl.PreferenceCategory
+import app.revanced.patches.shared.settings.impl.StringResource
+
+/**
+ * Customized reference category for Twitch.
```
oSumAtrIX, line 8:
```suggestion
 * Customized preference category for Twitch.
```

src/main/kotlin/app/revanced/patches/twitch/misc/settings/resource/patch/SettingsResourcePatch.kt
```diff
@@ -0,0 +1,133 @@
...
+import app.revanced.patches.youtube.misc.manifest.patch.FixLocaleConfigErrorPatch
+import app.revanced.util.resources.ResourceUtils
+import app.revanced.util.resources.ResourceUtils.copyResources
+import org.w3c.dom.Node
+
+@Name("settings-resource-patch")
+@SettingsCompatibility
+@DependsOn([FixLocaleConfigErrorPatch::class])
+@Version("0.0.1")
+class SettingsResourcePatch : ResourcePatch {
```
oSumAtrIX, line 27: This class is very similar to the `SettingsResourcePatch` class for YouTube. Probably, it is not a bad idea, to create an abstract class for this, the same way it has been done for the integrations patch for YouTube and TikTok.

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/twitch/misc/settings/resource/patch/SettingsResourcePatch.kt
ThePBone: Sure, I'll take a look. `SettingsResourcePatch` has most of this logic in a companion object, so I'd probably need to create an abstract class for the companion object and one for the actual patch.

oSumAtrIX: An abstract class was just an idea, if you see a better fitting implementation to get rid of the duplicate code, then do that.

ThePBone:
Had to force push because of accidental commits; the only new change since you last reviewed is d66ac52.

src/main/kotlin/app/revanced/patches/twitch/misc/settings/resource/patch/SettingsResourcePatch.kt
```diff
@@ -0,0 +1,62 @@
...
+        super.execute(context)
+
+        /* required for self-restart in settings */
+        context.xmlEditor["AndroidManifest.xml"].use { editor ->
+            editor.file.getElementsByTagName("manifest").item(0).also {
+                it.appendChild(it.ownerDocument.createElement("uses-permission").also { element ->
+                    element.setAttribute("android:name", "android.permission.SCHEDULE_EXACT_ALARM")
+                })
+            }
+        }
```
oSumAtrIX, line 30: I have seen this code block for another patch as well, creating a patch and using it as a dependency on both, this and the other patch which uses the same code is probably a good idea to prevent code duplication, or you move it to `AbstractSettingsResourcePatch` into a method or the super method.

src/main/kotlin/app/revanced/patches/twitch/misc/settings/fingerprints/MenuGroupsUpdatedFingerprint.kt
```diff
@@ -0,0 +1,16 @@
+package app.revanced.patches.twitch.misc.settings.fingerprints
+
+import app.revanced.patcher.annotation.Name
+import app.revanced.patcher.annotation.Version
+import app.revanced.patcher.fingerprint.method.impl.MethodFingerprint
+import app.revanced.patches.twitch.misc.settings.annotations.SettingsCompatibility
+
+@Name("settings-menu-groups-update-fingerprint")
+@SettingsCompatibility
+@Version("0.0.1")
```
oSumAtrIX, line 10: Annotations for fingerprints have been deprecated a while back, you can safely remove them.

oSumAtrIX (changes requested):

oSumAtrIX (approved):

oSumAtrIX:
Thanks.

PR #1082, opened 2022-11-21
fix(tiktok/settings): tiktok's settings patch name make 'revanced manager' confuse with youtube settings patch name

d4rkk3y:
https://discord.com/channels/952946952348270622/1044056142864711751


oSumAtrIX:
This shouldn't happen. Patches are constrained to their respective package. Meaning, if you have selected TikTok, it should show patches for TikTok. That way you wouldn't he able to include a patch from YouTube.

oSumAtrIX:
This could be a problem with the ReVanced Manager including the wrong patch.

d4rkk3y:
Yes, this is a problem with Revanced Manager because it is fine when patching with RevancedCLI. This PR is just a temporary fix for Revanced Manager. If it is better to wait for Revanced Manager to fix this issue, then please close this PR.

oSumAtrIX:
Can you open a draft PR to remove "tiktok" from the tiktok patches after this PR is merged. That way we can merge it after it is fixed on the ReVanced Manager side.

Issue #675, opened 2022-09-30
Make a patch for TikTok to work in restricted countries


mahdi921:
### Application

TikTok

### Issue

In some countries like Iran and India, access to TikTok is restricted by government. The issue is that TikTok somehow recognizes the device is in those countries and won't work even with vpn connection.
On the other hand when I remove any simcards and connect to a wifi with vpn, the application works without a problem. 
In the past, I was able to use app cloner to prevent TikTok from accessing sim and cell info by hiding them, but not without some problems showing up.
Now the app won't work after cloning.
The most ironic thing is that we can access tiktok through its website with vpn connection.
I have done everything suggested by google search to resolve the issue. NONE works.

### Patch

The patch should prevent the app from recognizing any of the simcard info and simcard country and/or anything related to help the app work in restricted countries

### Motivation

TikTok had many users from Iran and India. It would be good news for them to again access their accounts and their content or even use the app without a problem.

### Acknowledgements

- [X] I have searched the existing issues and this is a new and no duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I filled out all of the requested information in this issue properly.

oSumAtrIX:
I think it does by IP, for that you need a vpn.

mahdi921:
VPN won't work in this case. About revoking permissions, there's nothing in android settings as I have searched every corner of it. It can't be done without some tinkering the system or the app.
I thought maybe a patch could help to go around this issue. 
It's so ridiculous.

oSumAtrIX:
Isn't Android meant to opt in for permissions for apps? I can see a setting to disable access to APIs for apps on OxygenOS.

mahdi921:
Yeah right. That's true for some custom roms. But what about stock roms. Stock roms don't have this luxury as they are made to be luxurious. And most people don't even know what custom rom is.
I and many people are looking forward for a workaround to solve this issue. Nevertheless, in Iran we are in the process of changing the government, so the workaround isn't going to be needed in the future. However, it would be good for the people with non-changing governments with restricted access to get a workaround.
But in any case thank you for bringing back and continuing revanced.

PR #1088, opened 2022-11-22
feat(tiktok/sim-spoof): fakes sim card information to bypass regional restriction.

d4rkk3y:
depend on: https://github.com/revanced/revanced-integrations/pull/219
issue: https://github.com/revanced/revanced-patches/issues/675

Because I am not in restricted region, I can not test if it work as intended. Waiting for someone have regional restriction problem test this PR.


TheMerret:
Works well. Message about restricting access is disappeared, noew new videos can be viewed. I haven't tried uploading videos yet.

src/main/kotlin/app/revanced/patches/tiktok/misc/simspoof/patch/SimSpoofPatch.kt
```diff
@@ -0,0 +1,103 @@
...
+import app.revanced.patches.tiktok.misc.simspoof.annotations.SimSpoofCompatibility
+import org.jf.dexlib2.Opcode
+import org.jf.dexlib2.iface.ClassDef
+import org.jf.dexlib2.iface.Method
+import org.jf.dexlib2.iface.instruction.OneRegisterInstruction
+import org.jf.dexlib2.iface.instruction.formats.Instruction35c
+import org.jf.dexlib2.iface.reference.MethodReference
+import org.jf.dexlib2.immutable.reference.ImmutableMethodReference
+
+@Patch(false)
```
oSumAtrIX, line 29: Is this intended?

src/main/kotlin/app/revanced/patches/tiktok/misc/simspoof/patch/SimSpoofPatch.kt
```diff
@@ -0,0 +1,103 @@
...
+                0,
+                "invoke-static {}, Lapp/revanced/tiktok/settingsmenu/SettingsStatus;->enableSimSpoof()V"
+            )
+        }
+        return PatchResultSuccess()
+    }
+
+    //Patch android API and return fake sim information
+    private fun patchAPI(context: BytecodeContext, classDef: ClassDef, method: Method) {
+        with(context.proxy(classDef).mutableClass.findMutableMethodOf(method)) {
```
oSumAtrIX, line 86: Instead of having to iterate over all instructions again, you can just pass the index where the instruction you want to patch is located at.

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/tiktok/misc/simspoof/patch/SimSpoofPatch.kt
d4rkk3y: Yes, because I think this patch targets some regions only. It should not be included by default. It also modifies the content of the tiktok delivered to the user based on the region it pretends to be in.

src/main/kotlin/app/revanced/patches/tiktok/misc/simspoof/patch/SimSpoofPatch.kt
d4rkk3y: I did as you suggested before, but I ran into the issue that when two places in a method need to be patched, if the first place is patched, the index of the second place will change. Can you suggest a way to solve this problem?

oSumAtrIX: Collect all indices and pass them as an array. This method will add instructions in reverse order, so from bottom to top.

oSumAtrIX (changes requested):

oSumAtrIX:
Made some changes, please review if everything is correct and we can get it merged.

d4rkk3y:
MutableClass.findMutableMethodOf possibly resolve to wrong method. When I test, It make the patch fail. 
image

oSumAtrIX:
Should be resolved now, give it a try.

d4rkk3y:
> Should be resolved now, give it a try.

I tested and all work as intended with tiktok 27.1.3.

oSumAtrIX:
Thanks.

megapro17:
How to use it? It doesn't seem to work, there's no setting in revanced settings

oSumAtrIX:
Please open an issue in that case.

PR #1109, opened 2022-11-25
refactor: abstract settings patch

ThePBone:
This PR adds `AbstractSettingsResourcePatch`


src/main/kotlin/app/revanced/patches/shared/settings/patch/AbstractSettingsResourcePatch.kt
```diff
@@ -0,0 +1,137 @@
...
+abstract class AbstractSettingsResourcePatch(
+    val revancedPreferencesName: String,
+    val revancedPreferencesSourceDir: String,
+) : ResourcePatch {
+    override fun execute(context: ResourceContext): PatchResult {
+        /* used for self-restart */
+        context.xmlEditor["AndroidManifest.xml"].use { editor ->
+            editor.file.getElementsByTagName("manifest").item(0).also {
+                it.appendChild(it.ownerDocument.createElement("uses-permission").also { element ->
+                    element.setAttribute("android:name", "android.permission.SCHEDULE_EXACT_ALARM")
```
oSumAtrIX, line 38: I am not sure, but I think this is only required for YouTube.

ThePBone: I also need it for the Twitch patch; however, it isn't necessarily required by future settings patches for other apps, so I moved it back to YouTube's class.

oSumAtrIX:
Looks good, when I was reviewing, I noticed that the PreferenceScreen enum was left unabstracted. Your other PR also created such enum with almost the same implementation, was this considered?

ThePBone:
> Looks good, when I was reviewing, I noticed that the PreferenceScreen enum was left unabstracted. Your other PR also created such enum with almost the same implementation, was this considered?

The only difference is that the Twitch patch uses mainly PreferenceCategories and YouTube PreferenceScreens; I've implemented a compromise for this, check below.
The enum classes can't inherit from classes, nor can be extended, so I rewrote it completely.

I added (optional) support for PreferenceCategories nested within PreferenceScreens. It isn't used by the YouTube settings patch currently, but here's how it could be used:

```kotlin
internal object PreferenceScreens : AbstractPreferenceScreen() {
    val ADS = CategorizableScreen("ads", "Ads", "Ad related settings")
    val INTERACTIONS = CategorizableScreen("interactions", "Interaction", "Settings related to interactions")
    val LAYOUT = CategorizableScreen("layout", "Layout", "Settings related to the layout")
    val MISC = CategorizableScreen("misc", "Misc", "Miscellaneous patches")

    class CategorizableScreen(key: String, title: String, summary: String?) : Screen(key, title, summary) {
        val PLAYER = Category("player", "Video player")
        val COMMENTS = Category("comments", "Comments")
    }

    override fun commit(screen: PreferenceScreen) {
        addPreferenceScreen(screen)
    }
}
```

Patches can then categorize their options within PreferenceScreens: 
```kotlin
SettingsPatch.PreferenceScreens.LAYOUT.PLAYER.addPreferences(..)
```
Or add preferences to the root, like normal:
```kotlin
SettingsPatch.PreferenceScreens.LAYOUT.addPreferences(..)
```

oSumAtrIX:
An enum can extend on an interface with default implementations, but I am not sure the realization of that makes proper sense in sense of polymorphism. Fortunately, it would allow to keep the semantics of enums instead of having to change the `PreferenceScreens` class to add new categories and violate the open-closed principle. I'll leave this decision to you though, if you decide for either implementation, finalize this PR by populate the changes consistently for all patches.

ThePBone:
Yes, I think I'll stick to my current implementation.

All the patches have already been updated and tested with YouTube.

src/main/kotlin/app/revanced/patches/shared/settings/util/AbstractPreferenceScreens.kt
```diff
@@ -0,0 +1,89 @@
+package app.revanced.patches.shared.settings.util
+
+import app.revanced.patches.shared.settings.preference.BasePreference
+import app.revanced.patches.shared.settings.preference.impl.PreferenceCategory
+import app.revanced.patches.shared.settings.preference.impl.PreferenceScreen
+import app.revanced.patches.shared.settings.preference.impl.StringResource
+import java.io.Closeable
+
+internal abstract class AbstractPreferenceScreens(
```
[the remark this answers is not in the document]
ThePBone: Fixed.

oSumAtrIX (approved):

oSumAtrIX:
Thanks.

PR #1111, opened 2022-11-25
feat(YouTube): bump all patches to v17.45.36

johnconner122:


src/main/kotlin/app/revanced/patches/youtube/layout/pivotbar/createbutton/patch/CreateButtonRemoverPatch.kt
```diff
@@ -31,7 +32,7 @@ import app.revanced.patches.shared.settings.impl.SwitchPreference
 @Version("0.0.1")
 class CreateButtonRemoverPatch : BytecodePatch(
     listOf(
-        PivotBarFingerprint
+        PivotBarFingerprint, PivotBarNewFingerprint
```
oSumAtrIX, line 35: While this works, it is not an ideal solution and should only be done if theres no other way. This fingerprint also contains a fairly large fingerprint, which is why it would be relatively prone to errors. Instead, a fingerprint which works for both, the old and new versions should be created. Since the method which is modified has been proven to change frequently, it is better to resolve to another method, then for example use the `MethodWalker` to get to this method and dynamically find the places which need to be modified by iterating through the instructions or using fingerprints to find the indices of instruction opcode patterns. What I am asking is not too complex, but still might be unclear, which is why if you do not know how to realize this, convert the PR to a draft or close it.

src/main/kotlin/app/revanced/patches/youtube/layout/widesearchbar/fingerprints/WideSearchbarTwoParentFingerprint.kt
```diff
@@ -5,6 +5,6 @@ import app.revanced.patcher.fingerprint.method.impl.MethodFingerprint
 import org.jf.dexlib2.AccessFlags
 
 object WideSearchbarTwoParentFingerprint : MethodFingerprint(
-    "L", AccessFlags.PUBLIC or AccessFlags.STATIC, listOf("L"),
-    strings = listOf("VIDEO_QUALITIES_QUICK_MENU_BOTTOM_SHEET_FRAGMENT")
+    "L", AccessFlags.PUBLIC or AccessFlags.STATIC, null, null,
+    strings = listOf("Callback already registered.", "Failed to create SpotlightModeController.")
```
oSumAtrIX, line 9: Does this also guarantee compatibility with the older versions?
```suggestion
    strings = listOf("Callback already registered.", "Failed to create SpotlightModeController.")
```

oSumAtrIX (changes requested):

PR #1116, opened 2022-11-26
feat(backdrops): pro-unlock patch

josesilveiraa:


src/main/kotlin/app/revanced/patches/backdrops/misc/pro/patch/ProUnlockPatch.kt
```diff
@@ -0,0 +1,38 @@
...
+@ProUnlockCompatibility
+@Version("0.0.1")
+class ProUnlockPatch : BytecodePatch(
+    listOf(ProUnlockFingerprint)
+) {
+    override fun execute(context: BytecodeContext): PatchResult {
+        val result = ProUnlockFingerprint.result ?: return PatchResultError("${ProUnlockFingerprint.name} not found")
+
+        result.mutableMethod.addInstructions(
+            23,
```
oSumAtrIX, line 29: This index is arbitrary and can change if the app updates. Please use a fingerprint opcode pattern to get the index.

oSumAtrIX (changes requested):

PR #1121, opened 2022-11-26
feat(backdrops): `pro-unlock` patch

josesilveiraa:


src/main/kotlin/app/revanced/patches/backdrops/misc/pro/patch/ProUnlockPatch.kt
```diff
@@ -0,0 +1,38 @@
...
+    listOf(ProUnlockFingerprint)
+) {
+    override fun execute(context: BytecodeContext): PatchResult {
+        val result = ProUnlockFingerprint.result ?: return PatchResultError("${ProUnlockFingerprint.name} not found")
+
+        result.mutableMethod.addInstructions(
+            result.scanResult.patternScanResult!!.endIndex,
+            """
+                const/4 v4, 0x6
+                const/4 v2, 0x1
```
oSumAtrIX, line 32: The register 4 and 2 might change, are you confident they wont?

josesilveiraa: No, I don't think they'll change. the app haven't had an update for 2 years. But who knows

josesilveiraa: Also, what can I do in this kind of situation? I started learning smali Bytecode yesterday, so I don't know very much of this

oSumAtrIX: You'd dynamically get the registers based on the current context, such as existing instructions which use them.

oSumAtrIX (approved):

PR #1142, opened 2022-11-28
feat(twitch): add settings support for existing patches

ThePBone:
This is the last PR I have queued up for the Twitch settings patch.
I combined it into one PR because the changes per patch are minor.

### DIFF
--- a/src/main/kotlin/app/revanced/patches/twitch/chat/antidelete/patch/ShowDeletedMessagesPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/twitch/chat/antidelete/patch/ShowDeletedMessagesPatch.kt
@@ -6,13 +6,19 @@ import app.revanced.patcher.annotation.Version
 import app.revanced.patcher.data.BytecodeContext
 import app.revanced.patcher.extensions.*
 import app.revanced.patcher.patch.*
+import app.revanced.patcher.patch.annotations.DependsOn
 import app.revanced.patcher.patch.annotations.Patch
+import app.revanced.patcher.util.smali.ExternalLabel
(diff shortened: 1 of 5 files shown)


src/main/kotlin/app/revanced/patches/twitch/chat/antidelete/patch/ShowDeletedMessagesPatch.kt
```diff
@@ -21,25 +27,77 @@ class ShowDeletedMessagesPatch : BytecodePatch(
...
             )
         }
 
-        // Disable setHasModAccess setter
-        with(SetHasModAccessFingerprint.result!!.mutableMethod.implementation!!) {
-            addInstruction(0, BuilderInstruction10x(Opcode.RETURN_VOID))
+        // Spoiler mode: Disable setHasModAccess setter
+        with(SetHasModAccessFingerprint.result!!) {
+            mutableMethod.addInstruction(0, "return-void")
         }
```
oSumAtrIX, line 56: The [with scope function](https://kotlinlang.org/docs/scope-functions.html#with) generally makes sense if you want to make a semantically contextual switch in code. Here you use with to switch to the semantic context of "We are now working with the result of the `SetHasModAccessFingerprint` fingerprint". This usually makes sense if your code block involves lots of logic which is scoped to the new context, but for this code it is suggested to simply do the following:

```suggestion
        SetHasModAccessFingerprint.result!!.mutableMethod.addInstruction(0, "return-void")
```

You do not have to apply this in this PR. Also, while asserting non-null is easy and is commonly used through the patches, it is recommended to handle them, not required in this PR but in the future.

oSumAtrIX (approved):

PR #1172, opened 2022-12-01
feat(sleepasandroid): unlock premium features

aliveoutside:
a monkey patch which unlocks premium features in [Sleep as Android](play.google.com/store/apps/details?id=com.urbandroid.sleep)

### DIFF
--- a/src/main/kotlin/app/revanced/patches/urbandroid/sleep/annotations/UnlockPremiumCompatibility.kt
+++ b/src/main/kotlin/app/revanced/patches/urbandroid/sleep/annotations/UnlockPremiumCompatibility.kt
@@ -1,0 +1,9 @@
+package app.revanced.patches.urbandroid.sleep.annotations
+
+import app.revanced.patcher.annotation.Compatibility
+import app.revanced.patcher.annotation.Package
+
+@Compatibility([Package("com.urbandroid.sleep")])
+@Target(AnnotationTarget.CLASS)
+@Retention(AnnotationRetention.RUNTIME)
+internal annotation class UnlockPremiumCompatibility
--- a/src/main/kotlin/app/revanced/patches/urbandroid/sleep/patch/UnlockPremiumPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/urbandroid/sleep/patch/UnlockPremiumPatch.kt
@@ -1,0 +1,36 @@
+package app.revanced.patches.urbandroid.sleep.patch
+
+import app.revanced.patcher.annotation.Description
+import app.revanced.patcher.annotation.Name
+import app.revanced.patcher.data.BytecodeContext
+import app.revanced.patcher.extensions.addInstructions
+import app.revanced.patcher.patch.BytecodePatch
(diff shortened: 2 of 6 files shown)


src/main/kotlin/app/revanced/patches/urbandroid/sleep/patch/UnlockPremiumPatch.kt
```diff
@@ -0,0 +1,36 @@
...
+import app.revanced.patcher.extensions.addInstructions
+import app.revanced.patcher.patch.BytecodePatch
+import app.revanced.patcher.patch.PatchResult
+import app.revanced.patcher.patch.PatchResultSuccess
+import app.revanced.patcher.patch.annotations.Patch
+import app.revanced.patches.urbandroid.sleep.annotations.UnlockPremiumCompatibility
+import app.revanced.patches.urbandroid.sleep.fingerprints.IsTrialFingerprint
+
+@Patch
+@Name("sleep-premium-unlock")
```
oSumAtrIX, line 15:
```suggestion
@Name("unlock-premium")
```

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/urbandroid/sleep/annotations/UnlockPremiumCompatibility.kt
```diff
@@ -0,0 +1,9 @@
+package app.revanced.patches.urbandroid.sleep.annotations
```
FineFindus, line 1:
```suggestion
package app.revanced.patches.sleepasandroid.misc.annotations
```
I think it would be better to rename package to the app name and the category

aliveoutside: yeah, makes sense

oSumAtrIX (approved):

oSumAtrIX:
Thanks

Artaud:
Hi guys, this is Jiri from Urbandroid Team. I see that you just added a patch to unlock Sleep as Android. I'd like to ask you to remove it from further releases.
Please understand that we're independent developers, a three-head studio where Sleep as Android is our major product and almost all of our living depends on its (declining) sales. The price of the app has always been very low for what it does.

oSumAtrIX:
If you like, you can open a PR which reverts it.

Artaud:
Are you willing to then merge it in?

oSumAtrIX:
That's the plan, yeah.

Sculas:
Keep in mind, this will only remove the patch from our repo. People can easily fork this repo and apply the PR themselves, so if you really want to fix this you should look into improving your code so it can't be patched this easily :)

Artaud:
> We're okay with removing the patch. Although, keep in mind this will only remove the patch from our repo. People can easily fork this repo and apply the PR themselves, so if you really want to fix this you should look into improving your code so it can't be patched this easily :)

Thanks :) Yeah, I sure understand that, we're taking this as a lesson as well...

PR #1219, opened 2022-12-07
fix: use correct version

johnconner122:


oSumAtrIX:
The history has been altered manually, probably tags were messed up in the process. I am not sure if changing the version in that file is sufficient.

johnconner122:
> The history has been altered manually, probably tags were messed up in the process. I am not sure if changing the version in that file is sufficient.

oops, then I'll close it.

PR #1232, opened 2022-12-09
docs: init

oSumAtrIX:
## TODO:

- [ ] APIs


alexandreteles:
Well, it's time to start reviewing this.

alexandreteles (approved):
I will not write a review per file because the time required is not at hand. In any case, the documentation is clear enough to be merged. I have some changes to suggest, but they are trivial and wouldn't modify the general model of how docs are being written. All in all, everything looks good and is ready to be included in the main branch.

oSumAtrIX:
@alexandreteles There is currently one TODO left, do you think we should open a separate PR for it and merge as is now?

alexandreteles (changes requested):
My only concern is the `docs/1_introduction.md` file. It is essential not to consider that the user already knows how to include the patcher in his project. Make the `how it works` section a proper Kotlin/Java unit with all the includes and previous steps until the patch bundle loading step.

oSumAtrIX:
Imo that would at max belong to the patched repo. The patches should just quickly outline at which time they are used in the patcher and not explain how to use the patcher.

PR #1236, opened 2022-12-10
feat(twitter): Add predictive back gesture patch

bennett-sh:
This patch enables the new Android 13 predictive back gesture for Twitter.


src/main/kotlin/app/revanced/patches/twitter/misc/predictiveback/patch/PredictiveBackPatch.kt
```diff
@@ -0,0 +1,30 @@
...
+import app.revanced.patcher.patch.PatchResult
+import app.revanced.patcher.patch.PatchResultSuccess
+import app.revanced.patcher.patch.ResourcePatch
+import app.revanced.patcher.patch.annotations.Patch
+import app.revanced.patches.twitter.misc.predictiveback.annotations.PredictiveBackCompatibility
+
+@Patch
+@Name("predictive-back")
+@Description("Enables the new predictive back feature")
+@PredictiveBackCompatibility
```
oSumAtrIX, line 16: Does this patch technically not work with every app? If so, you can remove this constraining annotation. The patch would also require to check if the attribute already exists.

src/main/kotlin/app/revanced/patches/twitter/misc/predictiveback/patch/PredictiveBackPatch.kt
```diff
@@ -0,0 +1,30 @@
...
+@Version("0.0.1")
+class PredictiveBackPatch : ResourcePatch {
```
oSumAtrIX, line 18: Probably `PredictiveBackGesturePatch` would be better here.

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/twitter/misc/predictiveback/patch/PredictiveBackPatch.kt
bennett-sh: Yes, it does. But in which folder should I move it into?

oSumAtrIX (commented):
You forgot to commit the shared directory.

oSumAtrIX (approved):

Vucko130:
image

thegravity98:
Does it have any effect in lower version of Android 12, 11, ....

bennett-sh:
No it doesn't but to my knowledge there's no way to specify the minimum
Android version in Revanced.

kazimmt:
It should only be limited for ReVanced supported apps.

Otherwise the app list inside ReVanced Manager is biGGGGGGGGGGGGG

Screenshot_2022-12-12-01-34-22-596_app revanced manager flutter
Screenshot_2022-12-12-01-34-03-274_app revanced manager flutter
Screenshot_2022-12-12-01-34-10-964_app revanced manager flutter
Screenshot_2022-12-12-01-34-14-711_app revanced manager flutter
Screenshot_2022-12-12-01-34-19-512_app revanced manager flutter

bennett-sh:
I don't think you should limit this patch but instead the manager should only show apps with non-generic patches so if you want to patch any APK, it will work but the app list won't be filled.
Someone has already created an issue [here](https://github.com/revanced/revanced-manager/issues/579).

Canny1913:
> I don't think you should limit this patch but instead the manager should only show apps with non-generic patches so if you want to patch any APK, it will work but the app list won't be filled. Someone has already created an issue [here](https://github.com/revanced/revanced-manager/issues/579).

We already have a few ideas, such as sorting app list based on most amount of patches or adding an "Load more apps" button.

bennett-sh:
That's probably the best solution.

Vucko130:
how to solve that missing compatibility annotation  problem?

Vucko130:
> This is not a problem.

But builder is skipping that patch because of this problem.

oSumAtrIX:
There is no builder.

bennett-sh:
> how to solve that missing compatibility annotation problem?
it's already fixed

Vucko130:
> > how to solve that missing compatibility annotation problem?
> > it's already fixed

Where? You opened a pr to fix that maybe but it's closed.

src/main/kotlin/app/revanced/patches/all/interaction/gestures/patch/PredictiveBackGesturePatch.kt
```diff
@@ -8,26 +8,29 @@ import app.revanced.patcher.patch.PatchResult
...
-            if(application.attributes.getNamedItem("android:enableOnBackInvokedCallback") == null) {
-                val attr = document.createAttribute("android:enableOnBackInvokedCallback")
-                attr.value = "true"
-                application.attributes.setNamedItem(attr)
+            with(document.getElementsByTagName("application").item(0)) {
+                attributes.getNamedItem(FLAG)?.let {
+                    document.createAttribute(FLAG)
+                        .apply { value = "true" }
+                        .let(attributes::setNamedItem)
+                }
```
Sculas, line 26: @oSumAtrIX This looks very incorrect. The previous code checked if `attributes.getNamedItem("android:enableOnBackInvokedCallback")` is `null` before creating the attribute, but your change instead creates the attribute **only if the attribute already exists** (`?.let` means "if not null, run this block")

oSumAtrIX: Its not very incorrect, but the logic should be inverted, please open a PR or issue.

PR #1241, opened 2022-12-11
feat(twitter): Predictive back gesture

bennett-sh:


Canny1913:
This PR is exactly the same as this one
- https://github.com/revanced/revanced-patches/pull/1236

PR #1272, opened 2022-12-15
feat: add google play link for README pkg names

xafn:


oSumAtrIX (approved):

thegravity98:
We can add app icons as leader/before package name. 

https://serpapi.com/google-play-api

Icons are available under "thumbnails" in json file.

PR #1297, opened 2022-12-18
Decrease YT speed increments from 0.25 to 0.125

rorosaurus:
1.25x and 1.5x speed are too fast or slow for some videos. I think power users would want 1.125 and 1.375.

Downside: doubles the options in the already lengthy custom playback speed selector. Maybe in the future we could make the custom playback range/increments customizable in settings? (Or replace the array of selections with +/- buttons.)


oSumAtrIX:
You can already change the granularity via the patch options.

rorosaurus:
> You can already change the granularity via the patch options.

Would you accept a PR that moves these PatchOptions to SettingsPatch.PreferenceScreen? It seems like this would be useful for end-users to be able to adjust at runtime without a full dev environment (eventually).

oSumAtrIX:
Yep, this would be ideal.

PR #1302, opened 2022-12-18
feat(all): add clone app patch

### LINKED ISSUE #1215: feat(patch): "Meta patch" Clone and change icon
### Application

twitch and other apps

### Issue

we cannot clone twitch to test patche without removing the app each time.

### Patch

It will be great if we can have a patch aviable for each patchable app, that allow to make a vanced clone (and eventually change the icon of the app to not uninstall the wrong one ) 

On background, I see this patch like a meta patch always aviable , and specific patch with the possibility to ovewrite or prupoe an othe patch  for a specifics app

### Motivation

Less work, more feature

### Acknowledgements

FineFindus:
Adds a patch to all apps, which allows the `cloning` of an app / installing under a different package name.  Also closes ReVanced/revanced-patches-template#1636.

I would like to keep this as a draft for now / label the patch as experimental, since I haven't done a lot of extensive testing yet, and there are probably a lot of things that I forgot to overwrite.

### DIFF
--- a/src/main/kotlin/app/revanced/patches/all/misc/cloneapp/patch/CloneAppPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/all/misc/cloneapp/patch/CloneAppPatch.kt
@@ -1,0 +1,76 @@
+package app.revanced.patches.all.misc.cloneapp.patch
+
+import app.revanced.patcher.annotation.Description
+import app.revanced.patcher.annotation.Name
+import app.revanced.patcher.annotation.Version
+import app.revanced.patcher.data.ResourceContext
+import app.revanced.patcher.patch.*
+import app.revanced.patcher.patch.PatchResult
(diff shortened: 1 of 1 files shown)


oSumAtrIX:
Cloning an app requires much more. As a reference https://appcloner.app/ can be used.

FineFindus:
After researching this topic a bit more, I realized, that `aapt` has an [option to rename packages](https://developer.android.com/studio/command-line/aapt2#link_options). Maybe this would be a better idea than overwriting the manifest in a patch?

> Cloning an app requires much more.

I've never done something like this, so what else needs to be done?  When testing with simple apps, everything worked so far.

> As a reference https://appcloner.app/ can be used.

I would like to avoid looking at their code for now, to not get another DMCA strike.

FineFindus:
I don't have the time right now to work on this, I might reopen it in the future

PR #1315, opened 2022-12-19
feat(youtube/theme): custom splashscreen color for Android 11/12

0xrxL:


oSumAtrIX:
Does this consider `darkThemeBackgroundColor` and `lightThemeBackgroundColor`?

0xrxL:
> Does this consider `darkThemeBackgroundColor` and `lightThemeBackgroundColor`?

Obviously yes. I tried it before open the PR.

0xrxL:
> Does this consider `darkThemeBackgroundColor` and `lightThemeBackgroundColor`?

The tag ```?android:attr/textColorPrimaryInverse``` ensure the color change from light to dark and vice versa.

oSumAtrIX:
So is it possible to theme it to any arbitrary color with the patch options?

0xrxL:
> So is it possible to theme it to any arbitrary color with the patch options?

No, this is just the amoled version, like the current solution for Android 12 and above that has already been merged.

The custom color versions need of a separate PR, which will modify both Android 11 and below and Android 12 and above.

oSumAtrIX:
Without this PR is it possible to apply colors to the splash screen with the patch options? Regardless of that, this patch would force the user into using either completely dark or light colors which would make it a bug in the patch.

0xrxL:
> Without this PR is it possible to apply colors to the splash screen with the patch options? Regardless of that, this patch would force the user into using either completely dark or light colors which would make it a bug in the patch.

Again...NO. But this PR complements the existing (and already merged) one that changes the splashscreen color from light to amoled dynamically on Android 12 and above (on which the possibility of applying custom colors is missing too).

oSumAtrIX:
If it does not the other PR is subject to be reverted.

0xrxL:
@oSumAtrIX Ok...I tried to use context.xmlEditor to the best of my ability. Take a look at the code and let me know if there are any changes to apply.

Vucko130 (commented):
what about android 13?

0xrxL:
> what about android 13?

To be more clear: it's Android 11 and below and Android 12 and above.

src/main/resources/theme/drawable/quantum_launchscreen_youtube.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<layer-list android:opacity="opaque"
+  xmlns:android="http://schemas.android.com/apk/res/android">
+    <item android:drawable="?android:attr/textColorPrimaryInverse" />
+    <item android:id="@id/launchscreens_product_logo" android:bottom="@dimen/launchscreens_product_logo_bottom">
+        <bitmap android:gravity="center" android:src="@drawable/product_logo_youtube_color_144" />
+    </item>
+</layer-list>
```
inotia00, line 8: Patches do not use this file
You can remove it

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
```diff
@@ -45,9 +43,66 @@ class ThemePatch : ResourcePatch {
...
+
+                    if (node.getAttribute("name") == "Base.Theme.YouTube.Launcher") {
+                        newElement.appendChild(createTextNode("?attr/splashScreenColor"))
+
+                        node.appendChild(newElement)
+                    }
+                }
+            }
+        }
+        context.xmlEditor["res/drawable/quantum_launchscreen_youtube.xml"].use { editor ->
```
inotia00, line 100: You should probably include `drawable` and `drawable-sw600dp`
(`drawable-sw600dp` is used on tablets)

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
```diff
@@ -45,9 +43,68 @@ class ThemePatch : ResourcePatch {
...
+                    val node = resourcesNode.childNodes.item(i) as? Element ?: continue
+
+                    if (node.getAttribute("name") == "Base.Theme.YouTube.Launcher") {
+                        newElement.appendChild(createTextNode("?attr/splashScreenColor"))
+
+                        node.appendChild(newElement)
+                    }
+                }
+            }
+        }
```
oSumAtrIX, line 99: Quite a lot of duplicate code as well as boilerplate. Additionally the scope function `with` servers no purpose here.

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
```diff
@@ -45,9 +43,68 @@ class ThemePatch : ResourcePatch {
...
+                        "Base.Theme.YouTube.Launcher.Light" -> {
+                            newElement.appendChild(createTextNode(lightThemeBackgroundColor));
+
+                            node.appendChild(newElement)
+                        }
+                    }
+                }
+            }
+        }
+        context.xmlEditor["res/values-v31/styles.xml"].use { editor ->
```
oSumAtrIX, line 82: Constants should be used for the resource files.

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
```diff
@@ -45,9 +43,68 @@ class ThemePatch : ResourcePatch {
...
+        }
+        context.xmlEditor["res/values/styles.xml"].use { editor ->
+            with(editor.file) {
+                val resourcesNode = getElementsByTagName("resources").item(0) as Element
+
+                for (i in 0 until resourcesNode.childNodes.length) {
+                    val node = resourcesNode.childNodes.item(i) as? Element ?: continue
+
+                    val newElement: Element = createElement("item")
+                    newElement.setAttribute("name", "splashScreenColor")
```
oSumAtrIX, line 65: The scope function `apply` can be used for code such as this.

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
```diff
@@ -45,9 +43,68 @@ class ThemePatch : ResourcePatch {
...
+                            newElement.appendChild(createTextNode(darkThemeBackgroundColor))
+
+                            node.appendChild(newElement)
+                        }
+                        "Base.Theme.YouTube.Launcher.Light" -> {
+                            newElement.appendChild(createTextNode(lightThemeBackgroundColor));
+
+                            node.appendChild(newElement)
+                        }
+                    }
```
oSumAtrIX, line 78: You can reduce duplicate code here because `when` is an expression which you can return from the color.

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
```diff
@@ -45,9 +43,68 @@ class ThemePatch : ResourcePatch {
...
+
+                    if (node.getAttribute("name") == "Base.Theme.YouTube.Launcher") {
+                        newElement.appendChild(createTextNode("?attr/splashScreenColor"))
+
+                        node.appendChild(newElement)
+                    }
+                }
+            }
+        }
+        arrayOf("drawable", "drawable-sw600dp").forEach { drawablePath ->
```
oSumAtrIX, line 100: These are not all sizes, theres more dpis.

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
```diff
@@ -45,9 +43,68 @@ class ThemePatch : ResourcePatch {
...
+            context.xmlEditor["res/$drawablePath/quantum_launchscreen_youtube.xml"].use { editor ->
+                val resourcesNode = editor.file.getElementsByTagName("item").item(0) as Element
+
+                if (resourcesNode.attributes.getNamedItem("android:drawable") != null)
```
oSumAtrIX, line 104: Why is this check required here?

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
0xrxL: To make sure the line contains the ```android:drawable``` tag, before proceeding.

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
0xrxL: With a list of names?

oSumAtrIX: As an example `values-v31` is not self-explanatory and contributes against readability. A constant for it or the entire resource file path `res/values-v31/styles.xml` should be used. This should be done with the other non self-explanatory strings found in this patch.

0xrxL: Well so... A single string field that contains the path, with a dedicated comment to explain to what does this file correspond to.

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
0xrxL: A syntax example of this?

oSumAtrIX: [Here](https://kotlinlang.org/docs/scope-functions.html#apply) is a dedicated example.

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
oSumAtrIX: A comment is not necessary, the constant field name can explain it.

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
0xrxL: Good. Thanks.

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
0xrxL: > These are not all sizes, theres more dpis.

I checked better if there were other "quantum_launchscreen" files, but I didn't find any, except in the already listed folders.

oSumAtrIX: This is because the apk file you used does not have them, but depending on screen size you will get different different apks from play store.

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
0xrxL: In this case I can't do anything else, because I have no way to access these files and decompile them to find the folder names.

0xrxL: After thinking about it for a while, I thought it would be better for the readability of the code to create a list of "paths" with comments alongside

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
0xrxL: In this case I can't do anything else, because I have no way to access these files and decompile them to find the folder names.

oSumAtrIX: The folders exist:

image

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
```diff
@@ -25,30 +23,105 @@ class ThemePatch : ResourcePatch {
...
+                                else -> {"null"}
+                            }
+                        )
+                    )
+
+                    if (this.textContent != "null")
+                        node.appendChild(this)
+                }
+            }
+        }
```
oSumAtrIX, line 87: A lot here is duplicated with the code below. Please reduce it.

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
0xrxL: Ok so...I'll add all these to the list👍

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
0xrxL: What do you mean exactly with "duplicated"? You're referring to the names or other things?

oSumAtrIX: [This lines](https://github.com/revanced/revanced-patches/pull/1315/files#diff-5a51aa8d702357ed352218e03738bda901cd9181c815090f201b6060b753dddeR63-R68) are duplicate with [these](https://github.com/revanced/revanced-patches/pull/1315/files#diff-5a51aa8d702357ed352218e03738bda901cd9181c815090f201b6060b753dddeR89-R94) as an example.

0xrxL: Should be good now. I squashed the cloned code into a single one.

0xrxL:
Sorry for all these commits, but I made some mistakes.

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
```diff
@@ -24,29 +22,123 @@ class ThemePatch : ResourcePatch {
         val darkThemeBackgroundColor = darkThemeBackgroundColor!!
         val lightThemeBackgroundColor = lightThemeBackgroundColor!!
 
-        context.xmlEditor["res/values/colors.xml"].use { editor ->
-            val resourcesNode = editor.file.getElementsByTagName("resources").item(0) as Element
+        val valuesPaths: List<String> = listOf(
+            "values", // Android 11 (and below)
+            "values-v31" // Android 12 (and above)
+        )
```
oSumAtrIX, line 28: These should be two separate constants to prevent odd code [such as this](https://github.com/revanced/revanced-patches/pull/1315/files#diff-5a51aa8d702357ed352218e03738bda901cd9181c815090f201b6060b753dddeR31). You would create an array out of the constants later below [here](https://github.com/revanced/revanced-patches/pull/1315/files#diff-5a51aa8d702357ed352218e03738bda901cd9181c815090f201b6060b753dddeR84).

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
```diff
@@ -24,29 +22,123 @@ class ThemePatch : ResourcePatch {
...
+
+        // edit the resource color file to change the theme color
```
oSumAtrIX, line 30: What do you mean by "theme color"? Please be more concise.

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
```diff
@@ -24,29 +22,123 @@ class ThemePatch : ResourcePatch {
...
+            "drawable-night",
+            "drawable-nodpi",
+            "drawable-sw600dp",
+            "drawable-sw600dp-xhdpi",
+            "drawable-sw720dp-xhdpi",
+            "drawable-watch",
+            "drawable-xhdpi",
+            "drawable-xxhdpi",
+            "drawable-xxxhdpi"
         )
```
oSumAtrIX, line 70: Is `drawable-watch` even responsible for the splash screen? The same goes for the other resource directories. Are they all necessary?

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
```diff
@@ -24,29 +22,123 @@ class ThemePatch : ResourcePatch {
...
 
+        context.xmlEditor[attrsPath].use { editor ->
+            val file = editor.file
+
+            val resourceNode = file.getElementsByTagName("resources").item(0) as Element
```
oSumAtrIX, line 75: This is still duplicate to [this](https://github.com/revanced/revanced-patches/pull/1315/files#diff-5a51aa8d702357ed352218e03738bda901cd9181c815090f201b6060b753dddeR33). Instead, create a method, which accepts a [high-order function](https://kotlinlang.org/docs/lambdas.html) as a transformer to reduce duplicate code.

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
```diff
@@ -24,29 +22,123 @@ class ThemePatch : ResourcePatch {
...
+
+            resourceNode.appendChild(
+                file.createElement("attr").apply {
+                    setAttribute("format", "reference")
+                    setAttribute("name", "splashScreenColor")
+                }
+            )
```
oSumAtrIX, line 82: Please document this, it is unknown what this code does or why it is required.

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
```diff
@@ -24,29 +22,123 @@ class ThemePatch : ResourcePatch {
...
+                                        else -> "null"
+                                    }
+                                    else -> "null"
+                                }
+                            )
+                        )
+
+                        if (this.textContent != "null")
+                            node.appendChild(this)
+                    }
```
oSumAtrIX, line 125: Please add explanatory comments on why this is required, and what purpose it serves.

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
```diff
@@ -24,29 +22,123 @@ class ThemePatch : ResourcePatch {
...
+            val quantumLaunchscreenPath = "res/$drawablePath/quantum_launchscreen_youtube.xml"
+
+            if (context[quantumLaunchscreenPath].exists()) {
+                context.xmlEditor[quantumLaunchscreenPath].use { editor ->
+                    val resourcesNode = editor.file.getElementsByTagName("item").item(0) as Element
+
+                    if (resourcesNode.attributes.getNamedItem("android:drawable") != null)
+                        resourcesNode.setAttribute("android:drawable", "?attr/splashScreenColor")
+                }
+            }
```
oSumAtrIX, line 139: RE: https://github.com/revanced/revanced-patches/pull/1315/files#r1057310636

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
```diff
@@ -24,29 +22,123 @@ class ThemePatch : ResourcePatch {
...
+                }
+            )
+        }
+        valuesPaths.forEachIndexed { pathIndex, valuesPath ->
+            val stylesPath = "res/$valuesPath/styles.xml"
+
+            context.xmlEditor[stylesPath].use { editor ->
+                val file = editor.file
+
+                val childNodes = (file.getElementsByTagName("resources").item(0) as Element).childNodes
```
oSumAtrIX, line 90: RE: https://github.com/revanced/revanced-patches/pull/1315/files#r1057310258

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
0xrxL: As I said on a previous request: I don't know. I simply added all the folder names you posted to me with a screenshot. On the apk that I managed to find only ```drawable``` and ```drawable-sw600``` contain the ```quantum_launchscreen``` file

oSumAtrIX: It is unclear if all of them are required. It depends on what the purpose of these separate resource directories is. The folders I screenshotted and their purpose should've separately been investigated to confirm whether they are required to be considered in this patch. For that, you will need an APK which is "anydpi".

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
0xrxL: This code is useful to create a a new xml line inside ```attr.xml``` (with all required attributes of other lines inside the file), named ```splashScreenColor```

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
oSumAtrIX: What do you mean by "App colors"?

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
oSumAtrIX: It is unnecessary to explain that it creates a new line in the `attr.xml` file. Explain why it is required and what for purpose.

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
0xrxL: This needs to add change the value of field named ```splashScreenColor``` inside the file ```attr.xml```.

But to prevent a line from being added after ```Base.Theme.YouTube.Launcher```(check the code), I decided to add this only when a specific file is indexed by ```ForEachIndexed``` method

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
0xrxL: > What do you mean by "App colors"?

I am referring to app theme, theme colors, like amoled theme. How would you define it?

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
0xrxL: > It is unclear if all of them are required. It depends on what the purpose of these separate resource directories is. The folders I screenshotted and their purpose should've separately been investigated to confirm whether they are required to be considered in this patch. For that, you will need an APK which is "anydpi".

Wouldn't it be better to add these later, based on user feedback?

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
oSumAtrIX: > This needs to add change the value of field named splashScreenColor inside the file attr.xml.

Aside from the semantic issues I have with this sentence, you are supposed to comment on what the purpose of the code is and why it is required and not explain what the code does. We can simply read the code and see that it changes a field named `splashScreenColor`, but no comment explains the purpose or why it is required. I suggest looking up (this)[https://stackoverflow.blog/2021/12/23/best-practices-for-writing-code-comments/].

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
0xrxL: > It is unnecessary to explain that it creates a new line in the `attr.xml` file. Explain why it is required and what for purpose.

I'll add a comment inside the patch then.

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
0xrxL: > > This needs to add change the value of field named splashScreenColor inside the file attr.xml.
> 
> Aside from the semantic issues I have with this sentence, you are supposed to comment on what the purpose of the code is and why it is required and not explain what the code does. We can simply read the code and see that it changes a field named `splashScreenColor`, but no comment explains the purpose or why it is required. I suggest looking up [this](https://stackoverflow.blog/2021/12/23/best-practices-for-writing-code-comments/).

Yeah...sorry. There was a syntax error (now fixed) in the comment.

By the way ok...I'll add a comment for this too.

oSumAtrIX: Sounds good, make sure to follow the link I sent earlier, it is quite useful.

0xrxL: > Sounds good, make sure to follow the link I sent earlier, it is quite useful.

I'll add a comment with the same style of ones added on [this](https://github.com/revanced/revanced-integrations/commit/9ab8a646ed07d709c46fe7b5dd3238bc23301b8b) merged code

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
0xrxL: I'll leave this request open until you decide how to proceed.

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
```diff
@@ -9,11 +9,10 @@ import app.revanced.patcher.patch.annotations.DependsOn
 import app.revanced.patcher.patch.annotations.Patch
 import app.revanced.patches.youtube.layout.theme.annotations.ThemeCompatibility
 import app.revanced.patches.youtube.misc.manifest.patch.FixLocaleConfigErrorPatch
-import app.revanced.util.resources.ResourceUtils
-import app.revanced.util.resources.ResourceUtils.copyResources
+import org.w3c.dom.Document
 import org.w3c.dom.Element
 
-@Patch(include = false)
+@Patch(include = true)
```
oSumAtrIX, line 15:
```suggestion
@Patch
```

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
```diff
@@ -24,28 +23,164 @@ class ThemePatch : ResourcePatch {
...
+                 */
+                if (this.textContent != "null")
+                    indexedNode.appendChild(this)
+            }
+        }
+        fun changeSplashScreenBackgroundColor(rootNode: Element) {
+            /**
+             * Check if a node element with attribute 'android:drawable' exists
+             * then change its value to '?attr/splashScreenColor'
+             */
```
oSumAtrIX, line 122: This comment is useless as it simply duplicates code as prescribed in rule 1 [here](https://stackoverflow.blog/2021/12/23/best-practices-for-writing-code-comments).

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
```diff
@@ -24,28 +23,164 @@ class ThemePatch : ResourcePatch {
...
+                        }
+                    )
+                )
 
-                    "yt_white1", "yt_white1_opacity95", "yt_white1_opacity98", "yt_white2", "yt_white3", "yt_white4",
-                     -> lightThemeBackgroundColor
+                /**
+                 * Add the created node element after 'indexedNode', into xml file
+                 * only if the value of new node element is not equal to 'null'
+                 */
```
oSumAtrIX, line 113: RE: https://github.com/revanced/revanced-patches/pull/1315/files#r1057996961

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
```diff
@@ -24,28 +23,164 @@ class ThemePatch : ResourcePatch {
...
+                    }
+                )
 
-                node.textContent = when (node.getAttribute("name")) {
-                    "yt_black0", "yt_black1", "yt_black1_opacity95", "yt_black1_opacity98", "yt_black2", "yt_black3",
-                    "yt_black4", "yt_status_bar_background_dark", "material_grey_850" -> darkThemeBackgroundColor
+                /**
+                 * Fill the value of created node element with the field 'lightThemeBackgroundColor' or 'darkThemeBackgroundColor'
+                 * only if the name of 'indexedNode' is equal to the field 'nodeAttributeName'
+                 */
```
oSumAtrIX, line 92: RE: https://github.com/revanced/revanced-patches/pull/1315/files#r1057996961

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
```diff
@@ -24,28 +23,164 @@ class ThemePatch : ResourcePatch {
...
-                val node = resourcesNode.childNodes.item(i) as? Element ?: continue
+            /**
+             * Create a new node element of type 'item'
+             * with the attribute: 'name', and its value based on the field 'filePathIndex':
+             *
+             *      '/values/styles.xml' = 'splashScreenColor'
+             *      '/values-v31/styles.xml' = 'android:windowSplashScreenBackground'
+             *
+             * @field filePathIndex The current styles.xml path, indexed by the method 'forEachIndexed()'
+             */
```
oSumAtrIX, line 78: RE: https://github.com/revanced/revanced-patches/pull/1315/files#r1057996961

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
```diff
@@ -24,28 +23,164 @@ class ThemePatch : ResourcePatch {
...
+             * with attributes: 'format' (and its value = 'reference') & 'name' (and its value = 'splashScreenColor').
+             */
+            val newColorAttribute: Element = file.createElement("attr").apply {
+                setAttribute("format", "reference")
+                setAttribute("name", splashScreenColorAttributeName)
+            }
+
+            /**
+             * Add created node element after 'rootNode', into xml file
+             */
```
oSumAtrIX, line 64: RE: https://github.com/revanced/revanced-patches/pull/1315/files#r1057996961

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
```diff
@@ -24,28 +23,164 @@ class ThemePatch : ResourcePatch {
...
+                "yt_white3", "yt_white4" -> lightThemeBackgroundColor
+
+                else -> nodeAttribute
+            }
+        }
+        fun addNewColorAttribute(rootNode: Element, file: Document) {
+            /**
+             * Create a new node element of type 'attr'
+             * with attributes: 'format' (and its value = 'reference') & 'name' (and its value = 'splashScreenColor').
+             */
```
oSumAtrIX, line 56: RE: https://github.com/revanced/revanced-patches/pull/1315/files#r1057996961

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
0xrxL: So in this case I should remove the duplicate comments, or deduplicate the code together with comments?

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
```diff
@@ -24,28 +23,164 @@ class ThemePatch : ResourcePatch {
...
+            "drawable-sw600dp"
+        )
+        val splashScreenColorAttributeName = "splashScreenColor"
+
+        fun changeAppBackgroundColorValues(indexedNode: Element) {
+            /**
+             * Get all node elements with a specific value (example: "yt_white1" or "yt_black1")
+             * then change it with the content of field 'lightThemeBackgroundColor' or 'darkThemeBackgroundColor'
+             */
+            val nodeAttribute = indexedNode.getAttribute("name")
```
oSumAtrIX, line 39: Why not simply call this `name`?

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
```diff
@@ -24,28 +23,164 @@ class ThemePatch : ResourcePatch {
...
+            "drawable",
+            "drawable-sw600dp"
+        )
+        val splashScreenColorAttributeName = "splashScreenColor"
+
+        fun changeAppBackgroundColorValues(indexedNode: Element) {
+            /**
+             * Get all node elements with a specific value (example: "yt_white1" or "yt_black1")
+             * then change it with the content of field 'lightThemeBackgroundColor' or 'darkThemeBackgroundColor'
+             */
```
oSumAtrIX, line 38:
```suggestion
            // Replace the colors responsible for the background color to the specified new color.
```

oSumAtrIX, line 38:
```suggestion
            // Replace the colors responsible for the background color with the specified new color.
```

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
```diff
@@ -24,28 +23,164 @@ class ThemePatch : ResourcePatch {
...
 
-        context.xmlEditor["res/values/colors.xml"].use { editor ->
-            val resourcesNode = editor.file.getElementsByTagName("resources").item(0) as Element
+        val android11ValuesFolder = "values"
+        val android12ValuesFolder = "values-v31"
+        val drawableFolders: List<String> = listOf(
+            "drawable",
+            "drawable-sw600dp"
+        )
+        val splashScreenColorAttributeName = "splashScreenColor"
```
oSumAtrIX, line 32: These should be constants.

oSumAtrIX, line 32: Why is this a field? The field name already explains what it is.

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
```diff
@@ -24,28 +23,164 @@ class ThemePatch : ResourcePatch {
         val darkThemeBackgroundColor = darkThemeBackgroundColor!!
         val lightThemeBackgroundColor = lightThemeBackgroundColor!!
 
-        context.xmlEditor["res/values/colors.xml"].use { editor ->
-            val resourcesNode = editor.file.getElementsByTagName("resources").item(0) as Element
+        val android11ValuesFolder = "values"
+        val android12ValuesFolder = "values-v31"
+        val drawableFolders: List<String> = listOf(
```
oSumAtrIX, line 28: Fields should be private to this class.

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
```diff
@@ -24,28 +23,164 @@ class ThemePatch : ResourcePatch {
...
+            if (rootNode.attributes.getNamedItem("android:drawable") != null)
+                rootNode.setAttribute("android:drawable", "?attr/splashScreenColor")
+        }
+
+        fun editXMLFile(
+            filePath: String,
+            filePathIndex: Int,
+            elementsToSearchName: String,
+            patchIndex: Int
+        ) {
```
oSumAtrIX, line 132: This method's signature is unreasonable. What is `filePathIndex` or `patchIndex`? What does it have to do with `editXMLFile` as the method name conveys?

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
oSumAtrIX: No, a comment according to [this](https://stackoverflow.blog/2021/12/23/best-practices-for-writing-code-comments) should be added. I also gave an example [here](https://github.com/revanced/revanced-patches/pull/1315#discussion_r1057997631).

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
0xrxL: ```filePathIndex``` it is used to apply two distinct patches to the style.xml files.

```patchIndex``` it is used to choose which patch to apply.

But already there are comments that explain it.

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
0xrxL: There are some methods that use this name as parameters (3 in total). I decided to convert it to a field to deduplicate the code.

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
oSumAtrIX: This is too arbitrary, simplify and split your method reasonably.

src/main/kotlin/app/revanced/patches/youtube/layout/theme/patch/ThemePatch.kt
oSumAtrIX: If there are multiple methods using this same attribute, then the code should be deduplicated. A class field does not make sense for this.

0xrxL:
https://github.com/revanced/revanced-patches/issues/1377

PR #1353, opened 2022-12-23
refactor(youtube/vertical-scroll-patch): patch canScrollVertically() …

0xrxL:
…only

It's pointless to patch the entire method to return a false value. Only ```canScrollVertically()``` method should be patched instead, so as to leave the ```if(view instanceof ListView)``` condition untouched (which could be useful to prevent bugs of the swipe to reload gesture).

### DIFF
--- a/src/main/kotlin/app/revanced/patches/youtube/misc/fix/verticalscroll/fingerprints/CanScrollVerticallyFingerprint.kt
+++ b/src/main/kotlin/app/revanced/patches/youtube/misc/fix/verticalscroll/fingerprints/CanScrollVerticallyFingerprint.kt
@@ -8,12 +8,6 @@ object CanScrollVerticallyFingerprint : MethodFingerprint(
     "Z",
     parameters = emptyList(),
     opcodes = listOf(
-        Opcode.IGET_OBJECT,
-        Opcode.INSTANCE_OF,
-        Opcode.CONST_4,
-        Opcode.IF_EQZ,
-        Opcode.CHECK_CAST,
-        Opcode.INVOKE_STATIC,
         Opcode.MOVE_RESULT,
         Opcode.RETURN,
         Opcode.INVOKE_VIRTUAL,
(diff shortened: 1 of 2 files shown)


0xrxL:
Immagine 2022-12-23 222249

src/main/kotlin/app/revanced/patches/youtube/misc/fix/verticalscroll/fingerprints/CanScrollVerticallyFingerprint.kt
```diff
@@ -7,6 +7,17 @@ import org.jf.dexlib2.Opcode
...
+        Opcode.IGET_OBJECT,
+        Opcode.INSTANCE_OF,
+        Opcode.CONST_4,
+        Opcode.IF_EQZ,
+        Opcode.CHECK_CAST,
+        Opcode.INVOKE_STATIC,
+        Opcode.MOVE_RESULT,
+        Opcode.RETURN,
+        Opcode.INVOKE_VIRTUAL,
+        Opcode.MOVE_RESULT,
```
oSumAtrIX, line 20: Wouldn't these two suffice? The pattern is longer than necessary.

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/youtube/misc/fix/verticalscroll/fingerprints/CanScrollVerticallyFingerprint.kt
0xrxL: It is recommended, because I need them to get the registry value and the index where place the patch.

0xrxL: EDIT: ok...I reduced the fingerprint, and patch still works...also if I'm not 100% sure it is a good idea (because a too small a fingerprint could cause problems in the future imho).

oSumAtrIX: The class for this method is constrained by `customFingerprint`. I believe this would suffice for this fingerprint.

oSumAtrIX (approved):

src/main/kotlin/app/revanced/patches/youtube/misc/fix/verticalscroll/fingerprints/CanScrollVerticallyFingerprint.kt
0xrxL: > The class for this method is constrained by `customFingerprint`. I believe this would suffice for this fingerprint.

Ok...done

Issue #430, opened 2022-09-04
bug: Auto play doesn't work if the auto play button is not enabled


ASDosjani:
### Type

Other

### Bug description

Title says it all.

### Steps to reproduce

Turn off auto play button.
Enable auto play.

### Relevant log output

```shell
I don't have.
```

oSumAtrIX:
Autoplay is disabled if you disable the autoplay button.

ASDosjani:
> Autoplay is disabled if you disable the autoplay button.

Than why the autoplay switcb is available in settings if I disable the button? It doesn't make sense.

oSumAtrIX:
Both are separate patches, thats why.

ASDosjani:
> Both are separate patches, thats why.

Can't these patches be combined?

oSumAtrIX:
One disables the button and it's functionality of auto playing, the other makes it enabled by default, so if the button were disabled, how would you want to disable auto play?

ASDosjani:
> One disables the button and it's functionality of auto playing, the other makes it enabled by default, so if the button were disabled, how would you want to disable auto play?

In the default YouTube settings.
3rd option in settings. That's what I'm talking about the whole time.

oSumAtrIX:
Responsible code for this: https://github.com/revanced/revanced-patches/blob/217c7864d90b88810bba7fc95053d1edff7d7244/src/main/kotlin/app/revanced/patches/youtube/layout/autoplaybutton/patch/HideAutoplayButtonPatch.kt#L78

Getting rid of it would allow autoplayback even when the button is hidden.

ASDosjani:
Yes, but at least you could change it in the settings. I think most people doesn't use the button but the autoplay is useful when listening to music. I'm not familiar with this code so I can't make the fix.

ILoveOpenSourceApplications:
This is a very confusing issue. And with the option to toggle autoplay switch on/off present in the ReVanced settings, please try to ignore the autoplay setting by YouTube. Closing it now.

PR #1357, opened 2022-12-24
fix(youtube/fix-playback): stop playlist auto skips

inotia00:
Current `fix-playback` patch does not work properly with playlists https://github.com/revanced/revanced-patches/issues/1206

To be precise, the patch itself works fine, but there is an issue with playlists being skipped automatically

This PR fixes an issue where playlists are automatically skipped by the `fix-playback` patch

Unfortunately, since the playback issue is not reproduced on my device, I have not been able to test whether this PR works normally while fixing the issue that occurs in the UK region.

So I'll mark the PR as draft.

What I checked
- Playlists are no longer automatically skipped when `Fix video playback issues` is set to ON

Parts that require confirmation
- When `Fix video playback issues` is set to ON, the existing action works as intended, but the playlists auto skip issue is resolved.
(need an account that can reproduce the playback issue)

If this patch is not confirmed to be working properly by the maintainers, please leave a comment and close the PR.

### DIFF
--- a/src/main/kotlin/app/revanced/patches/youtube/misc/fix/playback/patch/FixPlaybackPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/youtube/misc/fix/playback/patch/FixPlaybackPatch.kt
@@ -5,19 +5,25 @@ import app.revanced.patcher.annotation.Description
 import app.revanced.patcher.annotation.Name
 import app.revanced.patcher.annotation.Version
 import app.revanced.patcher.data.BytecodeContext
+import app.revanced.patcher.extensions.addInstructions
 import app.revanced.patcher.extensions.removeInstruction
+import app.revanced.patcher.extensions.instruction
+import app.revanced.patcher.fingerprint.method.impl.MethodFingerprint.Companion.resolve
 import app.revanced.patcher.patch.PatchResult
 import app.revanced.patcher.patch.PatchResultSuccess
 import app.revanced.patcher.patch.BytecodePatch
 import app.revanced.patcher.patch.annotations.DependsOn
+import app.revanced.patcher.util.smali.ExternalLabel
 import app.revanced.patches.shared.settings.preference.impl.StringResource
 import app.revanced.patches.shared.settings.preference.impl.SwitchPreference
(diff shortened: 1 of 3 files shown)


src/main/kotlin/app/revanced/patches/youtube/misc/fix/playback/patch/FixPlaybackPatch.kt
```diff
@@ -44,6 +49,16 @@ class FixPlaybackPatch : ResourcePatch {
             )
         )
 
+        val result = VideoEndListenerFingerprint.result ?: return VideoEndListenerFingerprint.toErrorResult()
```
oSumAtrIX, line 52: This patch should only disable auto-play for a short time when the integrations skip the video to the end and enable autoplay again once it skips back to the initial position. Right now this PR will break auto-play functionality. And even when the fix is not enabled.

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/youtube/misc/fix/playback/patch/FixPlaybackPatch.kt
inotia00: In case I missed something, I tested it in various situations.

It's not because of this PR that auto-play is broken
It is due to an existing bug: https://github.com/revanced/revanced-patches/issues/430

If you change the `Hide autoplay button` to off, or exclude the `hide-autoplay-button` patch, the issue does not reproduce.

oSumAtrIX: The `hide-autoplay-button` does not prevent playlists from being auto-played iirc. With this PR it is now impossible to auto-play playlists even if the switch `revanced_fix_playback` is disabled.

inotia00:
Unfortunately, due to the limitations of my abilities, I couldn't implement it more ideally.

Instead, I added TODO as comments, so anyone who wants to complete this PR on my behalf may refer to it.

Issue #615, opened 2022-09-24
bug: Can not exit the app by pressing back button or using the back gesture.


ANi1XO:
### Type

Error at runtime

### Bug description

usually to exit the app, i just swipe back (gesture). but on revanced patch 2.69.0 when i try to exit the app this way, nothing happens at all. only way i can exit the app now is by pressing home button (swipe up gesture). i can see several users are also getting this bug.
Im on youtube version 17.36.37 cli v2.11.2 patches v2.69.0 (same issue can be seen on patch 2.60.0)

### Steps to reproduce

open app
Open revanced settings
Open layout settings
Change anything in layout settings that will prompt a restart.
Once the app restarts the bug is introduced and pressing back to exit app will stop working.

### Relevant log output

```shell
Null
```

### Acknowledgements

- [X] I have searched the existing issues and this is a new and no duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I filled out all of the requested information in this issue properly.

webzeroacs:
This is a known issue. I think Remove general ads is the problem. If you change this part to general ads are shown, the back button will exit well. If you turn off Remove general ads, there will be Home ads and ads under the video.

killerlearner:
Yeah check #599

rewqazxv:
Pull-to-refresh also doesn't work when you can't exit, caused by Remove general ads too

ANi1XO:
With patch 2.81.0, there is a way to circumvent this bug. Do a fresh install, go to settings, change settings to your desire,

SodaWithoutSparkles:
Can reproduce the issue. My current method of exiting is to open recent tasks and kill the app

Kalpataru6:
Yeah.. 3 button nevigation's back button also don't exist the app .

PR #1369, opened 2022-12-26
feat(youtube): `double-back-to-close` patch

inotia00:
Make it possible to close the app by double-pressing the back button in the home feed

Also, it solves this issue https://github.com/revanced/revanced-patches/issues/615

depends on:
https://github.com/revanced/revanced-integrations/pull/251

### DIFF
--- a/src/main/kotlin/app/revanced/patches/youtube/misc/doublebacktoclose/annotations/DoubleBackToCloseCompatibility.kt
+++ b/src/main/kotlin/app/revanced/patches/youtube/misc/doublebacktoclose/annotations/DoubleBackToCloseCompatibility.kt
@@ -1,0 +1,13 @@
+package app.revanced.patches.youtube.misc.doublebacktoclose.annotations
+
+import app.revanced.patcher.annotation.Compatibility
+import app.revanced.patcher.annotation.Package
+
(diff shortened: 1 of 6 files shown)


Sculas:
This doesn't add a feature but rather fixes original behavior from a broken patch, so shouldn't this be incorporated into the general-ads patch (without the settings)?

inotia00:
It's slightly different from making the existing back button work.
(I think this is one of the alternative ways to solve the issue)

So I didn't incorporate it and separated it into a new patch.
But if you think incorporating would be more ideal, I'll edit that.

Sculas:
@oSumAtrIX What is your stance on this?

oSumAtrIX:
I have commented in https://github.com/revanced/revanced-integrations/pull/251.

Sculas:
> I have commented in [revanced/revanced-integrations#251](https://github.com/revanced/revanced-integrations/pull/251).

That's about something completely different than what I was talking about. Please read my original comment again.

oSumAtrIX:
@Sculas No, it is not about something completely different. It is my stance on this PR as you have asked.

Sculas:
That's.. not what I meant :P

PR #1370, opened 2022-12-26
ReturnYouTubeDislike UI visual improvements, foreign languages fix 

LisoUseInAIKyrios:
changes for https://github.com/revanced/revanced-integrations/pull/252


src/main/kotlin/app/revanced/patches/youtube/layout/returnyoutubedislike/resource/patch/ReturnYouTubeDislikeResourcePatch.kt
```diff
@@ -13,6 +13,8 @@ import app.revanced.patches.shared.settings.preference.impl.StringResource
 import app.revanced.patches.youtube.layout.returnyoutubedislike.annotations.ReturnYouTubeDislikeCompatibility
 import app.revanced.patches.youtube.misc.manifest.patch.FixLocaleConfigErrorPatch
 import app.revanced.patches.youtube.misc.settings.bytecode.patch.SettingsPatch
+import app.revanced.util.resources.ResourceUtils
+import app.revanced.util.resources.ResourceUtils.copyResources
```
FineFindus, line 17: Is there a reason, these are added, but nothing else is changed in the file?

Domiiniik:
One more thing for RYD: How about RYD patch removing the white "Your opinion has been sent to the creator" popub box when dropping a dislike, since we see dislikes with the extension anyways.

src/main/kotlin/app/revanced/patches/youtube/layout/returnyoutubedislike/resource/patch/ReturnYouTubeDislikeResourcePatch.kt
LisoUseInAIKyrios: I had some logic to detect right to left language, by using a resource entry.  

I later removed the resource entries, and these imports are left over.

LisoUseInAIKyrios:
> One more thing for RYD: How about RYD patch removing the white "Your opinion has been sent to the creator" popub box when dropping a dislike, since we see dislikes with the extension anyways.

Not a bad idea.  Although I think the 'feedback shared with creator' popup is only displayed a few times and then its never shown again.

oSumAtrIX:
That is true, also removing it would be an unintended behavior. It is intended design to let the user know that feedback has been sent regardless of the RYD patch.

Domiiniik:
> That is true, also removing it would be an unintended behavior. It is intended design to let the user know that feedback has been sent regardless of the RYD patch.

Why it doesn't show that feedback has been sent if you like the video then through? And btw, this popup was added along with the dislike removal, I don't think both of these facts are a coincidence, It was clearly added since you can't normally see dislike count to not confuse user if this button even does anything after dislike count removal.
Also this popup shows every time, not just a few

0xrxL:
@LisoUseInAIKyrios Any update about this?

LisoUseInAIKyrios:
Yes.  I changed it to 3 separators:

3 separators

LisoUseInAIKyrios:
It's ready to go

oSumAtrIX (approved):

oSumAtrIX:
Thanks!

Issue #531, opened 2022-09-18
Copy link and Copy link with timestamp


Black-StellarSpectrum:
### Application

Youtube 

### Issue

No way to copy link with timestamp

### Patch

It'd be nice to directly copy a link to timestamp to share a particular part of the video if possible.
Like long pressing share will share link with timestamp!

### Motivation

Old YouTube Vanced

NewtonChutney:
Ohh moderators!! Please please pleeeaasseee prioritize this patch!  🥺🥺🥺

sawankumarbundelkhandi:
@oSumAtrIX I think this issue can be closed as it's already addressed by https://github.com/revanced/revanced-patches/pull/1402 with a button on video player to copy with timestamp.

PR #1402, opened 2022-12-30
feat(youtube): `copy-video-url` patch

aliernfrog:
Depends on https://github.com/revanced/revanced-patches/pull/1401

Should be merged with https://github.com/revanced/revanced-integrations/pull/263

Related issue: https://github.com/revanced/revanced-patches/issues/531

Adds copy video & copy timestamp URL buttons in video player.
IMG_20221230_234907


src/main/kotlin/app/revanced/patches/youtube/interaction/copyvideourl/bytecode/patch/CopyVideoUrlBytecodePatch.kt
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    override fun execute(context: BytecodeContext): PatchResult {
+        // Hook video time
+        with(VideoInformationPatch) {
+            videoTimeHook(
+                INTEGRATIONS_PATCH_UTILS_DESCRIPTOR,
+                "setVideoTime"
+            )
+        }
```
oSumAtrIX, line 47: The function `with` is not necessary here as the new scope is only used for a single function.

src/main/kotlin/app/revanced/patches/youtube/interaction/copyvideourl/bytecode/patch/CopyVideoUrlBytecodePatch.kt
```diff
@@ -0,0 +1,62 @@
...
+
+        // Inject call for video id
+        VideoIdPatch.injectCall("$INTEGRATIONS_PATCH_UTILS_DESCRIPTOR->setVideoId(Ljava/lang/String;)V")
```
oSumAtrIX, line 50: This is not necessary. There is another patch that can be used as a dependency for this one. The integrations of that patch contain a getter for the current video id. The only time `VideoIdPatch` is required is if you need to update something based on a new video or its id. For example, the `return-youtube-dislike` requires it, in order to update the dislike count. This PR adds buttons that only require to use of the mentioned getter for the video id.

oSumAtrIX (changes requested):
Looks good overall! I have addressed some small changes as well as in the depending integrations PR.

Domiiniik:
Can showing/hiding both buttons can be configured so I can for instance hide copy video button but leave copy timestamp button? Vanced did allowed this.

oSumAtrIX:
That is controlled by the settings patch.

src/main/kotlin/app/revanced/patches/youtube/interaction/copyvideourl/bytecode/patch/CopyVideoUrlBytecodePatch.kt
aliernfrog: Integrations of some patches (RYD, remember video quality and downloads) has getter method for `videoId`, but I couldn't find any other patch this patch can depend on.

oSumAtrIX: What I meant is that the `VideoInformation` patch should include a getter for the videoId. The `downloads` and the patch from this PR would then make use of it. For that the field `videoId` from the `downloads` patch should be moved to
 `app.revanced.integrations.patches.VideoInformation`. Make sure the respective changes are also reflected in the `VideoInformation` patch (such as adding `VideoIdPatch.injectCall` to it and making the `downloads` and your patch dependant on the `VideoInformation` patch).

aliernfrog:
Made following changes:
- Removed unneeded `with` function
- Added video id to VideoInformation patch and used that in copy video url and downloads patch
- Some changes from integrations PR
- Updated icons
IMG_20221231_163653

Indranil012:
>       IMG_20221231_163653

Icons are looking bigger and bolder than download and full screen icon

src/main/kotlin/app/revanced/patches/youtube/interaction/copyvideourl/bytecode/patch/CopyVideoUrlBytecodePatch.kt
```diff
@@ -0,0 +1,55 @@
...
+            "$INTEGRATIONS_PLAYER_PACKAGE/CopyVideoUrlTimestampButton;"
+        )
+    }
+
+    override fun execute(context: BytecodeContext): PatchResult {
+        // Hook video time
+        VideoInformationPatch.videoTimeHook(
+            INTEGRATIONS_PATCH_UTILS_DESCRIPTOR,
+            "setVideoTime"
+        )
```
oSumAtrIX, line 43: Is there a specific reason to do that? Can the integrations not call `VideoInformationPatch.getVideoTime` in the integrations?

oSumAtrIX (changes requested):

oSumAtrIX:
Thanks!

Issue #1133, opened 2022-11-27
bug: ReturnYouTubeDislike does not always show dislikes


LisoUseInAIKyrios:
### Bug description

If you navigate to a previously viewed video (in the watch history) or use forward/back buttons in the player, then the dislike count is not shown.  It can also happen if you switch to full screen and then exit full screen.

It appears YouTube is reusing a previously created UI layout and it's manually updating the number of likes without recreating the UI components.  Thus, the the hook in `ReturnYouTubeDislikePatch#onComponentCreated` cannot handle this since no new UI components are created.

### Steps to reproduce

While watching a video:

1. press the video forward button in the video UI overlay to load the next video.
2. press video back button in the video UI.

The dislike count is missing.

### Temporary fix

A temporary fix, is to scroll down the comments section (so the dislike count is off screen), and then scroll back up again.
Then the dislikes are visible again.

### Solution

The permanent fix, is the patch injection points need some adjustments.

Instead of intercepting the creation of the SpannableText, the patch needs to intercept the creation of the like/dislike buttons or the UI container holding the buttons and text.  Then ReVanced could add change listeners and override any text changes YouTube makes.  This also would likely allow fixing https://github.com/revanced/revanced-patches/issues/1140

I can do any fixes in ReVanced-Integrations, but I am unfamiliar figuring out the patch changes needed.

LisousEinaiKyrios:
The existing 'hide buttons patch' can hide the like and dislike buttons using a BlockRule.  So it's already known how to find the like/dislike buttons:
```
    public ButtonsPatch() {
        BlockRule like = new BlockRule(SettingsEnum.HIDE_LIKE_BUTTON, "|like_button");
        dislikeRule = new BlockRule(SettingsEnum.HIDE_DISLIKE_BUTTON, "dislike_button");
```

What's needed is the same logic to intercept these UI objects, but instead of just hiding the UI it needs to pass the UI object to some other ReVanced code.

Basically, a new class similar to BlockRule, except it will not block and instead pass the UI object to some other ReVanced method.


This might allow replace the injection point of `ReturnYouTubeDislikePatch#onComponentCreated(Object conversionContext, AtomicReference<Object> textRef)` with a declaration of this new class.

If something like this was available, it would be fairly easy to fix the dislike not updating issues.  It would also likely fix the [dislike layout issue](https://github.com/revanced/revanced-patches/issues/1140), since Integrations could modify the layout of the segmented like/dislike view.

[...]

PR #1448, opened 2023-01-07
fix for ReturnYouTubeDislikes not always shown

LisoUseInAIKyrios:
Fix for issue https://github.com/revanced/revanced-patches/issues/1133

This fixes the issue of dislikes not shown when switching out of full screen, and when using the next/previous video overlay buttons.


oSumAtrIX:
Can you add a comment on how it fixes it/Why it is necessary at this index?

LisoUseInAIKyrios:
I added some code comments to why index zero is needed.

oSumAtrIX (approved):

PR #1449, opened 2023-01-07
feat(YouTube): Force old UI layout patch

LisoUseInAIKyrios:
RE: revanced/revanced-integrations/pull/282

Imported code of experimental patch.

Seems to work ok.

If the patch is switched on and then off again, the old layout may hang around until YouTube decides to give the new UI again (but that is sort of expected).


src/main/kotlin/app/revanced/patches/youtube/layout/forceolduilayout/bytecode/fingerprints/ForceOldUILayoutFingerprint.kt
```diff
@@ -0,0 +1,19 @@
...
+    "L", AccessFlags.PUBLIC or AccessFlags.STATIC, listOf("L"), listOf(
+        Opcode.IGET_OBJECT,
+        Opcode.GOTO,
+        Opcode.CONST_STRING,
+    ),
+
+    // Instead of applying a bytecode patch, it might be possible to use only integration java code and
+    // manually set the desired version string as this keyed value in the SharedPreferences.
+    // But, this bytecode patch is simple and it works.
+    strings = listOf("pref_override_build_version_name")
```
oSumAtrIX, line 18: This might cause unintended behaviour. Technically this patch only allows to force the old UI over abusing this override, but does not convey that it might change other things. Should this patch not be called what it does? While it forces the old UI as a side effect, its doing something else which is `spoof-build-version-name`. The patch description can then explain what side effects it causes such as bring back the old layout.

LisoUseInAIKyrios: Yeah I think that's better.  Although I think 'spoof' or some other wording is better than 'override', because it may confuse the user to think the patch uses an older version of YouTube (when it's really just spoofing what YouTube thinks what version it is)

How should the settings present the disclaimer?  It's a lot to stick something like this into a summary title:
"When switching off, the old layout may persist until user is logged out or app data is cleared"

LisoUseInAIKyrios: Maybe call the patch "SpoofOldAppVersionPatch"?

oSumAtrIX: > Maybe call the patch "SpoofOldAppVersionPatch"?

In that case, it should be called `spoof-app-version`. The patch is spoofing the _app version_ **to** the old version and does not spoof the _old_ app version.

oSumAtrIX: > How should the settings present the disclaimer?

The description can be wordy, though it would break our current efforts of them being short. I don't have another solution yet, for that reason using a wordy description is fine. If possible, also open an issue for this for further reference after merge.

src/main/kotlin/app/revanced/patches/youtube/layout/spoofappversion/bytecode/patch/SpoofAppVersionPatch.kt
```diff
@@ -0,0 +1,64 @@
...
+import app.revanced.patches.youtube.layout.spoofappversion.annotations.SpoofAppVersionCompatibility
+import app.revanced.patches.youtube.layout.spoofappversion.bytecode.fingerprints.SpoofAppVersionFingerprint
+import app.revanced.patches.youtube.misc.integrations.patch.IntegrationsPatch
+import app.revanced.patches.youtube.misc.settings.bytecode.patch.SettingsPatch
+import org.jf.dexlib2.iface.instruction.OneRegisterInstruction
+
+@Patch
+@DependsOn([IntegrationsPatch::class, SettingsPatch::class])
+@Name("spoof-app-version")
+@Description("Makes YouTube think the app is an older version.  This gives the old UI layout, but may cause unknown side effects.")
```
oSumAtrIX, line 25:
```suggestion
@Description("Makes YouTube think the app is an older version. This gives the old UI layout, but may cause unknown side effects.")
```

src/main/kotlin/app/revanced/patches/youtube/layout/spoofappversion/bytecode/patch/SpoofAppVersionPatch.kt
```diff
@@ -0,0 +1,64 @@
...
+        const val INTEGRATIONS_CLASS_DESCRIPTOR = "Lapp/revanced/integrations/patches/SpoofAppVersionPatch"
+    }
+
+    override fun execute(context: BytecodeContext): PatchResult {
+        SettingsPatch.PreferenceScreen.LAYOUT.addPreferences(
+            SwitchPreference(
+                "revanced_spoof_app_version",
+                StringResource("revanced_spoof_app_version_title", "Spoof app version (Show old UI)"),
+                false,
+                StringResource("revanced_spoof_app_version_summary_on", "Version spoofed to 17.30.34.  If switched off, the old UI layout may remain until user logs out or the app data is cleared"),
```
oSumAtrIX, line 43: Through the comment, the old UI is implicated, allowing for the title to be short and clean.

```suggestion
                StringResource("revanced_spoof_app_version_title", "Spoof app version"),
                false,
                StringResource("revanced_spoof_app_version_summary_on", "Version spoofed to 17.30.34. If switched off, the old UI layout may remain until logging out or clearing app data"),
```

src/main/kotlin/app/revanced/patches/youtube/layout/spoofappversion/bytecode/fingerprints/SpoofAppVersionFingerprint.kt
```diff
@@ -0,0 +1,19 @@
...
+object SpoofAppVersionFingerprint : MethodFingerprint(
+    "L", AccessFlags.PUBLIC or AccessFlags.STATIC, listOf("L"), listOf(
+        Opcode.IGET_OBJECT,
+        Opcode.GOTO,
+        Opcode.CONST_STRING,
+    ),
+
+    // Instead of applying a bytecode patch, it might be possible to use only integration java code and
+    // manually set the desired version string as this keyed value in the SharedPreferences.
+    // But, this bytecode patch is simple and it works.
```
oSumAtrIX, line 17:
```suggestion
    // Instead of applying a bytecode patch, it might be possible to only rely on code from the integrations and
    // manually set the desired version string as this keyed value in the SharedPreferences.
    // But, this bytecode patch is simple and it works.
```

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/youtube/layout/spoofappversion/bytecode/patch/SpoofAppVersionPatch.kt
```diff
@@ -0,0 +1,64 @@
...
+import app.revanced.patches.youtube.layout.spoofappversion.annotations.SpoofAppVersionCompatibility
+import app.revanced.patches.youtube.layout.spoofappversion.bytecode.fingerprints.SpoofAppVersionFingerprint
+import app.revanced.patches.youtube.misc.integrations.patch.IntegrationsPatch
+import app.revanced.patches.youtube.misc.settings.bytecode.patch.SettingsPatch
+import org.jf.dexlib2.iface.instruction.OneRegisterInstruction
+
+@Patch
+@DependsOn([IntegrationsPatch::class, SettingsPatch::class])
+@Name("spoof-app-version")
+@Description("Makes YouTube think the app is an older version. This gives the old UI layout, but may cause unknown side effects.")
```
oSumAtrIX, line 25:
```suggestion
@Description("Tricks YouTube into thinking, you are running an older version of the app. An (un)intended side effect is the former design of YouTube.")
```

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/youtube/layout/spoofappversion/bytecode/patch/SpoofAppVersionPatch.kt
LisoUseInAIKyrios: It's not unintended for the patch itself, because the intention is to give the old UI.

oSumAtrIX: This is why it's "(un)intended", not "unintended". Maybe:

```suggestion
@Description("Tricks YouTube into thinking, you are running an older version of the app. A side effect is the former design of YouTube.")
```

LisoUseInAIKyrios: How about:

"Tricks YouTube into thinking, you are running an older version of the app. This restores the old UI layout, but may cause unknown side effects."

The wording 'former design of youtube' may not convey it's the entire UI that changes.

I think the description should have a disclaimer about unknown effects, because it's not conclusively known that it will always work 100% correctly.  Could new features behave differently?  Could the app crash because it's served up old data it doesn't understand?  All are possible.

oSumAtrIX: In that case, it can be said "One of the side-effects..." which allows getting rid of the last sentence module "but may cause unknown side effects"

```suggestion
@Description("Tricks YouTube into thinking, you are running an older version of the app. One of the side effects also includes restoring the old UI.")
```

LisoUseInAIKyrios:
Looks good

oSumAtrIX:
Thanks.

PR #1453, opened 2023-01-08
feat(YouTube-Music/compatibility): update to 5.38.53

FineFindus:
Updates the compatibility of the YouTube Music patches to support `5.28.53`. All patches still work.

The `CodecsLockFingerprint` failed, so I changed `@FuzzyPatternScanMethod(2)` to `@FuzzyPatternScanMethod(3)`, which fixed it.

### DIFF
--- a/src/main/kotlin/app/revanced/patches/music/audio/codecs/fingerprints/CodecsLockFingerprint.kt
+++ b/src/main/kotlin/app/revanced/patches/music/audio/codecs/fingerprints/CodecsLockFingerprint.kt
@@ -7,7 +7,7 @@ import org.jf.dexlib2.AccessFlags
 import org.jf.dexlib2.Opcode
 
 
-@FuzzyPatternScanMethod(2) // FIXME: Test this threshold and find the best value.
+@FuzzyPatternScanMethod(3) // FIXME: Test this threshold and find the best value.
 object CodecsLockFingerprint : MethodFingerprint(
     "L", AccessFlags.PUBLIC or AccessFlags.STATIC, opcodes = listOf(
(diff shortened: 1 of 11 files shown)


oSumAtrIX:
The `@FuzzyPatternScanMethod` annotation is only an assist for fingerprints. If the fuzzy scanner was able to succeed with 1 or more mismatches, the pattern scan result will contain a list of warnings that can be used to fix the fingerprint in order to guarantee, that the fingerprint continues to work properly in the future.

FineFindus:
I updated the `CodecsLockFingerprint` so it works with `@FuzzyPatternScanMethod(2)` again.

src/main/kotlin/app/revanced/patches/music/audio/codecs/fingerprints/CodecsLockFingerprint.kt
```diff
@@ -22,7 +22,7 @@ object CodecsLockFingerprint : MethodFingerprint(
         Opcode.INVOKE_STATIC,
         Opcode.MOVE_RESULT_OBJECT,
         Opcode.INVOKE_INTERFACE,
-        Opcode.INVOKE_DIRECT,
+        Opcode.INVOKE_VIRTUAL,
```
oSumAtrIX, line 25: Can you assure, this fingerprint will work with previous versions mentioned in the compatible versions list as well?

FineFindus: I checked the two latest and the first version, so I'm guessing that they all will work.

PR #1454, opened 2023-01-08
feat(YouTube): ReturnYouTubeDislike compact like button layout

LisoUseInAIKyrios:
RE: https://github.com/revanced/revanced-integrations/pull/284

Added an option for a slightly more narrow like button layout, that removes the left separator and tightens the space between the like/dislike by a few more pixels.

Mostly intended for users of devices with strange system fonts (such as [Xiaomi](https://github.com/revanced/revanced-integrations/issues/276)), or any other device where the left separator does not style well.

compact layout

full size

### DIFF
--- a/src/main/resources/returnyoutubedislike/host/values/strings.xml
+++ b/src/main/resources/returnyoutubedislike/host/values/strings.xml
@@ -17,8 +17,8 @@
     <string name="revanced_ryd_dislike_percentage_summary_off">Dislikes shown as number</string>
 
     <string name="revanced_ryd_compact_layout_title">Compact like button</string>
-    <string name="revanced_ryd_compact_layout_summary_on">Likes and Dislikes minimally styled</string>
-    <string name="revanced_ryd_compact_layout_summary_off">Likes and Dislikes fully styled</string>
+    <string name="revanced_ryd_compact_layout_summary_on">Likes and Dislikes styled minimally</string>
+    <string name="revanced_ryd_compact_layout_summary_off">Likes and Dislikes styled fully</string>
 
     <string name="revanced_ryd_attribution_title">ReturnYouTubeDislike.com</string>
     <string name="revanced_ryd_attribution_summary">Dislike data is provided by the Return YouTube Dislike API.  Tap here to learn more.</string>
--- a/src/main/resources/returnyoutubedislike/host/values/strings.xml
+++ b/src/main/resources/returnyoutubedislike/host/values/strings.xml
@@ -18,7 +18,7 @@
 
     <string name="revanced_ryd_compact_layout_title">Compact like button</string>
     <string name="revanced_ryd_compact_layout_summary_on">Likes and Dislikes styled minimally</string>
-    <string name="revanced_ryd_compact_layout_summary_off">Likes and Dislikes styled fully</string>
+    <string name="revanced_ryd_compact_layout_summary_off">Likes and Dislikes styled symmetrically</string>
 
     <string name="revanced_ryd_attribution_title">ReturnYouTubeDislike.com</string>
     <string name="revanced_ryd_attribution_summary">Dislike data is provided by the Return YouTube Dislike API.  Tap here to learn more.</string>
--- a/src/main/resources/returnyoutubedislike/host/values/strings.xml
+++ b/src/main/resources/returnyoutubedislike/host/values/strings.xml
@@ -17,8 +17,8 @@
     <string name="revanced_ryd_dislike_percentage_summary_off">Dislikes shown as number</string>
 
     <string name="revanced_ryd_compact_layout_title">Compact like button</string>
-    <string name="revanced_ryd_compact_layout_summary_on">Like and Dislike displayed together</string>
-    <string name="revanced_ryd_compact_layout_summary_off">Like and Dislike displayed separately</string>
+    <string name="revanced_ryd_compact_layout_summary_on">Likes and Dislikes displayed together</string>
+    <string name="revanced_ryd_compact_layout_summary_off">Likes and Dislikes displayed separately</string>
 
     <string name="revanced_ryd_attribution_title">ReturnYouTubeDislike.com</string>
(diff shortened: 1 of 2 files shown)


oSumAtrIX:
Another addition that could solve the issue with styling for all users is to add options that allow changing the style to the settings.

src/main/resources/returnyoutubedislike/host/values/strings.xml
```diff
@@ -17,6 +16,10 @@
     <string name="revanced_ryd_dislike_percentage_summary_on">Dislikes shown as percentage</string>
     <string name="revanced_ryd_dislike_percentage_summary_off">Dislikes shown as number</string>
 
+    <string name="revanced_ryd_compact_layout_title">Compact like button layout</string>
+    <string name="revanced_ryd_compact_layout_summary_on">Showing narrow like button</string>
+    <string name="revanced_ryd_compact_layout_summary_off">Showing normal like button</string>
```
oSumAtrIX, line 21: These do not convey what this switch does properly. Instead, it should say, that it makes it compact by getting rid of the left separator.

oSumAtrIX (changes requested):

LisoUseInAIKyrios:
I think letting users make adjustments to the style layout is good.  Would need to figure out how to present that to the user in a way that's not confusing or too clunky.

src/main/resources/returnyoutubedislike/host/values/strings.xml
LisoUseInAIKyrios: It does describe what happens: the likes button becomes more narrow.  

The user may not realize what the left separator is.
They'll figure it out when they try turning it on and off.

LisoUseInAIKyrios: maybe something like:

"Showing minimal like button layout"
"Showing sparse like button layout"

But I feel that 'narrow' describes what happens better.

Maybe some other suggestions?

oSumAtrIX:
I'd go with a sub-preference screen with input boxes for inputting offsets and absolute values, like the `custom-buffer` patch.

src/main/resources/returnyoutubedislike/host/values/strings.xml
oSumAtrIX: My following suggestion should be formulated/worded better, but something like this maybe:
```suggestion
    <string name="revanced_ryd_compact_layout_title">Compact like button text</string>
    <string name="revanced_ryd_compact_layout_summary_on">Only showing like and dislike separately</string>
    <string name="revanced_ryd_compact_layout_summary_off">Fully formatting like and dislike symmetrically</string>
```

LisoUseInAIKyrios: maybe:
Compact like button
Like and Dislike displayed together
Like and Dislike displayed separately

oSumAtrIX: > Like and Dislike displayed together

But they are not displayed together at all are they?

LisoUseInAIKyrios: I think this describes it well, and it reads ok too.

separately

together

LisoUseInAIKyrios: Although I'll change it to so "Like**s** and Dislike**s**"

LisoUseInAIKyrios: > > Like and Dislike displayed together
> 
> But they are not displayed together at all are they?

yeah I guess they are not.  It's "more together" but they still have the • separator between them.

LisoUseInAIKyrios: Maybe:
Likes and Dislikes minimally styled
Likes and Dislikes fully styled

LisoUseInAIKyrios: fully styled


minimally styled

oSumAtrIX: Tough. Maybe 

Likes/ Dislikes buttons styled symmetrically
Likes/ Dislikes button occupy a minimum amount of width

LisoUseInAIKyrios: "occupy a minimum amount of width" is a bit wordy and might be confusing.

How about:

"Likes and Dislikes styled minimally"
"Likes and Dislikes styled symmetrically"

oSumAtrIX: Following on that suggestion, maybe this way:

"Likes and Dislikes styled for minimal width"
"Likes and Dislikes styled for best appearance"

LisoUseInAIKyrios: yes, I like that best

LisoUseInAIKyrios: best appearance

minimal width


I think this looks best

oSumAtrIX:
Then the font of the like button span will look different from the system font.

LisoUseInAIKyrios:
@YT-Advanced 
The issue is not so much the type of font, but instead the system default font size.

Most android phones use the default Android 'Roboto' font.  The exceptions to this include Google Pixel and a lot of Chinese manufacturers (Xiaomi, Oppo, OnePlus, etc).

Using a Google Pixel phone, I tried setting the entire like/dislike span to Roboto font, and then used the same segmented layout adjustments from Samsung (which also uses Roboto).

This did not work well.

Pixel phone:
Roboto font mixed with Google Sans font

Samsung phone:
Samsung phone

For the Pixel phone, the font of the 'likes' button visually looks the same as the other buttons.  Roboto is not very different from Google Sans, but with a different system font it might be noticeable.

Using the same font type does not make the layout adjustments universal.  Trying the same layout adjustments, these two devices had a different visual appearance, and it ends up about the same as a different no-name device running the currently released code (which does not change the font).

So unfortunately this won't help.

oSumAtrIX (approved):

PR #1456, opened 2023-01-09
ci: set node version and update dependencies

PalmDevs:
*Forgive me for the branch name.*

**Recently `semantic-release` has updated to v20 with minimal breaking changes for end users.** I figured it would be the time to update all dependencies and fix some workflow issues I've found.

Changes include:
 - Setting Node version when setting up Node.js, now it targets v18 instead of latest, which is what `semantic-release` supports.
 - Using `npm exec` instead of `npx`. `npx` is not a way to execute local commands in a project's environment but rather to download and test packages and it may be unreliable. If anything fails, `npm exec` will always be the one to fail first. `npx` may still prompt you to download new packages instead of stopping if a package cannot be found.
 - Updating `semantic-release` dependencies to the latest versions. Minimal breaking change is present. They've switched to ESM-only.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -44,4 +44,4 @@ jobs:
       - name: Release
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: npx semantic-release
+        run: npm exec semantic-release
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,7 +27,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@v3
         with:
-          node-version: 'latest'
+          node-version: '18'
           cache: 'npm'
       - name: Setup Android SDK
         uses: amyu/setup-android@v2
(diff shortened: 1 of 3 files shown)


PalmDevs:
@oSumAtrIX This needs to be merged to `main`. If you merge it to `dev`, it may fail later.

PalmDevs:
Do not merge this yet, it still does not function correctly. I will fix it in a few hours.

oSumAtrIX:
Why should this be merged to `main` instead of `dev`?

PalmDevs:
> Why should this be merged to `main` instead of `dev`?

So we can immediately test the workflow right after. Doesn't GitHub require you to push the updated workflow to main to be able to test? If anything fails, we can immediately make a fix instead of waiting to merge `dev` which includes new changes as well.

she11sh0cked:
> > Why should this be merged to `main` instead of `dev`?
> 
> So we can immediately test the workflow right after. Doesn't GitHub require you to push the updated workflow to main to be able to test? If anything fails, we can immediately make a fix instead of waiting to merge `dev` which includes new changes as well.

The workflow gets triggered on pushes and pull request changes to `main` _and_ `dev`. Should be fine then, right?

PalmDevs:
@she11sh0cked And it manages releases too. If it fails while in `dev`, we wouldn't know until it gets merged to `main`.

PR #1463, opened 2023-01-09
feat(nova-launcher): unlock-prime patch

illerokcob:
This patch unlocks the full version of Nova Launcher.

### DIFF
--- a/src/main/kotlin/app/revanced/patches/nova/prime/patch/UnlockPrimePatch.kt
+++ b/src/main/kotlin/app/revanced/patches/nova/prime/patch/UnlockPrimePatch.kt
@@ -4,17 +4,19 @@ import app.revanced.patcher.annotation.Description
 import app.revanced.patcher.annotation.Name
 import app.revanced.patcher.annotation.Version
 import app.revanced.patcher.data.BytecodeContext
-import app.revanced.patcher.extensions.addInstruction
 import app.revanced.patcher.patch.BytecodePatch
 import app.revanced.patcher.patch.PatchResult
 import app.revanced.patcher.patch.PatchResultSuccess
 import app.revanced.patcher.patch.annotations.Patch
 import app.revanced.patches.nova.prime.annotations.UnlockPrimeCompatibility
 import app.revanced.patches.nova.prime.fingerprints.UnlockPrimeFingerprint
+import org.jf.dexlib2.Opcode
+import org.jf.dexlib2.builder.instruction.BuilderInstruction11x
+import org.jf.dexlib2.builder.instruction.BuilderInstruction21s
 
 @Patch
 @Name("unlock-prime")
-@Description("Unlock Nova Prime.")
+@Description("Unlocks Nova Prime and all functions of the app.")
 @UnlockPrimeCompatibility
 @Version("0.0.1")
 class UnlockPrimePatch : BytecodePatch(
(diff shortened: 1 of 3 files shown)


src/main/kotlin/app/revanced/patches/nova/prime/patch/UnlockPrimePatch.kt
```diff
@@ -0,0 +1,37 @@
...
+class UnlockPrimePatch : BytecodePatch(
+    listOf(
+        UnlockPrimeFingerprint
+    )
+) {
+    override fun execute(context: BytecodeContext): PatchResult {
+        val method = UnlockPrimeFingerprint.result!!.mutableMethod
+
+        method.addInstruction(
+            8,
```
FineFindus, line 29: This will break, when the instruction gets moved in a future update

oSumAtrIX (changes requested):

illerokcob:
I hope it's more reliable now. Unfortunately, there's always a chance that the whole code gets replaced with one update or messed up. 
Tested, and works on latest beta and latest release as well.

Another solution could be injecting/integrating SharedPreferences Editor in Application's onCreate to always overwrite the app's current state at launch. It's more reliable, for sure, but unnecessary, in my opinion.

src/main/kotlin/app/revanced/patches/nova/prime/patch/UnlockPrimePatch.kt
```diff
@@ -0,0 +1,46 @@
...
+    )
+) {
+
+    //Insert the hex value 0x512 right after the pattern which checks the version of the app (paid or free).
+    //Free version is '0', paid version is '0x512'.
+
+    override fun execute(context: BytecodeContext): PatchResult {
+        val result = UnlockPrimeFingerprint.result!!
+        val methodImplementation = result.mutableMethod.implementation
+        val startIndex = result.scanResult.patternScanResult?.startIndex!!
```
oSumAtrIX, line 34: Simply use endIndex here since that's what you apparently need.

src/main/kotlin/app/revanced/patches/nova/prime/patch/UnlockPrimePatch.kt
```diff
@@ -0,0 +1,46 @@
...
+
+        val replaceIndex = startIndex + 5
+
+        methodImplementation!!.addInstruction(
+            replaceIndex,
+            BuilderInstruction21s(
+                Opcode.CONST_16, (methodImplementation.instructions[replaceIndex - 1] as BuilderInstruction11x).registerA, 1298)    // 0x512
```
oSumAtrIX, line 41: Please use a named const val for this magic value.and give it a good name for future reference.

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/nova/prime/patch/UnlockPrimePatch.kt
illerokcob: It seems that if I use any value which is not "0" the app is switches to the full version. (Found it because I mixed up the hex value with the decimal one). I started using this magic value because the unlocker app uses the same integer as well. After some digging it seems the app is checking only for a non-zero value and the number "512" plays a role only in some weird protection to reset preferences when this value changes (I may have triggered this already during testing). Anyway, I would stick with the original value and include a comment about this in the code.

oSumAtrIX (approved):
Please review if the refactor works properly and we can get it merged.

src/main/kotlin/app/revanced/patches/nova/prime/patch/UnlockPrimePatch.kt
```diff
@@ -0,0 +1,50 @@
...
+) {
+    private companion object {
+        // Any value except 0 unlocks prime, but 512 is needed for a protection mechanism
+        // which would reset the preferences if the value on disk had changed after a restart.
+        const val PRIME_STATUS: Int = 512
+    }
+
+    override fun execute(context: BytecodeContext): PatchResult {
+        UnlockPrimeFingerprint.result?.apply {
+            val insertIndex = scanResult.patternScanResult?.endIndex!! + 1
```
oSumAtrIX, line 35:
```suggestion
            val insertIndex = scanResult.patternScanResult!!.endIndex + 1
```

illerokcob (commented):
Works properly with all versions which I've tested before.

oSumAtrIX (approved):

oSumAtrIX:
Thanks.

Issue #1187, opened 2022-12-03
bug: YouTube - hide info cards patch always hides infocards (even if hideinfo cards is not enabled in ReVanced settings)


LisoUseInAIKyrios:
### Type

Cosmetic

### Bug description

If using the hideinfocards patch, then infocards are always hidden -- even if "hide info-cards" is turned off in ReVanced settings.

### Steps to reproduce

- patch YouTube with default patches (which includes "hide-info-cards" patch)
- Launch ReVanced YouTube, navigate to ReVanced settings, and ensure "Hide info-cards" is switched off
-  watch any video with infocards, such as [https://www.youtube.com/watch?v=qiVqicC8sIA](https://github.com/revanced/revanced-patches/issues/url))
- an infocard should be seen at 5s, but no infocard is shown.

Matty003:
#623

oSumAtrIX:
These are closed, the issue is up to date.

LisoUseInAIKyrios:
Until the issue is fixed, maybe change the patch so it is not included in patching by default?

At least that way the user has to manually pick the patch for inclusion in ReVanced Manager.  

Because currently, users get the patch by default and then after installation there is no way of turning it off.

oSumAtrIX:
Or just fix the invalid switch setting name. Probably a typo. It's the same effort as makong it exclude by default.

PR #1464, opened 2023-01-10
fix(youtube/hide-info-cards) Fix for info cards always hiding, even if setting is turned off

LisoUseInAIKyrios:
Fix for issue https://github.com/revanced/revanced-patches/issues/1187

Changed the patch so it does not replace the existing method call, as it appears the original method makes the card visible to begin with.

Verified this fix using both normal and incognito mode.


oSumAtrIX (approved):

oSumAtrIX:
Thanks

PR #1483, opened 2023-01-13
fix(tiktok/fingerprint): update 'feedfilter' and 'settings' fingerprints to new version

d4rkk3y:
update outdated fingerprints.

### DIFF
--- a/src/main/kotlin/app/revanced/patches/tiktok/feedfilter/annotations/FeedFilterCompatibility.kt
+++ b/src/main/kotlin/app/revanced/patches/tiktok/feedfilter/annotations/FeedFilterCompatibility.kt
@@ -5,8 +5,8 @@ import app.revanced.patcher.annotation.Package
 
 @Compatibility(
     [
-        Package("com.ss.android.ugc.trill"),
-        Package("com.zhiliaoapp.musically")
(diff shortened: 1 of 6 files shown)


FineFindus (commented):
Do the new fingerprints still work with the older app versions?

d4rkk3y:
> Do the new fingerprints still work with the older app versions?

No. It doesn't. Is it better to add version to Compatibility annotations?

oSumAtrIX:
If you are unsure if the patch will work for future versions, add the annotation, if you think it will remain to work, don't.

d4rkk3y:
sorry I create PR to wrong branch.

oSumAtrIX:
You can change the target branch on GitHub by clicking on `edit`.

oSumAtrIX (commented):
Are you sure the fingerprints will work with future versions?

d4rkk3y:
> Are you sure the fingerprints will work with future versions?

No, I am not. Because recently I notice that the settings patch's fingerprint change. For example, the last time you update settings fingerprint but now it changed. So I add version to patches depend on settings patch too.

oSumAtrIX (approved):

TheJeterLP:
This does not work. See #1487

PR #1501, opened 2023-01-15
feat: package name patchoption in microg patch

Aunali321:


src/main/kotlin/app/revanced/patches/youtube/misc/microg/patch/bytecode/MicroGBytecodePatch.kt
```diff
@@ -72,4 +70,15 @@ class MicroGBytecodePatch : BytecodePatch(
...
         return PatchResultSuccess()
     }
+
+    companion object : OptionsContainer() {
+        var packageName: String? by option(
+            PatchOption.StringOption(
+                key = "packageName",
+                default = REVANCED_PACKAGE_NAME,
+                title = "Package name",
+                description = "The package name of MicroG compatible app."
```
oSumAtrIX, line 80:
```suggestion
                description = "The name of the package."
```

src/main/kotlin/app/revanced/patches/music/misc/microg/patch/bytecode/MusicMicroGBytecodePatch.kt
```diff
@@ -46,7 +47,7 @@ class MusicMicroGBytecodePatch : BytecodePatch(
             arrayOf(
                 MicroGBytecodeHelper.packageNameTransform(
                     Constants.PACKAGE_NAME,
-                    Constants.REVANCED_PACKAGE_NAME
+                    packageName!!
```
oSumAtrIX, line 50: This will not work. This patch uses the options from another unrelated patch.

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/youtube/misc/microg/patch/bytecode/MicroGBytecodePatch.kt
```diff
@@ -72,4 +70,15 @@ class MicroGBytecodePatch : BytecodePatch(
 
         return PatchResultSuccess()
     }
+
+    companion object : OptionsContainer() {
+        var packageName: String? by option(
```
oSumAtrIX, line 75: This option should likely exist for the equivalent patch for YouTube Music. Stronger abstraction is required. Both patches have to be turned into an abstract which then is inherited from both patches. They can then override the execute method each. The abstract patch class can have this companion object.

oSumAtrIX (changes requested):

PR #1511, opened 2023-01-17
feat(YouTube) show toast on ReVanced error

LisoUseInAIKyrios:
Patch changes for https://github.com/revanced/revanced-integrations/pull/293


src/main/kotlin/app/revanced/patches/youtube/misc/debugging/patch/DebuggingPatch.kt
```diff
@@ -38,11 +38,21 @@ class DebuggingPatch : ResourcePatch {
                         "revanced_debug_stacktrace_enabled",
                         StringResource(
                             "revanced_debug_stacktrace_enabled_title",
-                            "Print stack traces"
+                            "Debug log stack trace"
```
oSumAtrIX, line 41: Usually, we go with [Verb] [Noun] such as "Print stack traces", are you sure this is better?

oSumAtrIX (approved):

src/main/kotlin/app/revanced/patches/youtube/misc/debugging/patch/DebuggingPatch.kt
LisoUseInAIKyrios: Maybe "debug logging stack trace"?

oSumAtrIX: If you want to indicate, that it applies ro logging, you can use "Log stack traces".
 out of the motivation that "Debuglogging" or "Debuglog" are not verbs.

oSumAtrIX (approved):

src/main/kotlin/app/revanced/patches/youtube/layout/sponsorblock/bytecode/patch/SponsorBlockBytecodePatch.kt
```diff
@@ -263,7 +263,7 @@ class SponsorBlockBytecodePatch : BytecodePatch(
                         "replaceMeWithsetSponsorBarRect" ->
                             method.replaceStringInstruction(index, it, rectangleFieldName)
 
-                        "replaceMeWithsetMillisecondMethod" ->
+                        "replaceMeWithsetMillisecondMethod" -> // Edit: Where in integrations is this method used?  This might be dead code
```
oSumAtrIX, line 266: It seems like both are dead and can be removed.

PR #1518, opened 2023-01-18
feat(finanzonline): Add root and bootloader detection removal patch

1fexd:
Removes the root and bootloader check from [FinanzOnline [+]](https://play.google.com/store/apps/details?id=at.gv.bmf.bmf2go)

### DIFF
--- a/src/main/kotlin/app/revanced/patches/finanzonline/detection/bootloader/patch/BootloaderDetectionPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/finanzonline/detection/bootloader/patch/BootloaderDetectionPatch.kt
@@ -30,7 +30,7 @@ class BootloaderDetectionPatch : BytecodePatch(
                 """
                         const/4 v0, 0x1
                         return v0
-                      """
+                """
             ) ?: return fingerprint.toErrorResult()
         }
         return PatchResultSuccess()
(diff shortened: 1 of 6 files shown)


src/main/kotlin/app/revanced/patches/finanzonline/detection/bootloader/fingerprints/BootloaderDetection2Fingerprint.kt
```diff
@@ -0,0 +1,13 @@
+package app.revanced.patches.finanzonline.detection.bootloader.fingerprints
+
+import app.revanced.patcher.fingerprint.method.impl.MethodFingerprint
+import org.jf.dexlib2.AccessFlags
+
+object BootloaderDetection2Fingerprint : MethodFingerprint(
```
FineFindus, line 6: That not a good name. Maybe rename it to `BootStateFingerprint`?

src/main/kotlin/app/revanced/patches/finanzonline/detection/bootloader/patch/BootloaderDetectionPatch.kt
```diff
@@ -0,0 +1,37 @@
...
+import app.revanced.patcher.patch.BytecodePatch
+import app.revanced.patcher.patch.PatchResult
+import app.revanced.patcher.patch.PatchResultSuccess
+import app.revanced.patcher.patch.annotations.Patch
+import app.revanced.patches.finanzonline.detection.bootloader.fingerprints.BootloaderDetection2Fingerprint
+import app.revanced.patches.finanzonline.detection.bootloader.fingerprints.BootloaderDetectionFingerprint
+
+
+@Patch
+@Name("remove-unlocked-bootloader-detection")
```
FineFindus, line 19: Kind of verbose name. Could be simply shortened to `remove-bootloader-detection` or `bootloader-detection`.

oSumAtrIX:
Why is this patch relevant? Does the app prevent being used if the device is rooted?

1fexd:
> 

Ye

> Why is this patch relevant? Does the app prevent being used if the device is rooted?

Yeah it does

PR #1527, opened 2023-01-20
feat: spoof-wifi-connection patch

Linus789:
Use case: Some apps don’t realize they have an internet connection with Reverse Tethering or Ethernet. This patch makes an app believe, they always have Wifi (on a best effort basis; not really always, but when you have an actual connection it should work for most cases).

Tested with [GeometricWeather](https://github.com/WangDaYeeeeee/GeometricWeather) on Android 13:

- no patch: no connection with [gnirehtet Reverse Tethering](https://github.com/Genymobile/gnirehtet)
- with patch: connection works with gnirehtet Reverse Tethering

Of course it would be better to create a pull request for GeometricWeather to fix this behaviour, but this won’t be possible for closed source apps, so this patch still might prove useful.

Related integrations PR: [revanced-integrations#297](https://github.com/revanced/revanced-integrations/pull/297)

### DIFF
--- a/src/main/kotlin/app/revanced/patches/all/connectivity/wifi/spoof/patch/SpoofWifiPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/all/connectivity/wifi/spoof/patch/SpoofWifiPatch.kt
@@ -6,10 +6,6 @@ import app.revanced.patcher.annotation.Version
 import app.revanced.patcher.patch.annotations.Patch
 import app.revanced.patcher.util.proxy.mutableTypes.MutableMethod
 import app.revanced.util.patch.*
-import app.revanced.util.patch.AbstractTransformInstructionsPatch
-import app.revanced.util.patch.IMethodCall
-import app.revanced.util.patch.Instruction35cInfo
-import app.revanced.util.patch.filterMapInstruction35c
 import org.jf.dexlib2.iface.ClassDef
 import org.jf.dexlib2.iface.Method
 import org.jf.dexlib2.iface.instruction.Instruction
@@ -22,8 +18,8 @@ import java.util.*
 internal class SpoofWifiPatch : AbstractTransformInstructionsPatch<Instruction35cInfo>() {
 
     private companion object {
-        const val INTEGRATIONS_CLASS_DESCRIPTOR_START = "Lapp/revanced/all/connectivity/wifi/spoof/SpoofWifiPatch"
-        const val INTEGRATIONS_CLASS_DESCRIPTOR = "${INTEGRATIONS_CLASS_DESCRIPTOR_START};"
+        const val INTEGRATIONS_CLASS_DESCRIPTOR_PREFIX = "Lapp/revanced/all/connectivity/wifi/spoof/SpoofWifiPatch"
+        const val INTEGRATIONS_CLASS_DESCRIPTOR = "${INTEGRATIONS_CLASS_DESCRIPTOR_PREFIX};"
     }
 
     // Information about method calls we want to replace
@@ -196,7 +192,12 @@ internal class SpoofWifiPatch : AbstractTransformInstructionsPatch<Instruction35
         method: Method,
         instruction: Instruction,
         instructionIndex: Int
-    ) = filterMapInstruction35c<MethodCall>(INTEGRATIONS_CLASS_DESCRIPTOR_START, classDef, instruction, instructionIndex)
+    ) = filterMapInstruction35c<MethodCall>(
+        INTEGRATIONS_CLASS_DESCRIPTOR_PREFIX,
+        classDef,
+        instruction,
+        instructionIndex
+    )
 
     override fun transform(mutableMethod: MutableMethod, entry: Instruction35cInfo) {
         val (methodType, instruction, instructionIndex) = entry
--- a/src/main/kotlin/app/revanced/patches/all/screenshot/removerestriction/patch/RemoveScreenshotRestrictionPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/all/screenshot/removerestriction/patch/RemoveScreenshotRestrictionPatch.kt
@@ -6,10 +6,6 @@ import app.revanced.patcher.annotation.Version
 import app.revanced.patcher.patch.annotations.Patch
 import app.revanced.patcher.util.proxy.mutableTypes.MutableMethod
 import app.revanced.util.patch.*
-import app.revanced.util.patch.AbstractTransformInstructionsPatch
-import app.revanced.util.patch.IMethodCall
-import app.revanced.util.patch.Instruction35cInfo
-import app.revanced.util.patch.filterMapInstruction35c
 import org.jf.dexlib2.iface.ClassDef
 import org.jf.dexlib2.iface.Method
 import org.jf.dexlib2.iface.instruction.Instruction
@@ -22,8 +18,9 @@ import java.util.*
 internal class RemoveScreenshotRestrictionPatch : AbstractTransformInstructionsPatch<Instruction35cInfo>() {
 
     private companion object {
-        const val INTEGRATIONS_CLASS_DESCRIPTOR_START = "Lapp/revanced/all/screenshot/removerestriction/RemoveScreenshotRestrictionPatch"
-        const val INTEGRATIONS_CLASS_DESCRIPTOR = "${INTEGRATIONS_CLASS_DESCRIPTOR_START};"
+        const val INTEGRATIONS_CLASS_DESCRIPTOR_PREFIX =
+            "Lapp/revanced/all/screenshot/removerestriction/RemoveScreenshotRestrictionPatch"
+        const val INTEGRATIONS_CLASS_DESCRIPTOR = "$INTEGRATIONS_CLASS_DESCRIPTOR_PREFIX;"
     }
 
     // Information about method calls we want to replace
@@ -46,7 +43,12 @@ internal class RemoveScreenshotRestrictionPatch : AbstractTransformInstructionsP
         method: Method,
         instruction: Instruction,
         instructionIndex: Int
-    ) = filterMapInstruction35c<MethodCall>(INTEGRATIONS_CLASS_DESCRIPTOR_START, classDef, instruction, instructionIndex)
+    ) = filterMapInstruction35c<MethodCall>(
+        INTEGRATIONS_CLASS_DESCRIPTOR_PREFIX,
+        classDef,
+        instruction,
+        instructionIndex
+    )
 
     override fun transform(mutableMethod: MutableMethod, entry: Instruction35cInfo) {
         val (methodType, instruction, instructionIndex) = entry
--- a/src/main/kotlin/app/revanced/util/patch/MethodCall.kt
+++ b/src/main/kotlin/app/revanced/util/patch/MethodCall.kt
@@ -29,36 +29,53 @@ internal interface IMethodCall {
      *
      * replacement method: Integrations#setFlags(Window, int, int)
      */
-    fun replaceInvokeVirtualWithIntegrations(integrationsClassDescriptor: String, method: MutableMethod, instruction: Instruction35c, instructionIndex: Int) {
-        val registers = arrayOf(instruction.registerC, instruction.registerD, instruction.registerE, instruction.registerF, instruction.registerG)
+    fun replaceInvokeVirtualWithIntegrations(
+        definingClassDescriptor: String,
+        method: MutableMethod,
+        instruction: Instruction35c,
+        instructionIndex: Int
+    ) {
+        val registers = arrayOf(
+            instruction.registerC,
+            instruction.registerD,
+            instruction.registerE,
+            instruction.registerF,
+            instruction.registerG
+        )
         val argsNum = methodParams.size + 1 // + 1 for instance of definedClassName
         if (argsNum > registers.size) {
             // should never happen, but just to be sure (also for the future) a safety check
-            throw RuntimeException("not enough registers for ${definedClassName}#${methodName}: required $argsNum registers, but only got ${registers.size}")
+            throw RuntimeException(
+                "Not enough registers for ${definedClassName}#${methodName}: " +
+                        "Required $argsNum registers, but only got ${registers.size}."
+            )
         }
+
         val args = registers.take(argsNum).joinToString(separator = ", ") { reg -> "v${reg}" }
-        val replacementMethodDefinition = "${methodName}(${definedClassName}${methodParams.joinToString(separator = "")})${returnType}"
+        val replacementMethodDefinition =
+            "${methodName}(${definedClassName}${methodParams.joinToString(separator = "")})${returnType}"
+
         method.replaceInstruction(
             instructionIndex,
-            "invoke-static { $args }, ${integrationsClassDescriptor}->${replacementMethodDefinition}"
+            "invoke-static { $args }, ${definingClassDescriptor}->${replacementMethodDefinition}"
         )
     }
 }
 
 internal inline fun <reified E> fromMethodReference(methodReference: MethodReference)
-        where E: Enum<E>, E: IMethodCall = enumValues<E>().firstOrNull { search ->
+        where E : Enum<E>, E : IMethodCall = enumValues<E>().firstOrNull { search ->
     search.definedClassName == methodReference.definingClass
             && search.methodName == methodReference.name
             && methodReference.parameterTypes.toTypedArray().contentEquals(search.methodParams)
 }
 
 internal inline fun <reified E> filterMapInstruction35c(
-    integrationsClassDescriptorStart: String,
+    integrationsClassDescriptorPrefix: String,
     classDef: ClassDef,
     instruction: Instruction,
     instructionIndex: Int
-): Instruction35cInfo? where E: Enum<E>, E: IMethodCall {
-    if (classDef.type.startsWith(integrationsClassDescriptorStart)) {
+): Instruction35cInfo? where E : Enum<E>, E : IMethodCall {
+    if (classDef.type.startsWith(integrationsClassDescriptorPrefix)) {
         // avoid infinite recursion
         return null
     }


FineFindus (changes requested):
I think it would be better to rename this to something like `spoof-wifi-connection`.

src/main/kotlin/app/revanced/patches/all/connectivity/fakewifi/patch/FakeWifiPatch.kt
```diff
@@ -0,0 +1,252 @@
...
+            val args = registers.take(argsNum).joinToString(separator = ", ") { reg -> "v${reg}" }
+            val replacementMethodDefinition = "${this.methodName}(${this.definedClassName}${this.methodParams.joinToString(separator = "")})${this.returnType}"
+            method.replaceInstruction(
+                instructionIndex,
+                "invoke-static { $args }, ${INTEGRATIONS_CLASS_DESCRIPTOR}->${replacementMethodDefinition}"
+            )
+        }
+
+        companion object {
+            fun fromMethodReference(methodReference: MethodReference) = values().firstOrNull { search ->
```
oSumAtrIX, line 196: You can use `MethodUtil.methodSignaturesMatch` here.

src/main/kotlin/app/revanced/patches/all/connectivity/fakewifi/patch/FakeWifiPatch.kt
```diff
@@ -0,0 +1,252 @@
...
+            }
+        }
+    }
+
+    override fun execute(context: BytecodeContext): PatchResult {
+        // Find all instructions where one of the methods is called
+        buildMap {
+            context.classes.forEach { classDef ->
+                if (classDef.type == INTEGRATIONS_CLASS_DESCRIPTOR) {
+                    // avoid infinite recursion
```
oSumAtrIX, line 209: This code has a lot of similarity with `RemoveScreenshotRestrictionPatch`. For that reason it is advised to make use of OOP patterns and abstract this patch. As reference `AbstractIntegrationsPatch` can be used.

src/main/kotlin/app/revanced/patches/all/connectivity/fakewifi/patch/FakeWifiPatch.kt
```diff
@@ -0,0 +1,252 @@
+package app.revanced.patches.all.connectivity.fakewifi.patch
```
oSumAtrIX, line 1: This patch should be in the package `app.revanced.patches.all.connectivity.wifi.fake.patch`

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/all/connectivity/fakewifi/patch/FakeWifiPatch.kt
Linus789: Or `app.revanced.patches.all.connectivity.wifi.spoof.patch` as suggested by FineFindus?

oSumAtrIX: Works.

src/main/kotlin/app/revanced/patches/all/connectivity/fakewifi/patch/FakeWifiPatch.kt
Linus789: using
```kotlin
MethodUtil.methodSignaturesMatch(methodReference, object : MethodReference {
    override fun getDefiningClass() = search.definedClassName
    override fun getName() = search.methodName
    override fun getParameterTypes() = search.methodParams.toList()
    override fun getReturnType() = search.returnType
    override fun validateReference() = Unit
    override fun compareTo(other: MethodReference) = throw NotImplementedError()
})
```
crashed my app for some reason.

src/main/kotlin/app/revanced/patches/all/connectivity/fakewifi/patch/FakeWifiPatch.kt
Linus789: is it fine like this?

src/main/kotlin/app/revanced/patches/all/connectivity/fakewifi/patch/FakeWifiPatch.kt
Linus789: done

src/main/kotlin/app/revanced/patches/all/connectivity/fakewifi/patch/FakeWifiPatch.kt
oSumAtrIX: Which exception did you get? In any case this looks too complex, you can use the current implementation.

Linus789: [log.txt](https://github.com/revanced/revanced-patches/files/10473228/log.txt)

src/main/kotlin/app/revanced/patches/all/connectivity/wifi/spoof/patch/SpoofWifiPatch.kt
```diff
@@ -0,0 +1,238 @@
...
+        }
+    }
+
+    override fun filterMap(
+        classDef: ClassDef,
+        method: Method,
+        instruction: Instruction,
+        instructionIndex: Int
+    ): Optional<InstructionInfo> {
+        if (classDef.type == INTEGRATIONS_CLASS_DESCRIPTOR) {
```
Linus789, line 218: small thing i want to edit here

Linus789: done

src/main/kotlin/app/revanced/patches/all/connectivity/wifi/spoof/patch/SpoofWifiPatch.kt
```diff
@@ -0,0 +1,232 @@
...
+            }
+        }
+    }
+
+    override fun filterMap(
+        classDef: ClassDef,
+        method: Method,
+        instruction: Instruction,
+        instructionIndex: Int
+    ): Optional<InstructionInfo> {
```
oSumAtrIX, line 211: Is there a specific reason `Optional` is used instead of nullable?

Linus789: No.

oSumAtrIX: In that case, please use the nullable type `T?`.

Linus789: Done.

src/main/kotlin/app/revanced/patches/all/connectivity/wifi/spoof/patch/SpoofWifiPatch.kt
```diff
@@ -0,0 +1,232 @@
...
+            )
+        }
+
+        companion object {
+            fun fromMethodReference(methodReference: MethodReference) = values().firstOrNull { search ->
+                search.definedClassName == methodReference.definingClass
+                        && search.methodName == methodReference.name
+                        && methodReference.parameterTypes.toTypedArray().contentEquals(search.methodParams)
+            }
+        }
```
oSumAtrIX, line 203: This method is a duplicate of the method from `RemoveScreenshotRestrictionPatch` apart from the parameters. If that is not expected, this should be refactored to a single commonly used method.

src/main/kotlin/app/revanced/patches/all/connectivity/wifi/spoof/patch/SpoofWifiPatch.kt
```diff
@@ -0,0 +1,232 @@
...
+
+        fun replaceInstruction(method: MutableMethod, instruction: Instruction35c, instructionIndex: Int) {
+            val registers = arrayOf(instruction.registerC, instruction.registerD, instruction.registerE, instruction.registerF, instruction.registerG)
+            val argsNum = this.methodParams.size + 1 // + 1 for instance of definedClassName
+            if (argsNum > registers.size) {
+                // should never happen, but just to be sure (also for the future) a safety check
+                throw RuntimeException("not enough registers for ${this.definedClassName}#${this.methodName}: required $argsNum registers, but only got ${registers.size}")
+            }
+            val args = registers.take(argsNum).joinToString(separator = ", ") { reg -> "v${reg}" }
+            val replacementMethodDefinition = "${this.methodName}(${this.definedClassName}${this.methodParams.joinToString(separator = "")})${this.returnType}"
```
oSumAtrIX, line 190: Using `this` adds noise to the code. Removing it turns `${this.methodName}` into `$methodName`.

src/main/kotlin/app/revanced/patches/all/connectivity/wifi/spoof/patch/SpoofWifiPatch.kt
```diff
@@ -0,0 +1,232 @@
...
+@Description("Spoofs an existing Wi-Fi connection.")
+@Version("0.0.1")
+class SpoofWifiPatch : AbstractTransformInstructionsPatch<InstructionInfo>() {
+
+    private companion object {
+        const val INTEGRATIONS_CLASS_DESCRIPTOR = "Lapp/revanced/all/connectivity/wifi/spoof/SpoofWifiPatch;"
+    }
+
+    // Information about method calls we want to replace
+    enum class MethodCall(
```
oSumAtrIX, line 31: Access to this enum can be private.

src/main/kotlin/app/revanced/patches/all/connectivity/wifi/spoof/patch/SpoofWifiPatch.kt
```diff
@@ -0,0 +1,232 @@
...
+        }
+    }
+
+    override fun filterMap(
+        classDef: ClassDef,
+        method: Method,
+        instruction: Instruction,
+        instructionIndex: Int
+    ): InstructionInfo? {
+        if (classDef.type.startsWith(INTEGRATIONS_CLASS_DESCRIPTOR.removeSuffix(";"))) {
```
oSumAtrIX, line 212: Wouldn't it be sufficient to check for equality since we can assume there is only one class that has to be considered to avoid recursion? This would allow getting rid of calling `removeSuffix` for each instruction in all classes and methods.

src/main/kotlin/app/revanced/patches/all/screenshot/removerestriction/patch/RemoveScreenshotRestrictionPatch.kt
```diff
@@ -1,35 +1,37 @@
...
-class RemoveScreenshotRestrictionPatch : BytecodePatch() {
+class RemoveScreenshotRestrictionPatch : AbstractTransformInstructionsPatch<InstructionInfo>() {
 
     private companion object {
         const val INTEGRATIONS_CLASS_DESCRIPTOR = "Lapp/revanced/all/screenshot/removerestriction/RemoveScreenshotRestrictionPatch;"
     }
 
     // Information about method calls we want to replace
-    private enum class MethodCall(
+    enum class MethodCall(
```
oSumAtrIX, line 31: This enum can be internal to the package, since it is now necessary to access its fields.

src/main/kotlin/app/revanced/patches/all/connectivity/wifi/spoof/patch/SpoofWifiPatch.kt
Linus789: Won’t work because of
```kotlin
private typealias InstructionInfo = Triple<SpoofWifiPatch.MethodCall, Instruction35c, Int>
class SpoofWifiPatch : AbstractTransformInstructionsPatch<InstructionInfo>() {
```

src/main/kotlin/app/revanced/patches/all/screenshot/removerestriction/patch/RemoveScreenshotRestrictionPatch.kt
Linus789: Won’t work because of
```kotlin
private typealias InstructionInfo = Triple<RemoveScreenshotRestrictionPatch.MethodCall, Instruction35c, Int>
class RemoveScreenshotRestrictionPatch : AbstractTransformInstructionsPatch<InstructionInfo>() {
```

src/main/kotlin/app/revanced/patches/all/connectivity/wifi/spoof/patch/SpoofWifiPatch.kt
Linus789: I’m not sure what would be a good way to refactor this, therefore I’d leave it as it is right now.

src/main/kotlin/app/revanced/patches/all/screenshot/removerestriction/patch/RemoveScreenshotRestrictionPatch.kt
oSumAtrIX: That's odd. The type alias is private so it should've been possible to make `MethodCall` private.

src/main/kotlin/app/revanced/patches/all/connectivity/wifi/spoof/patch/SpoofWifiPatch.kt
oSumAtrIX: A utility class inside the patches package can be created.

src/main/kotlin/app/revanced/patches/all/connectivity/wifi/spoof/patch/SpoofWifiPatch.kt
Linus789: Yes, it would be sufficient. Although it isn’t stated in the comment, another reason to do this check is to simply avoid messing up the logic of our patch, so that we can still see what the real network info looks like and so on. Currently, there is no such logic in the subclasses, but I’d still prefer to keep it that way as that might change in the future and it keeps the instruction replacement logic consistent for the integrations class and its subclasses. Also, I don’t think the `removeSuffix` call has a huge performance impact.

src/main/kotlin/app/revanced/patches/all/connectivity/wifi/spoof/patch/SpoofWifiPatch.kt
Linus789: Sure, I can do that. But I’ll wait for your other responses first.

src/main/kotlin/app/revanced/patches/all/connectivity/wifi/spoof/patch/SpoofWifiPatch.kt
Linus789: Do you want me to only refactor the check if the methods match, or the entire `fromMethodReference` method?

src/main/kotlin/app/revanced/patches/all/screenshot/removerestriction/patch/RemoveScreenshotRestrictionPatch.kt
Linus789: > private means that the member is visible inside this class only (including all its members).

https://kotlinlang.org/docs/visibility-modifiers.html#class-members

src/main/kotlin/app/revanced/patches/all/connectivity/wifi/spoof/patch/SpoofWifiPatch.kt
oSumAtrIX: Removing it sounds good to me.

src/main/kotlin/app/revanced/patches/all/connectivity/wifi/spoof/patch/SpoofWifiPatch.kt
oSumAtrIX: The entire method can be outsourced into a utility class which both patches can make use of.

Linus789: Could you tell me what the method definition of that utility function should look like? I still have no good idea.

oSumAtrIX: The method takes a MethodReference as a parameter and returns the first match out of your method call enum. Because you can not instantiate enum classes, you will have to rely on inheritencd in order to be able to utilize the utility method from both patches with different MethodCall types.

Linus789: Perhaps something like this?
```kotlin
interface IMethodCall {
    val definedClassName: String
    val methodName: String
    val methodParams: Array<String>
    val returnType: String
}

open class MethodCallResolver<E>(private val clazz: Class<E>) where E: IMethodCall, E: Enum<E> {
    fun fromMethodReference(methodReference: MethodReference) = clazz.enumConstants.firstOrNull { search ->
        search.definedClassName == methodReference.definingClass && search.methodName == methodReference.name
    }
}

enum class MethodCall(
    override val definedClassName: String,
    override val methodName: String,
    override val methodParams: Array<String>,
    override val returnType: String
): IMethodCall {
    SetFlags(
        "Landroid/view/Window;",
        "setFlags",
        arrayOf("I", "I"),
        "V",
    );

    companion object : MethodCallResolver<MethodCall>(MethodCall::class.java)
}
```

oSumAtrIX: You can generically constrain your method to the required types instead of having to instatiate a class MethodCallResolver for each ClassDef and instead of an enum, pass a list of your enums values mapped to the provided interface. This would allow getting rid of making reflective calls (which how it is right now can be improved by simply creating a field for the enum constants in your class MethodCallResolver). If you want I can take on this later this day if it's unclear.

Linus789: Like this?
```kotlin
interface IMethodCall {
    val definedClassName: String
    val methodName: String
    val methodParams: Array<String>
    val returnType: String
}

inline fun <reified E> fromMethodReference(methodReference: MethodReference)
        where E: Enum<E>, E: IMethodCall = enumValues<E>().firstOrNull { search ->
    search.definedClassName == methodReference.definingClass
            && search.methodName == methodReference.name
            && methodReference.parameterTypes.toTypedArray().contentEquals(search.methodParams)
}

enum class MethodCall(
    override val definedClassName: String,
    override val methodName: String,
    override val methodParams: Array<String>,
    override val returnType: String
): IMethodCall {
    SetFlags(
        "Landroid/view/Window;",
        "setFlags",
        arrayOf("I", "I"),
        "V",
    );
}
```

oSumAtrIX: That looks fine now.

src/main/kotlin/app/revanced/patches/all/connectivity/wifi/spoof/patch/SpoofWifiPatch.kt
Linus789: Done.

src/main/kotlin/app/revanced/util/patch/MethodCall.kt
```diff
@@ -0,0 +1,48 @@
+package app.revanced.util.patch
+
+import app.revanced.patcher.extensions.replaceInstruction
+import app.revanced.patcher.util.proxy.mutableTypes.MutableMethod
+import org.jf.dexlib2.iface.instruction.formats.Instruction35c
+import org.jf.dexlib2.iface.reference.MethodReference
+
+interface IMethodCall {
```
oSumAtrIX, line 8: This interface should be package private "internal".

src/main/kotlin/app/revanced/patches/all/connectivity/wifi/spoof/patch/SpoofWifiPatch.kt
```diff
@@ -0,0 +1,210 @@
...
+import app.revanced.util.patch.fromMethodReference
+import org.jf.dexlib2.Opcode
+import org.jf.dexlib2.iface.ClassDef
+import org.jf.dexlib2.iface.Method
+import org.jf.dexlib2.iface.instruction.Instruction
+import org.jf.dexlib2.iface.instruction.formats.Instruction35c
+import org.jf.dexlib2.iface.reference.MethodReference
+import java.util.*
+
+private typealias InstructionInfo = Triple<SpoofWifiPatch.MethodCall, Instruction35c, Int>
```
oSumAtrIX, line 19: This alias is duplicate of another except for the first generic type. Aliases have generic type parameter specification so you can move this alias to a common place as package internal the following way:
```suggestion
internal typealias InstructionInfo<T: IMethodCall> = Triple<T, Instruction35c, Int>
```

oSumAtrIX: Also, a suitable name can be used instead as well, because the triple only considers instructions of type 35c.

src/main/kotlin/app/revanced/patches/all/screenshot/removerestriction/patch/RemoveScreenshotRestrictionPatch.kt
```diff
@@ -1,106 +1,72 @@
...
-                    )
-                }
-            }
+    override fun filterMap(
+        classDef: ClassDef,
+        method: Method,
+        instruction: Instruction,
+        instructionIndex: Int
+    ): InstructionInfo? {
+        if (classDef.type == INTEGRATIONS_CLASS_DESCRIPTOR) {
```
oSumAtrIX, line 52: This line is similar to another above, with the exception that you do not remove the semicolon. If this is unexpected, you should create another constant INTEGRATIONS_DESCRIPTOR_START and append the semicolon for the currently used constant in order to not call the method to remove the suffix from the constant every time this method is called. This way you can use either constant for its own purpose.

src/main/kotlin/app/revanced/patches/all/connectivity/wifi/spoof/patch/SpoofWifiPatch.kt
oSumAtrIX: This would also allow not having to deconstruct the triple to pass it to replaceInvokeVirtualWithIntegrations, as you can then use the typealias as a parameter because it is package private

Linus789: It says `Bounds are not allowed on type alias parameters`. For now I settled with
```kotlin
internal typealias Instruction35cInfo = Triple<IMethodCall, Instruction35c, Int>
```
but I could also use
```kotlin
internal typealias InstructionInfo<T> = Triple<T, Instruction35c, Int>
```

src/main/kotlin/app/revanced/util/patch/MethodCall.kt
Linus789: Done.

Linus789:
Why the force-push?

oSumAtrIX:
The force push was done on the `dev` branch because of an issue with `semantic-release` trying to create the same tag repeatedly. It does not affect this PR.

oSumAtrIX:
Thanks.

PR #1541, opened 2023-01-23
feat(music): update patches compatibility to v5.40.51

FineFindus:
Updates the music patches to compatibility to `v5.40.51`. All patches work, although it feels like more premium ads have been appearing lately than usually, I can exclude it from the pr if you wish.

### DIFF
--- a/src/main/kotlin/app/revanced/patches/music/layout/premium/annotations/HideGetPremiumCompatibility.kt
+++ b/src/main/kotlin/app/revanced/patches/music/layout/premium/annotations/HideGetPremiumCompatibility.kt
@@ -23,8 +23,7 @@ import app.revanced.patcher.annotation.Package
             "5.34.51",
             "5.36.51",
             "5.38.53",
-            "5.39.52",
-            "5.40.51"
+            "5.39.52"
         )
     )]
 )


oSumAtrIX:
Are you refering to the `hide-get-premium` patch?

oSumAtrIX:
You can exclude it from bumping if it does not work for this version.

FineFindus:
I'm not sure if it does or does not work. These ads have always been appearing, even in Vanced Music, and are too random to really test it. Only going of my subjective sense, that it feels like if there have been lately more of them, but if so, that would also affect the older version and would require fixing the patch.

oSumAtrIX:
In that case, it should not be bumped until fixed.

AMSAMSAMSAMS:
WARNING: Skipping hide-get-premium: Incompatible with version 5.40.51. This patch is only compatible with version com.google.android.apps.youtube.music: 5.14.53, 5.16.51, 5.17.51, 5.21.52, 5.22.54, 5.23.50, 5.25.51, 5.25.52, 5.26.52, 5.27.51, 5.28.52, 5.29.52, 5.31.50, 5.34.51, 5.36.51, 5.38.53, 5.39.52

PR #1550, opened 2023-01-28
fix(YouTube) error when switching on/off interaction setting

LisoUseInAIKyrios:
fix for error message when switching on/off download video or copy video url.  The error occurred because the preference sub screen used the same key as an existing settings enum.

```
revanced: ReVancedSettingsFragment: Setting COPY_VIDEO_URL_BUTTON_SHOWN was changed. Preference revanced_copy_video_url: Copy video URL settings Settings related to copy URL buttons in video player

revanced: ReVancedSettingsFragment: Setting cannot be handled: class android.preference.PreferenceScreen Copy video URL settings Settings related to copy URL buttons in video player
```

the fix is to use a unique key for the setting enum.

integration changes: https://github.com/revanced/revanced-integrations/pull/307

### DIFF
--- a/src/main/kotlin/app/revanced/patches/youtube/interaction/downloads/resource/patch/DownloadsResourcePatch.kt
+++ b/src/main/kotlin/app/revanced/patches/youtube/interaction/downloads/resource/patch/DownloadsResourcePatch.kt
@@ -26,11 +26,11 @@ class DownloadsResourcePatch : ResourcePatch {
     override fun execute(context: ResourceContext): PatchResult {
         SettingsPatch.PreferenceScreen.INTERACTIONS.addPreferences(
             PreferenceScreen(
-                "revanced_downloads_settings",
+                "revanced_downloads",
                 StringResource("revanced_downloads_title", "Download settings"),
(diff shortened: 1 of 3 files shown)


src/main/kotlin/app/revanced/patches/youtube/interaction/downloads/resource/patch/DownloadsResourcePatch.kt
```diff
@@ -26,7 +26,7 @@ class DownloadsResourcePatch : ResourcePatch {
     override fun execute(context: ResourceContext): PatchResult {
         SettingsPatch.PreferenceScreen.INTERACTIONS.addPreferences(
             PreferenceScreen(
-                "revanced_downloads",
+                "revanced_downloads_settings",
```
oSumAtrIX, line 29: By our current convention, we use `revanced_downloads` as the preference screen key and `revanced_downloads_enabled` as the preference key for the switch. For consistency, this should be applied here too.

oSumAtrIX:
Thanks.

PR #1556, opened 2023-01-28
feat(spotify-lite): enable on-demand patch

illerokcob:
This patch tries to enable the on-demand feature of the Spotify Lite app (unlimited skips, play any song from any playlist/artist/album). Partially addresses: ReVanced/revanced-patches-template#1462

The patch **DOES NOT** remove any ads according to my testing. Should I indicate this in the patch description? Because people might get confused.

Also, is this correct when searching for a method with no parameters?
`parameters = listOf()`

### DIFF
--- a/src/main/kotlin/app/revanced/patches/spotify/lite/ondemand/fingerprints/OnDemandFingerprint.kt
+++ b/src/main/kotlin/app/revanced/patches/spotify/lite/ondemand/fingerprints/OnDemandFingerprint.kt
@@ -1,8 +1,10 @@
 package app.revanced.patches.spotify.lite.ondemand.fingerprints
 
+import app.revanced.patcher.fingerprint.method.annotation.FuzzyPatternScanMethod
 import app.revanced.patcher.fingerprint.method.impl.MethodFingerprint
 import org.jf.dexlib2.Opcode
 
+@FuzzyPatternScanMethod(2)
 object OnDemandFingerprint : MethodFingerprint(
     "L",
     parameters = listOf(),
--- a/src/main/kotlin/app/revanced/patches/spotify/lite/ondemand/fingerprints/OnDemandFingerprint.kt
+++ b/src/main/kotlin/app/revanced/patches/spotify/lite/ondemand/fingerprints/OnDemandFingerprint.kt
@@ -1,0 +1,23 @@
+package app.revanced.patches.spotify.lite.ondemand.fingerprints
+
+import app.revanced.patcher.fingerprint.method.impl.MethodFingerprint
+import org.jf.dexlib2.Opcode
+
+object OnDemandFingerprint : MethodFingerprint(
+    "L",
+    parameters = listOf(),
+    opcodes = listOf(
+        Opcode.INVOKE_STATIC,
+        Opcode.MOVE_RESULT,
+        Opcode.INVOKE_STATIC,
+        Opcode.MOVE_RESULT_OBJECT,
+        Opcode.IF_EQZ,
+        Opcode.SGET_OBJECT,
+        Opcode.GOTO,
+        Opcode.SGET_OBJECT,
+        Opcode.INVOKE_VIRTUAL,
+        Opcode.MOVE_RESULT,
+        Opcode.IPUT,
+        Opcode.RETURN_OBJECT
+    )
+)
--- a/src/main/kotlin/app/revanced/patches/spotify/lite/ondemand/patch/OnDemandPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/spotify/lite/ondemand/patch/OnDemandPatch.kt
@@ -1,5 +1,6 @@
 package app.revanced.patches.spotify.lite.ondemand.patch
 
+import app.revanced.extensions.toErrorResult
 import app.revanced.patcher.annotation.Description
 import app.revanced.patcher.annotation.Name
 import app.revanced.patcher.annotation.Version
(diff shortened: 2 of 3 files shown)


oSumAtrIX:
To create a fingerprint for a method that has no parameters, you are correctly using `parameters = listOf()
`. 

As for the description, the minority will understand what `On demand` is. I am not sure if the current description is descriptive enough for that reason. It is technically fine as it refers to and explains what the patch does. For that reason, it would also 
 be enough, but I will let you decide if you want to change the description.

src/main/kotlin/app/revanced/patches/spotify/lite/ondemand/fingerprints/OnDemandFingerprint.kt
```diff
@@ -0,0 +1,23 @@
...
+        Opcode.IF_EQZ,
+        Opcode.SGET_OBJECT,
+        Opcode.GOTO,
+        Opcode.SGET_OBJECT,
+        Opcode.INVOKE_VIRTUAL,
+        Opcode.MOVE_RESULT,
+        Opcode.IPUT,
+        Opcode.RETURN_OBJECT
+    )
+)
```
oSumAtrIX, line 23: This could break for future versions. You can compensate for that by using the annotation, which allows scanning the pattern fuzzy. Additionally, it might be beneficial to constrain the compatibility of this patch with targeting versions of Spotify Lite.

src/main/kotlin/app/revanced/patches/spotify/lite/ondemand/patch/OnDemandPatch.kt
```diff
@@ -0,0 +1,33 @@
...
+import app.revanced.patcher.extensions.addInstruction
+import app.revanced.patcher.patch.BytecodePatch
+import app.revanced.patcher.patch.PatchResult
+import app.revanced.patcher.patch.PatchResultSuccess
+import app.revanced.patcher.patch.annotations.Patch
+import app.revanced.patches.spotify.lite.ondemand.annotations.OnDemandCompatibility
+import app.revanced.patches.spotify.lite.ondemand.fingerprints.OnDemandFingerprint
+
+@Patch
+@Name("enable-ondemand")
```
oSumAtrIX, line 16:
```suggestion
@Name("enable-on-demand")
```

src/main/kotlin/app/revanced/patches/spotify/lite/ondemand/patch/OnDemandPatch.kt
```diff
@@ -0,0 +1,33 @@
...
+@Description("Enables On-Demand to play any song from any artist.")
```
oSumAtrIX, line 17:
```suggestion
@Description("Enables listening to songs on-demand, allowing to play any song from playlists, albums or artists without limitations.")
```

src/main/kotlin/app/revanced/patches/spotify/lite/ondemand/patch/OnDemandPatch.kt
```diff
@@ -0,0 +1,33 @@
...
+    listOf(
+        OnDemandFingerprint
+    )
+) {
+    override fun execute(context: BytecodeContext): PatchResult {
+        OnDemandFingerprint.result?.apply {
+            val insertIndex = scanResult.patternScanResult!!.endIndex - 1
+            // Force the UI to behave like with a Premium account
+            mutableMethod.addInstruction(insertIndex,"const/4 v0, 0x2")
+        }
```
oSumAtrIX, line 30: You can return here in case the fingerprint fails to resolve. Make sure to fix imports. 
```suggestion
        } ?: return OnDemandFingerprint.toErrorResult()
```

oSumAtrIX (changes requested):

illerokcob:
I've been testing the patch on an older version of the app and the opcode pattern was the same in the newer versions of the app as well. Now I added this fuzzy pattern scan, but I'm not sure about the threshold. Two means two extra instructions in the pattern?
Adding certain versions only is unnecessary because, as I said, in older versions of the app the pattern matches the current one, so probably it's not frequently modified.

src/main/kotlin/app/revanced/patches/spotify/lite/ondemand/patch/OnDemandPatch.kt
```diff
@@ -0,0 +1,34 @@
...
+import app.revanced.patcher.patch.BytecodePatch
+import app.revanced.patcher.patch.PatchResult
+import app.revanced.patcher.patch.PatchResultSuccess
+import app.revanced.patcher.patch.annotations.Patch
+import app.revanced.patches.spotify.lite.ondemand.annotations.OnDemandCompatibility
+import app.revanced.patches.spotify.lite.ondemand.fingerprints.OnDemandFingerprint
+
+@Patch
+@Name("enable-on-demand")
+@Description("Enables listening to songs on-demand, allowing to play any song from playlists, albums or artists without limitations. Does not disable ads.")
```
oSumAtrIX, line 18:
```suggestion
@Description("Enables listening to songs on-demand, allowing to play any song from playlists, albums or artists without limitations. This does not remove ads.")
```

src/main/kotlin/app/revanced/patches/spotify/lite/ondemand/patch/OnDemandPatch.kt
```diff
@@ -0,0 +1,34 @@
...
+class OnDemandPatch : BytecodePatch(
+    listOf(
+        OnDemandFingerprint
+    )
+) {
+    override fun execute(context: BytecodeContext): PatchResult {
+        OnDemandFingerprint.result?.apply {
+            val insertIndex = scanResult.patternScanResult!!.endIndex - 1
+            // Force the UI to behave like with a Premium account
+            mutableMethod.addInstruction(insertIndex,"const/4 v0, 0x2")
```
oSumAtrIX, line 30:
```suggestion
            // Spoof a premium account
            mutableMethod.addInstruction(insertIndex, "const/4 v0, 0x2")
```

oSumAtrIX (approved):

PR #1576, opened 2023-01-31
fix(youtube/custom-branding): use highest resolution adaptive icons

LisoUseInAIKyrios:
I fixed the tiny app icon, and also increased the size of the adaptive images to the full resolution (108px, 162px, 216px, 324px, 432px)

I did this using the original high resolution image here on [GitHub](https://avatars.githubusercontent.com/u/101597779?s=2000&v=4).  By coincidence, it was the ideal size.

I also updated the old legacy ic_launcher.png and ic_launcher_round.png, although I don't think those are ever used because Android 8 uses the adaptive icons (and YouTube requires Android 8 or higher).

Before:
before

After:
after (whose the tiny one now?)

PR #1580, opened 2023-02-01
fix(youtube/custom-branding): use correct scaling

LisoUseInAIKyrios:
I recreated using 'revanced-logo-shape.svg' for the foreground to have a smoother alpha channel,
and using the same logo to circle ratio as the original SVG, and added the gradient circle background.

Each adaptive icon was made using the original SVG (and not a resized PNG from a single SVG creation).

earlier discussion: https://github.com/revanced/revanced-patches/pull/1576

What was done to create these:

For each icon size:

create foreground PNG:
- export `revanced-logo-shape.svg` to PNG at 2/3'rd the target icon size (288px, for xxxhpi which uses 432px)
- increase PNG canvas size to desired icon size (432px for xxxhdpi)

create background PNG:
- manually edit `revanced-logo.svg`, and comment out the center logo (leaving only the circle and gradient)
- change the circle radius to the viewbox width multiplied by: `((2/3) * 9/10) / 2`  (change the radius to `576`)
- export modified svg at icon size (432px for xxxhdpi)

Repeat above for each icon size.


oSumAtrIX (approved):

oSumAtrIX:
It seems only the foreground was scaled because the background technically clips to the correct size, though to be exact, the background should have the correct ratio to the shape. It should match the ratio between the shape and background of the original logo. I believe applying the same scaling factor to the background as you applied to the shape would solve this.

oSumAtrIX:
If we were to add a margin, it would only be useful for Android and not as a general logo, which requires no margin.

oSumAtrIX (approved):

PR #1581, opened 2023-02-01
feat(netguard): remove-broadcasts-restriction

Linus789:
Makes it possible to start/stop [NetGuard](https://github.com/M66B/NetGuard) via automation apps. Example for ADB to start NetGuard:
```
adb shell am broadcast --user 0 -a eu.faircode.netguard.ON eu.faircode.netguard
```

### DIFF
--- a/src/main/kotlin/app/revanced/patches/netguard/broadcasts/removerestriction/resource/patch/RemoveBroadcastsRestrictionPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/netguard/broadcasts/removerestriction/resource/patch/RemoveBroadcastsRestrictionPatch.kt
@@ -1,0 +1,42 @@
+package app.revanced.patches.netguard.broadcasts.removerestriction.resource.patch
+
+import app.revanced.patcher.annotation.Description
+import app.revanced.patcher.annotation.Name
+import app.revanced.patcher.annotation.Version
+import app.revanced.patcher.data.ResourceContext
+import app.revanced.patcher.patch.PatchResult
+import app.revanced.patcher.patch.PatchResultSuccess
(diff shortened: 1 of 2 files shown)


oSumAtrIX:
NetGuard is open source; why would you want to patch it if you can precompile from source?

Linus789:
Because it is a lot easier just to patch it. It is only a single Manifest that is modified, so there is no compelling reason for me to recompile the entire app.

Linus789:
I don’t think that a PR like this would be accepted, because the ability for third-party-apps to stop NetGuard might impose a security risk.

src/main/kotlin/app/revanced/patches/netguard/broadcasts/removerestriction/resource/patch/RemoveBroadcastsRestrictionPatch.kt
```diff
@@ -0,0 +1,42 @@
...
+@Patch(false)
+@Name("remove-broadcasts-restriction")
+@Description("Enables starting/stopping NetGuard via broadcasts.")
+@RemoveBroadcastsRestrictionCompatibility
+@Version("0.0.1")
+class RemoveBroadcastsRestrictionPatch : ResourcePatch {
+    override fun execute(context: ResourceContext): PatchResult {
+        // create an xml editor instance
+        context.xmlEditor["AndroidManifest.xml"].use { dom ->
+            // get the application node
```
oSumAtrIX, line 23:
```suggestion
        context.xmlEditor["AndroidManifest.xml"].use { dom ->
```

oSumAtrIX (approved):

PR #1624, opened 2023-02-13
ci: update dependencies and target node version

PalmDevs:
**Recently `semantic-release` has updated to v20 with some breaking changes for end users.** I figured it would be the time to update all dependencies and fix some workflow issues I've found.

Changes include:
 - Setting Node version when setting up Node.js, now it targets v18 instead of latest, which is what `semantic-release` supports.
 - Using `npm exec` instead of `npx`. `npx` is not a way to execute local commands in a project's environment but rather to download and test packages and it may be unreliable. If anything fails, `npm exec` will always be the one to fail first. `npx` may still prompt you to download new packages instead of stopping if a package cannot be found.
 - Updating `semantic-release` dependencies to the latest versions, no breaking change is present because `semantic-release-backmerge` finally supports the new `semantic-release` version.*

**This was tested on my repository, here's  [the workflow run](https://github.com/PalmDevs/rvp/actions/runs/4159392559/jobs/7195403175).**

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -45,7 +45,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: ./gradlew generateMeta clean --no-daemon
       - name: Setup semantic-release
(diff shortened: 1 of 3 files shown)


oSumAtrIX (approved):

oSumAtrIX:
This change can be populated to other repositories.

PalmDevs:
Yep, will do that as well

PR #1641, opened 2023-02-17
feat(twelve-widgets): Add unlock paid widgets

1fexd:
Unlocks all paid widgets for [Android 12 Widgets](https://play.google.com/store/apps/details?id=com.dci.dev.androidtwelvewidgets)


src/main/kotlin/app/revanced/patches/twelvewidgets/unlock/annotations/DetectionCompatibility.kt
```diff
@@ -0,0 +1,9 @@
+package app.revanced.patches.twelvewidgets.unlock.annotations
+
+import app.revanced.patcher.annotation.Compatibility
+import app.revanced.patcher.annotation.Package
+
+@Compatibility([Package("com.dci.dev.androidtwelvewidgets", arrayOf("1.4.5", "1.4.2"))])
```
oSumAtrIX, line 6: Is the constraint on the version really required? The patch doesn't seem like it'll break anytime soon.

Issue #352, opened 2022-08-18
bug: ReTwitter ads still showing on main timeline on Android 13


midgelfc:
### Type

Other

### Bug description

Installed ReTwitter using revanced builder on Android and installed ReVanced-Twitter-v9.54.1-release.0-ar m64-v8a-cli 2.9.3-all-v2.9.3.jarpatche s_2.38.0-v2.38.0.apk. Upon lunch and login to the app, ads are still showing on the main timeline

### Steps to reproduce

Open Twitter
login
main timeline loads with ads

### Relevant log output

```shell
N/A
```

### Screenshots or videos

Screenshot showing a promoted ad on the main timeline

![image](https://user-images.githubusercontent.com/78561981/185402064-06c4124a-b2e0-47c2-9675-b93a3ddf0510.png)

App details
![image](https://user-images.githubusercontent.com/78561981/185402156-0e1e593c-c4fa-4dd7-b2cd-88b547c83b5e.png)

### Solution

N/A

### Additional context

Phone: Pixel 6 Pro
OS: Android 13
Build Number: TP1A.220624.021

Running the latest public release of Android 13

oSumAtrIX:
Did you clear the cache?

midgelfc:
Yip, cleared cache, also cleared storage have tried fresh install and also tried the previous version. Ads still showing on the main timeline

oSumAtrIX:
Use

Flat:
Same issue for me. Seems to be happening with anything newer than 9.50 twitter.

<details><summary>CLI patch log</summary>
<p>

```
❯ java -jar ~\Documents\revanced\revanced-cli.jar -c -b ~\Documents\revanced\revanced-patches.jar --keystore ~\Documents\revanced\revancedtwitter.keystore -d 19261FDEE006XD -a Downloads\Twitter_v9.56.2-release.0_apkpure.com.apk -o Downloads\Twitter.apk
INFO: Decoding resources
INFO: Reading dex files
WARNING: Skipping general-reddit-ads: Incompatible with com.twitter.android. This patch is only compatible with com.reddit.frontpage
WARNING: Skipping pflotsh-ecmwf-subscription-unlock: Incompatible with com.twitter.android. This patch is only compatible with com.garzotto.pflotsh.ecmwf_a
WARNING: Skipping minimized-playback-music: Incompatible with com.twitter.android. This patch is only compatible with com.google.android.apps.youtube.music
[...]

PR #1646, opened 2023-02-19
fix(twitter/timeline-ads): specify most recent app version that fully works

LisoUseInAIKyrios:
fixes https://github.com/revanced/revanced-patches/issues/352

changed:
- `timeline-ads` to require Twitter version 9.65.3
- `hide-views-stats` changed from default to optional (and added last version that works, `9.71.0`)

### DIFF
--- a/src/main/kotlin/app/revanced/patches/twitter/layout/hideviews/patch/HideViewsResourcePatch.kt
+++ b/src/main/kotlin/app/revanced/patches/twitter/layout/hideviews/patch/HideViewsResourcePatch.kt
@@ -10,7 +10,7 @@ import app.revanced.patcher.patch.annotations.Patch
 import app.revanced.patches.twitter.layout.hideviews.annotations.HideViewsCompatibility
 import org.w3c.dom.Element
 
-@Patch(false)
+@Patch
 @DependsOn([HideViewsBytecodePatch::class])
 @Name("hide-views-stats")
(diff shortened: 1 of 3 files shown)


src/main/kotlin/app/revanced/patches/twitter/layout/hideviews/patch/HideViewsResourcePatch.kt
```diff
@@ -10,7 +10,7 @@ import app.revanced.patcher.patch.annotations.Patch
 import app.revanced.patches.twitter.layout.hideviews.annotations.HideViewsCompatibility
 import org.w3c.dom.Element
 
-@Patch
+@Patch(false)
```
oSumAtrIX, line 13: Why is this not recommended?

oSumAtrIX: The current motivation to make it excluded by default is only supported by the fact that another patch doesn't support that version, but patches shouldn't depend on each other so this motivation isn't backed. When the patch itself is undesired, for example, if it has a side effect or causes unintended behavior then it is recommended to make it not included by default. If the patches are included by default, but the version does not match, the patch will be automatically skipped by the CLI and not included by the ReVanced Manager app.

oSumAtrIX (approved):

PR #1663, opened 2023-02-23
fix: actually use original app name

PalmDevs:
Makes the `music-microg-support` use the original app name instead of appending ReVanced. MicroG support patch shouldn't be changing the app name in the first place.

### DIFF
--- a/src/main/kotlin/app/revanced/patches/music/misc/microg/shared/Constants.kt
+++ b/src/main/kotlin/app/revanced/patches/music/misc/microg/shared/Constants.kt
@@ -1,7 +1,7 @@
 package app.revanced.patches.music.misc.microg.shared
 
 object Constants {
-    internal const val REVANCED_MUSIC_APP_NAME = "YT Music ReVanced"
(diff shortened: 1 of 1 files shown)


oSumAtrIX:
The MicroG patch changes the package name, respectively the name of the app should be changed as well, otherwise the app name will conflict with the original packages app name. For that reason the respective patch for branding has to be abstracted. The abstraction should allow for patches to extend on it. These patches can then be depended on, such as by the MicroG patch.

Ehab-d:
@oSumAtrIX If you are right, why would a YouTube app build with the original display name if a custom-branding patch is included? Please add custom-branding patch for YT Music. The current display name of the app is taking two lines on my launcher. This looks so bad,

oSumAtrIX:
@Ehab-d The custom branding patch only changes the icon iirc, not the package name.

Ushie:
It changes the Icon and App name

oSumAtrIX:
Right, that is the case for YouTube, but not YouTube Music.

PR #1672, opened 2023-02-24
fix(instagram/hide-timeline-ads): fix compatibility with newer versions

falafelp:
Tested for version 271.1.0.21.84


falafelp:
but app crashes when merging integrations

oSumAtrIX:
The patch does not make use of the integrations, so this can not be the case.

sharunkumar:
> Tested for version 271.1.0.21.84

what's the latest version with which this works?

PR #1695, opened 2023-03-02
feat(Instagram): Add fix integrations patch

falafelp:
should fix https://github.com/revanced/revanced-patches/pull/1672#issuecomment-1444265925
and makes it possible to develop patches that use integrations for instagram


Ushie:
This PR isn't necessary, it's incorrect to do this as the issue lies within the Patcher/Frontends (such as CLI/Manager), not in the patches.

Ushie:
The issue is already resolved in Dev branches of the responsible repositories, only the ReVanced Manager remains.

oSumAtrIX:
Well, merging shouldn't cause an issue to begin with, this PR would be useful in the case integrations are required in the future

oSumAtrIX:
How does it work/ Why does it fail in the first place when integrations are merged?

src/main/kotlin/app/revanced/patches/instagram/patches/integrations/patch/FixIntegrationsPatch.kt
```diff
@@ -0,0 +1,46 @@
...
+import app.revanced.patches.instagram.patches.integrations.fingerprints.ReflectionFingerprint
+import org.jf.dexlib2.dexbacked.reference.DexBackedFieldReference
+import org.jf.dexlib2.dexbacked.reference.DexBackedTypeReference
+import org.jf.dexlib2.iface.instruction.formats.Instruction21c
+
+@Patch
+@Name("fix-integrations")
+@Description("Stops app from crashing when integrations are merged.")
+@Compatibility([Package("com.instagram.android")])
+@Version("0.0.1")
```
oSumAtrIX, line 19: This patch should only be used as a dependency for other patches which depend on possibly future integrations
```suggestion
@Description("Prevents crashing when new classes are added.")
@Version("0.0.1")
```

src/main/kotlin/app/revanced/patches/instagram/patches/integrations/patch/FixIntegrationsPatch.kt
```diff
@@ -0,0 +1,46 @@
...
+        ReflectionFingerprint
+    )
+) {
+    override fun execute(context: BytecodeContext): PatchResult {
+        ReflectionFingerprint.apply {  ->
+            val startIndex = result?.scanResult?.patternScanResult?.startIndex
+                ?: return toErrorResult()
+
+            val reference = ((result!!.method.implementation!!.instructions.elementAt(startIndex) as Instruction21c).reference as DexBackedTypeReference).type
+            val field = ((result!!.method.implementation!!.instructions.elementAt(startIndex+2) as Instruction21c).reference as DexBackedFieldReference)
```
oSumAtrIX, line 31: You can use result.mutableMethod.instruction() here. Please reduce duplicates here.

src/main/kotlin/app/revanced/patches/instagram/patches/integrations/patch/FixIntegrationsPatch.kt
```diff
@@ -0,0 +1,46 @@
...
+            val fieldname = field.name
+            val classname = field.definingClass
```
oSumAtrIX, line 33: Please give the variables a proper explanatory name. It is unclear what they are.

oSumAtrIX (changes requested):
Please check on the review

PR #1711, opened 2023-03-06
refactor(tiktok/tiktok-settings): slightly better patch method

kazimmt:
fixes: ReVanced/revanced-patches-template#1573


oSumAtrIX:
There's too many changes, also a refactor isn't a fix, please split the PR.

PR #1745, opened 2023-03-14
feat(youtube/protobuf-spoof): video playback fix

LisoUseInAIKyrios:
Imported code from Inotia00

I expanded the method signature, and added an additional check to ensure the walked method is what's expected.


Axelen123:
Shouldn't the MicroG patch depend on this?

gotham8x:
Should we add credit to "Inotia00" somewhere in the README or documents/contributors?

oSumAtrIX:
The commit is referenced to this PR where the credit can be given. In README it is autogenerated from commits

PR #1751, opened 2023-03-14
feat(patch): Add shared patch to expose activities of an app

bennett-sh:
This adds the patch as explained in ReVanced/revanced-patches-template#1259

### DIFF
--- a/src/main/kotlin/app/revanced/patches/all/activity/exportAll/patch/ExportAllActivitiesPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/all/activity/exportAll/patch/ExportAllActivitiesPatch.kt
@@ -20,17 +20,18 @@ class ExportAllActivitiesPatch : ResourcePatch {
             val activities = document.getElementsByTagName("activity")
 
             for(i in 0..activities.length) {
-                activities.item(i)?.let {
-                    val exportedAttribute = it.attributes.getNamedItem(EXPORTED_FLAG)
-                    if (exportedAttribute != null) {
-                        if(exportedAttribute.nodeValue != "true")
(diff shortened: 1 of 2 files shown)


src/main/kotlin/app/revanced/patches/all/activity/exportAll/patch/ExportAllActivitiesPatch.kt
```diff
@@ -0,0 +1,44 @@
...
+                    val exportedAttribute = it.attributes.getNamedItem(EXPORTED_FLAG)
+                    if (exportedAttribute != null) {
+                        if(exportedAttribute.nodeValue != "true")
+                            exportedAttribute.nodeValue = "true"
+                        return@let
+                    }
+
+                    document.createAttribute(EXPORTED_FLAG)
+                        .apply { value = "true" }
+                        .let(it.attributes::setNamedItem)
```
oSumAtrIX, line 33: Why is this required in addition to the code above?

oSumAtrIX (approved):

src/main/kotlin/app/revanced/patches/all/activity/exportAll/patch/ExportAllActivitiesPatch.kt
bennett-sh: Apps targeting a SDK below Android 12 are not required to include the attribute iirc so this code is for adding it if it doesn't already exist.

oSumAtrIX (approved):

bennett-sh:
apps can mark activities as not exported. this means that only the system &
the app itself can launch them. this patch removes these restrictions
allowing apps like activity launcher to open them

Issue #1758, opened 2023-03-16
feat(patch): Crunchyroll DL (STILL WORKS)


juniorjacob:
### Application

Crunchyroll (https://www.apkmirror.com/apk/ellation-inc/crunchyroll/crunchyroll-3-25-1-release/)
Currently on 3.25.1

### Issue

!!!BRING BACK DOWNLOADING!!!

The old patch continues to work TO THIS DAY!

### Patch

Allow the player to sync like a Premium+ Subscription

### Motivation

YOU ALREADY WROTE THE CODE, SO READD IT!

### Acknowledgements

- [X] I have searched the existing issues and this is a new and no duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I filled out all of the requested information in this issue properly.

juniorjacob:
@johnconner122 @oSumAtrIX 

I have tested this feature thoroughly for every season of Attack on Titan. For 3-4 days I lacked internet & I disabled cellular after realizing the synced videos worked without it. I have tested it with & without internet. Downloaded / Synced videos play better and work on Airplane Mode.

I will admit that when STARTING a series you need some tricks to get it loaded. When I first started watching Attack on Titan, no episodes were watched yet, and tried to skip to any episode other than episode 1 it wouldn't load UNLESS I loaded up the precursor episodes.

Simple process:
1. Play the episode without any Watch History closest to the episode desired
2. Once loaded go off the episode
3. Play the next episode until it loads
4. Repeat until at desired episode

This wasn't required **_every_** time but when I first tested Attack on Titan, when I still had internet and intentionally disabled it, this was necessary. While being annoying, if prefaced to users then I see no harm. After all, they **do** have functional downloaded versions of the videos.

Please add this feature back, I really want to update my Crunchyroll 😌

oSumAtrIX:
Before adding back the patch, the issue you mentioned would have to be solved. For that please open a draft PR that reverts the commit that removes this patch.

juniorjacob:
Working on it

juniorjacob:
> Working on it

#1759

PR #1759, opened 2023-03-16
feat(crunchyroll): `downloads` patch

juniorjacob:
# TODO

- [ ] https://github.com/revanced/revanced-patches/issues/1758#issuecomment-1471109121


oSumAtrIX (changes requested):
This patch still has the issue which is mentioned in this PR.

Issue #1756, opened 2023-03-15
The "Remember playback patch "doesn't  work 


EhabxAlaa:
### Type

Error at runtime

### Bug description

Video playback speed is always remembered no matter what the state of the switch is, so now videos start with the same speed as the last played video

### Steps to reproduce

1-Include "Remember playback speed patch"
2-Select a video and change the playback speed
3-Now this playback speed is also applied to next videos no matter what the state of the switch is (Playback speed is always remembered)

### Relevant log output

```shell
No output
```

### Additional context

The problem was there since the introduction of the patch , i thought people will report it and it will get fixed quickly but no one did

### Acknowledgements

- [X] I have searched the existing issues and this is a new and no duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I filled out all of the requested information in this issue properly.

EhabxAlaa:
> You're correct. The switch does nothing and the playback rate is always remembered.
> 
> `REMEMBER_PLAYBACK_RATE_SELECTED` in integrations has the wrong key (it should be `revanced_remember_playback_rate_last_selected`
> 
> But when I fix that, the playback rate cannot be changed when the feature is turned off (playback rate does not change). This will need a bit more time to figure out what's going wrong.
> 
> For now, exclude `remember-playback-rate` patch if you wish

Sure, thanks for replying but if i remember correctly excluding the patch doesn't fix the issue as every time you change the playback speed you get the pop-up of "playback speed remembered" but i will try to exclude it again and test it, and I will reply back to you

EhabxAlaa:
> You're correct. The switch does nothing and the playback rate is always remembered.
> 
> `REMEMBER_PLAYBACK_RATE_SELECTED` in integrations has the wrong key (it should be `revanced_remember_playback_rate_last_selected`
> 
> But when I fix that, the playback rate cannot be changed when the feature is turned off (playback rate does not change). This will need a bit more time to figure out what's going wrong.
> 
> For now, exclude `remember-playback-rate` patch if you wish

Nevermind excluding it works just fine ,my memory is just bad

PR #1762, opened 2023-03-16
fix(youtube/remember-playback-speed):  allow to not remember speed

### LINKED ISSUE #1254: The "Remember playback patch "doesn't  work
### Type

Error at runtime

### Bug description

Video playback speed is always remembered no matter what the state of the switch is, so now videos start with the same speed as the last played video

### Steps to reproduce

1-Include "Remember playback speed patch"
2-Select a video and change the playback speed
3-Now this playback speed is also applied to next videos no matter what the state of the switch is (Playback speed is always remembered)

### Relevant log output

```shell
No output
```

### Screenshots or videos

_No response_

### Solution

_No response_

### Additional context

The problem was there since the introduction of the patch , i thought people will report it and it will get fixed quickly but no one did

### Acknowledgements

LisoUseInAIKyrios:
fix for https://github.com/revanced/revanced-patches/issues/1756

[integration changes](https://github.com/revanced/revanced-integrations/pull/338)

### DIFF
--- a/src/main/kotlin/app/revanced/patches/youtube/misc/video/speed/remember/patch/RememberPlaybackSpeedPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/youtube/misc/video/speed/remember/patch/RememberPlaybackSpeedPatch.kt
@@ -36,27 +36,6 @@ import org.jf.dexlib2.iface.instruction.ReferenceInstruction
 class RememberPlaybackSpeedPatch : BytecodePatch(
     listOf(ChangePlaybackSpeedFragmentStateFingerprint)
 ) {
-    private companion object {
-        const val INTEGRATIONS_CLASS_DESCRIPTOR =
-            "Lapp/revanced/integrations/patches/playback/speed/RememberPlaybackSpeedPatch;"
-
-        fun MethodFingerprint.getReference(offsetFromPatternScanResultStartIndex: Int = 0) = this.result!!.let {
-            val referenceInstruction = it.mutableMethod
-                .instruction(it.scanResult.patternScanResult!!.startIndex + offsetFromPatternScanResultStartIndex) as ReferenceInstruction
-            referenceInstruction.reference.toString()
-        }
-
-        fun BytecodeContext.resolveFingerprints() {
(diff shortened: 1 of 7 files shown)


oSumAtrIX:
> It might be worth renaming the patch to "remember-playback-speed", as the YouTube UI uses 'speed' and not 'rate'.

In that case, refactoring changes will be needed to the class names and other variables, currently, the class name has "rate" but the method has "speed". Internally YouTube uses "rate" which is why I reflected it to the frontend.

src/main/kotlin/app/revanced/patches/youtube/misc/video/speed/remember/patch/RememberPlaybackSpeedPatch.kt
```diff
@@ -18,120 +18,123 @@ import app.revanced.patches.shared.settings.preference.impl.StringResource
...
                 )
             )
         )
 
         context.resolveFingerprints()
 
-        // Set the remembered playback rate.
-        InitializePlaybackRateValuesFingerprint.result!!.apply {
-            // Infer everything necessary for setPlaybackRate()
+        VideoIdPatch.injectCall("${INTEGRATIONS_CLASS_DESCRIPTOR}->newVideoLoaded(Ljava/lang/String;)V")
```
oSumAtrIX, line 61: Why is this required? The integrations just load the saved playback speed when loading a new video and then the speed is returned at the injection point. Can't the setting simply be returned?

oSumAtrIX (approved):

src/main/kotlin/app/revanced/patches/youtube/misc/video/speed/remember/patch/RememberPlaybackSpeedPatch.kt
oSumAtrIX: > then it needs to know when a new video is loaded so it can reset the video speed back to the default.

There is another injection point which is "getCurrentPlaybackSpeed". Why can't you just return the remembered speed there?

oSumAtrIX (approved):

PR #1786, opened 2023-03-22
feat(twitter): remove patch `monochrome-icon`

Axelen123:
The official app now has a monochrome icon, and that renders this patch obsolete.


oSumAtrIX (approved):

kazimmt:
Instead of removing, isn't it good to make this patch compatible with the version that doesn't have monochrome icon by default?
So, if someone patch an old version also can get monochrome icon.

PR #1793, opened 2023-03-24
fixes and extra features (ignore older commits) (apply this one)

CyclicPoet44782:
fixed couldnt apply the patch because settings did not succeed and similar
added telegram support, added extra features for some other apps


oSumAtrIX (changes requested):
The JSON is autogenerated, you should not commit to it.

PR #1794, opened 2023-03-24
Update README.md

CyclicPoet44782:
i just added support for telegram


reisxd:
That's not how it works.

oSumAtrIX (changes requested):
The README file is autogenerated, you should not commit to it.

PR #1801, opened 2023-03-25
feat(youtube/settings): disable preference control if the feature is turned off.  Show a dialog explaining side effects of some patches

LisoUseInAIKyrios:
added user notice message, explaining side effects of feature

Integration change: https://github.com/revanced/revanced-integrations/pull/328


src/main/kotlin/app/revanced/patches/youtube/misc/fix/playback/patch/SpoofSignatureVerificationPatch.kt
```diff
@@ -82,6 +82,9 @@ class SpoofSignatureVerificationPatch : BytecodePatch(
 
         } ?: return OpenCronetDataSourceFingerprint.toErrorResult()
 
+        SettingsPatch.addString("revanced_user_notice_message_signature_spoofing",
```
oSumAtrIX, line 85: Unless the user uses the switch, it will be unclear what the switch does. Usually, you'd use those dialogs to confirm the action taken or give a heads-up, not as a substitute to describe what the switch does.

src/main/kotlin/app/revanced/patches/youtube/layout/spoofappversion/bytecode/patch/SpoofAppVersionPatch.kt
```diff
@@ -59,6 +59,10 @@ class SpoofAppVersionPatch : BytecodePatch(
             )
         } ?: return SpoofAppVersionFingerprint.toErrorResult()
 
+        SettingsPatch.addString("revanced_user_notice_message_app_version_spoofing",
```
oSumAtrIX, line 62: Where is this string being used?

oSumAtrIX: The way I understood is that the string is used for the dialog string when toggling a setting. In that case, it is weird to inject the string like that. A better API would be that you can pass the string along with the setting it is associated with and let the settings patch add the string. What is the reason for separating it and adding the string manually?

oSumAtrIX: Currently, `SettingsPatch` incorporates a lot of boilerplate with how it works. Technically it is possible to reduce boilerplate and improve API design in the following way:


```diff
SettingsPatch.PreferenceScreen.LAYOUT.addPreferences(
    SwitchPreference(
        "revanced_spoof_app_version",
-        StringResource("revanced_spoof_app_version_title", "Spoof app version"),
+        title = "Spoof app version",
-        false,
+        default = false,
-        StringResource("revanced_spoof_app_version_summary_on", "Version spoofed to 17.30.34"),
+        summaryOn = "Version spoofed to 17.30.34",
-        StringResource("revanced_spoof_app_version_summary_off", "Version not spoofed")
+        summaryOff = "Version not spoofed"
   )
)
````

This way the API can be used conveniently:

```kotlin
SettingsPatch.PreferenceScreen.LAYOUT.addPreferences(
    SwitchPreference(
        "revanced_spoof_app_version",
        "Spoof app version"
        false,
        "Version spoofed to 17.30.34",
        "Version not spoofed",
        dialogString = "App version will be spoofed to 17.30.34. This will give the old UI layout, but unknown side effects may occur. If later turned off, the old UI layout may remain until you log out or clear the app data."
   )
)
```

The patch would simply autogenerate the proper keys based on the property, reducing boilerplate. For that reason, for now, adding the string should be incorporated into the API the following way. An example:

```diff
SettingsPatch.PreferenceScreen.LAYOUT.addPreferences(
    SwitchPreference(
        "revanced_spoof_app_version",
        StringResource("revanced_spoof_app_version_title", "Spoof app version"),
        false,
        StringResource("revanced_spoof_app_version_summary_on", "Version spoofed to 17.30.34"),
        StringResource("revanced_spoof_app_version_summary_off", "Version not spoofed")
+       dialogString = StringResource("revanced_user_notice_message_app_version_spoofing", "App version will be spoofed to 17.30.34. This will give the old UI layout, but unknown side effects may occur. If later turned off, the old UI layout may remain until you log out or clear the app data.")
    )
)
````

Comment (oSumAtrIX) [reply src/main/kotlin/app/revanced/patches/youtube/layout/spoofappversion/bytecode/patch/SpoofAppVersionPatch.kt:62]: I was having something more simpler in mind. You could simply make `SettingsPatch` add the string and use the key for the string in `SettingsEnum`. This way you don't have to manually use `SettingsPatch.addString()`.

File: src/main/kotlin/app/revanced/patches/shared/settings/preference/impl/SwitchPreference.kt [hunk 3]
```diff
@@ -16,15 +13,21 @@ import org.w3c.dom.Element
...
  * @param summaryOff The summary to show when the preference is disabled.
+ * @param userDialogMessage dialog message to show, if the user changes setting from the default value.
  */
 internal class SwitchPreference(
     key: String, title: StringResource,
     val default: Boolean = false,
     val summaryOn: StringResource? = null,
-    val summaryOff: StringResource? = null
+    val summaryOff: StringResource? = null,
+    userDialogMessage: StringResource? = null
```
oSumAtrIX, line 23: Is this appropriate here? Aren't their other preferences that might need a `userDialogMessage`? If there might, you can create an abstract "`BaseDialogMessagePreference`" class.

src/main/kotlin/app/revanced/patches/youtube/misc/settings/resource/patch/SettingsResourcePatch.kt
```diff
@@ -77,8 +77,8 @@ class SettingsResourcePatch : AbstractSettingsResourcePatch(
             )
         )
 
-        SettingsPatch.addString("revanced_enable_setting_user_notice_message_title_text",
-            "Do you wish to proceed?")
+        addString("revanced_settings_confirm_user_dialog_title",
+            "Do you wish to proceed?", false)
```
oSumAtrIX, line 81: Conventionally, strings are added via a host XML file. The `downloads` patch for example uses a [host](https://github.com/revanced/revanced-patches/blob/80c1d45c1cd6bb590d67071a931d908c86ee8a2a/src/main/resources/downloads/host/values/strings.xml) file to [merge](https://github.com/revanced/revanced-patches/blob/80c1d45c1cd6bb590d67071a931d908c86ee8a2a/src/main/kotlin/app/revanced/patches/youtube/interaction/downloads/resource/patch/DownloadsResourcePatch.kt#L55) when patching. This is recommended, so that in the future when new strings are added, we don't have to add multiple `addString` calls.

src/main/kotlin/app/revanced/patches/shared/settings/preference/impl/SwitchPreference.kt
```diff
@@ -16,15 +13,21 @@ import org.w3c.dom.Element
  * @param default The default value of the switch.
  * @param summaryOn The summary to show when the preference is enabled.
  * @param summaryOff The summary to show when the preference is disabled.
+ * @param userDialogMessage dialog message to show, if the user changes setting from the default value.
```
oSumAtrIX, line 16:
```suggestion
 * @param userDialogMessage The message to show in a dialog when the user toggles the preference.
```

src/main/kotlin/app/revanced/patches/shared/settings/preference/impl/SwitchPreference.kt
```diff
@@ -16,15 +13,21 @@ import org.w3c.dom.Element
...
     val summaryOn: StringResource? = null,
-    val summaryOff: StringResource? = null
+    val summaryOff: StringResource? = null,
+    userDialogMessage: StringResource? = null
 ) : BasePreference(key, title) {
     override val tag: String = "SwitchPreference"
 
+    init {
+        userDialogMessage?.include()
+    }
```
oSumAtrIX, line 29: While this technically works, it isn't really a clean solution. Instead, the string should be merged the same way as the other parameter strings in the `serialize` method below.

src/main/kotlin/app/revanced/patches/shared/settings/preference/impl/SwitchPreference.kt
```diff
@@ -16,16 +13,19 @@ import org.w3c.dom.Element
...
     val default: Boolean = false,
     val summaryOn: StringResource? = null,
-    val summaryOff: StringResource? = null
+    val summaryOff: StringResource? = null,
+    val userDialogMessage: StringResource? = null
 ) : BasePreference(key, title) {
     override val tag: String = "SwitchPreference"
 
     override fun serialize(ownerDocument: Document, resourceCallback: ((IResource) -> Unit)?): Element {
+        userDialogMessage?.include()
```
oSumAtrIX, line 28: While this works, out of consistency and convention reasons this `serialize` method should look like this:

```diff
 override fun serialize(ownerDocument: Document, resourceCallback: ((IResource) -> Unit)?): Element {
-        userDialogMessage?.include()
        return super.serialize(ownerDocument, resourceCallback).apply {
            addDefault(default)
            addSummary(summaryOn?.also { resourceCallback?.invoke(it) }, SummaryType.ON)
            addSummary(summaryOff?.also { resourceCallback?.invoke(it) }, SummaryType.OFF)
+           addUserDialogMessage(userDialogMessage?.also { resourceCallback?.invoke(it) }
        }
}
```

oSumAtrIX: That is a good point. What do you suggest? The [initial approach](https://github.com/revanced/revanced-patches/pull/1801/commits/f6f5efb76b1a071bde0c8f371a56de43e12dfe4d) seems a bit off to me as well since as you say, the string is not serialized into the XML switch preference.

oSumAtrIX: You noted that it is not part of the process of serializing the preference into XML though which I agree with. While it's easy to simply add it to `init` or to `serialize` both approaches don't seem right to me.

oSumAtrIX: That would lead to the initial issue that the string is meant for preference and not the patch, doesn't sound really right to me.

oSumAtrIX: The preference is meant for the patch. The string is meant for preference. This is a transitive dependency, for that reason the logic to add the string should be where the logic for the preference is.

oSumAtrIX: Alright, in that case, it should be fine as is for now.

oSumAtrIX (approved):

PR #1828, opened 2023-04-01
feat: inshorts ads patch

Aunali321:
Removes ads from Inshorts app.

### DIFF
--- a/src/main/kotlin/app/revanced/patches/inshorts/ad/fingerprints/InshortsAdsFingerprint.kt
+++ b/src/main/kotlin/app/revanced/patches/inshorts/ad/fingerprints/InshortsAdsFingerprint.kt
@@ -5,13 +5,5 @@ import org.jf.dexlib2.Opcode
 
 object InshortsAdsFingerprint : MethodFingerprint(
     "V",
-    strings = listOf("GoogleAdLoader","exception in requestAd"),
-    opcodes = listOf(
-        Opcode.INVOKE_VIRTUAL_RANGE,
-        Opcode.GOTO,
-        Opcode.MOVE_EXCEPTION,
-        Opcode.CONST_STRING,
-        Opcode.CONST_STRING,
-        Opcode.INVOKE_STATIC,
-     ),
+    strings = listOf("GoogleAdLoader","exception in requestAd")
 )
--- a/src/main/kotlin/app/revanced/patches/inshorts/ad/fingerprints/InshortsAdsFingerprint.kt
+++ b/src/main/kotlin/app/revanced/patches/inshorts/ad/fingerprints/InshortsAdsFingerprint.kt
@@ -1,0 +1,8 @@
+package app.revanced.patches.inshorts.ad.fingerprints
+
+import app.revanced.patcher.fingerprint.method.impl.MethodFingerprint
+
+object InshortsAdsFingerprint : MethodFingerprint(
+    "V",
+    strings = listOf("GoogleAdLoader","exception in requestAd")
+)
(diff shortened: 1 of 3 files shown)


PalmDevs:
I thought we're adding ads to Manager for April Fools.

src/main/kotlin/app/revanced/patches/inshorts/ad/fingerprints/InshortsAdsFingerprint.kt
```diff
@@ -0,0 +1,8 @@
+package app.revanced.patches.inshorts.ad.fingerprints
+
+import app.revanced.patcher.fingerprint.method.impl.MethodFingerprint
+
+object InshortsAdsFingerprint : MethodFingerprint(
+    "V",
+    strings = listOf("GoogleAdLoader","exception in requestAd")
```
oSumAtrIX, line 7: This is quite generic, could probably work for a lot of other apps as well.

src/main/kotlin/app/revanced/patches/inshorts/ad/fingerprints/InshortsAdsFingerprint.kt
```diff
@@ -1,8 +1,17 @@
 package app.revanced.patches.inshorts.ad.fingerprints
 
 import app.revanced.patcher.fingerprint.method.impl.MethodFingerprint
+import org.jf.dexlib2.Opcode
 
 object InshortsAdsFingerprint : MethodFingerprint(
     "V",
-    strings = listOf("GoogleAdLoader","exception in requestAd")
+    strings = listOf("GoogleAdLoader","exception in requestAd"),
+    opcodes = listOf(
```
oSumAtrIX, line 9: Why was an opcode pattern added to the previously working fingerprint?

src/main/kotlin/app/revanced/patches/inshorts/ad/patch/InshortsAdsPatch.kt
```diff
@@ -0,0 +1,41 @@
...
+@Name("hide-ads")
+@Description("Removes ads from Inshorts.")
+@HideAdsCompatibility
+@Version("0.0.1")
+class HideAdsPatch : BytecodePatch(
+    listOf(InshortsAdsFingerprint)
+) {
+    override fun execute(context: BytecodeContext): PatchResult {
+
+        val requestAdMethod = InshortsAdsFingerprint.result!!.mutableMethod
```
oSumAtrIX, line 30: If you assert here it could fail. Check on the latest patch that was added in commit history how to do it better.

src/main/kotlin/app/revanced/patches/inshorts/ad/fingerprints/InshortsAdsFingerprint.kt
```diff
@@ -0,0 +1,17 @@
+package app.revanced.patches.inshorts.ad.fingerprints
+
+import app.revanced.patcher.fingerprint.method.impl.MethodFingerprint
+import org.jf.dexlib2.Opcode
+
+object InshortsAdsFingerprint : MethodFingerprint(
```
oSumAtrIX, line 6: Is the method this fingerprint resolves to an Inshort specific thing or a method shaded from the Google ads library? If it is the latter, the patch should be converted into a generic Google ads patch.

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/inshorts/ad/fingerprints/InshortsAdsFingerprint.kt
```diff
@@ -0,0 +1,17 @@
...
+    "V",
+    strings = listOf("GoogleAdLoader","exception in requestAd"),
+    opcodes = listOf(
+        Opcode.INVOKE_VIRTUAL_RANGE,
+        Opcode.GOTO,
+        Opcode.MOVE_EXCEPTION,
+        Opcode.CONST_STRING,
+        Opcode.CONST_STRING,
+        Opcode.INVOKE_STATIC,
+     ),
```
oSumAtrIX, line 16: Is the opcode pattern still necessary? It could break the fingerprint in the future

Aunali321: Fixed

oSumAtrIX (approved):

PR #1832, opened 2023-04-01
feat: add shared patch to change the package name

andrisas:
Add a patch that replaces the package name in `AndroidManifest.xml `
This can be used to install multiple versions of the same app or to patch system apps without root

### DIFF
--- a/src/main/kotlin/app/revanced/patches/all/misc/packagename/patch/PackageNamePatch.kt
+++ b/src/main/kotlin/app/revanced/patches/all/misc/packagename/patch/PackageNamePatch.kt
@@ -1,0 +1,42 @@
+package app.revanced.patches.all.misc.packagename.patch
+
+import app.revanced.patcher.annotation.Description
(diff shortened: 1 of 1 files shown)


oSumAtrIX:
Many apps could stop working as they will have references to the old package name.

oSumAtrIX:
Please review if the changes work correctly

andrisas:
i get this error when installing the apk when the same app is installed with the original package name (this did not happen before because it replaced every reference to the old package name in `AndroidManifest.xml`) `failed to install out.apk: Failure [INSTALL_FAILED_CONFLICTING_PROVIDER: Scanning Failed.: Can't install because provider name FROMPACKAGENAME.provider (in package TOPACKAGENAME) is already used by FROMPACKAGENAME]`
and i also had some issues with the default set to `null`

oSumAtrIX:
In that case replace the package name via a string replace.

PR #1864, opened 2023-04-06
feat: add shared patch to change the package name

andrisas:
i opened a new pr instead of #1832 because it used the wrong branch and changing it failed

### DIFF
--- a/src/main/kotlin/app/revanced/patches/all/misc/packagename/patch/ChangePackageNamePatch.kt
+++ b/src/main/kotlin/app/revanced/patches/all/misc/packagename/patch/ChangePackageNamePatch.kt
@@ -1,0 +1,44 @@
+package app.revanced.patches.all.misc.packagename.patch
+
+import app.revanced.patcher.annotation.Description
+import app.revanced.patcher.annotation.Name
+import app.revanced.patcher.annotation.Version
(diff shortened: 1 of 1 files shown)


src/main/kotlin/app/revanced/patches/all/misc/packagename/patch/ChangePackageNamePatch.kt
```diff
@@ -0,0 +1,44 @@
...
+        var originalPackageName = ""
+        packageName?.let { packageName ->
+            val packageNameRegex = Regex("^[a-z]\\w*(\\.[a-z]\\w*)+\$")
+            if (!packageName.matches(packageNameRegex)) return PatchResultError("Invalid package name")
+
+            context.xmlEditor["AndroidManifest.xml"].use { editor ->
+                val manifest = editor.file.getElementsByTagName("manifest").item(0) as Element
+                originalPackageName = manifest.getAttribute("package")
+            }
+            if (!originalPackageName.matches(packageNameRegex)) return PatchResultError("Failed to get the original package name")
```
oSumAtrIX, line 26:
```suggestion

            if (!originalPackageName.matches(packageNameRegex)) return PatchResultError("Failed to get the original package name")
```

PR #1873, opened 2023-04-08
fix(youtube/return-youtube-dislike): render dislikes when scrolling into the screen

LisoUseInAIKyrios:
Changed the existing patch point to fix [893](https://github.com/revanced/revanced-patches/issues/893)

And added an additional patch point to fix [1853](https://github.com/revanced/revanced-patches/issues/1853)

[integration changes](https://github.com/revanced/revanced-integrations/pull/350)


oSumAtrIX:
The implementation isn't really good, I am currently working on a better one, leaving PR open for the time being if my approach doesn't work.

oSumAtrIX:
It is odd to me, why two are necessary. It is likely easier to trace calls from both places to a common method call that really sets the text. That way you won't need to hook two places, but instead one. If that turns out to be too much work, I'd say two would be fine too as well for the time being.

src/main/kotlin/app/revanced/patches/youtube/layout/returnyoutubedislike/patch/ReturnYouTubeDislikePatch.kt
```diff
@@ -123,39 +130,44 @@ class ReturnYouTubeDislikePatch : BytecodePatch(
...
-            "Lapp/revanced/integrations/patches/ReturnYouTubeDislikePatch;"
-
-        const val ON_COMPONENT_CREATED_DESCRIPTOR =
-            "$INTEGRATIONS_PATCH_CLASS_DESCRIPTOR->onComponentCreated(Ljava/lang/Object;Ljava/util/concurrent/atomic/AtomicReference;)V"
+            "Lapp/revanced/integrations/returnyoutubedislike/ReturnYouTubeDislike;"
 
         private fun MethodFingerprint.toPatch(voteKind: Vote) = VotePatch(this, voteKind)
     }
 
+    private fun MutableMethod.insertShorts(index: Int, register: Int) {
```
oSumAtrIX, line 163: Maybe accompany this extension method with MutableMethod.insertPlayer accordingly?

oSumAtrIX (changes requested):

oSumAtrIX:
> The second hook handles every situation, except when the user dislikes (it works correctly when a user likes a video). It's probably because the UI is not updated after disliking.

In that case I recommend a different approach. One Fingeprint that locates and resolves to the method. Another two to resolve each index using the first fingerprint. If style allows, you can also find the first required index with the first fingerprint and that way save the use of a third fingerprint that way.

src/main/kotlin/app/revanced/patches/youtube/layout/returnyoutubedislike/fingerprints/TextComponentAtomicReferenceFingerprint.kt
```diff
@@ -0,0 +1,37 @@
+package app.revanced.patches.youtube.layout.returnyoutubedislike.fingerprints
```
oSumAtrIX, line 1: Sorry, I think I didn't clarify myself properly. You resolve one fingerprint on all classes/ methods. Later when you know the method, you can use the MethodFingerprint.resolve overload to resolve a method using another fingerprint. That other fingerprint does not need anything other than an opcode pattern if the fingerprint is only used to find the index of the matching pattern. That said, the first fingerprint is fine as is (looks like it could match another method accidentally because it doesn't look too unique) but the second fingerprint you need to resolve on the first fingerprints resulting method.

oSumAtrIX: Looks good, haven't seen some changes :)

src/main/kotlin/app/revanced/patches/youtube/layout/returnyoutubedislike/patch/ReturnYouTubeDislikePatch.kt
```diff
@@ -78,39 +80,69 @@ class ReturnYouTubeDislikePatch : BytecodePatch(
...
+        // region Hook all cases, except when a video is disliked
 
-                // Overwritten after injected code, which is why it can be used.
-                val clobberRegister = atomicReferenceInstruction.registerD
+        TextComponentSpecFingerprint.result?.let {
+            with(it.mutableMethod) {
+                val returnSpanIndex = it.scanResult.patternScanResult!!.endIndex
+                val spannedStringRegister = (instruction(returnSpanIndex) as OneRegisterInstruction).registerA
+                // Make sure the context register is not the same as the spanned string register.
+                val contextTempRegister = if (spannedStringRegister == 0) 1 else 0
```
oSumAtrIX, line 127: This is a bit strange; what is the motivation for this?

oSumAtrIX (approved):
Left some refactoring changes

src/main/kotlin/app/revanced/patches/youtube/layout/returnyoutubedislike/patch/ReturnYouTubeDislikePatch.kt
oSumAtrIX, line 127:
```suggestion
                val contextTempRegister = spannedStringRegister.xor(1)
```

src/main/kotlin/app/revanced/patches/youtube/layout/returnyoutubedislike/patch/ReturnYouTubeDislikePatch.kt
```diff
@@ -78,39 +80,69 @@ class ReturnYouTubeDislikePatch : BytecodePatch(
...
+        // region Hook all cases, except when a video is disliked
 
-                // Overwritten after injected code, which is why it can be used.
-                val clobberRegister = atomicReferenceInstruction.registerD
+        TextComponentSpecFingerprint.result?.let {
+            with(it.mutableMethod) {
+                val returnSpanIndex = it.scanResult.patternScanResult!!.endIndex
+                val spannedStringRegister = (instruction(returnSpanIndex) as OneRegisterInstruction).registerA
+                // Store the context in any free register, which is any register except the one already used by the return value
+                val contextTempRegister = spannedStringRegister.xor(1)
```
oSumAtrIX, line 127: I may note this might malfunction if `spannedStringRegister` is anything but 0 or 1.

oSumAtrIX:
Have sou checked how often the method gets called? Looks like a lower leveled hook as previously mentioned.

src/main/kotlin/app/revanced/patches/youtube/layout/returnyoutubedislike/fingerprints/TextComponentAtomicReferenceFingerprint.kt
```diff
@@ -0,0 +1,34 @@
...
+
+import app.revanced.patcher.extensions.or
+import app.revanced.patcher.fingerprint.method.impl.MethodFingerprint
+import org.jf.dexlib2.AccessFlags
+import org.jf.dexlib2.Opcode
+
+/**
+ * Matches to the same method as [TextComponentContextFingerprint].
```
oSumAtrIX, line 9:
```suggestion
 * Resolves against the same method that [TextComponentContextFingerprint] resolves to.
```

src/main/kotlin/app/revanced/patches/youtube/layout/returnyoutubedislike/fingerprints/TextComponentAtomicReferenceFingerprint.kt
```diff
@@ -0,0 +1,34 @@
...
+    parameters = listOf("L"),
+    opcodes = listOf(
+        Opcode.MOVE_OBJECT, // available unused register
+        Opcode.MOVE_OBJECT_FROM16,
+        Opcode.MOVE_OBJECT_FROM16,
+        Opcode.MOVE_FROM16,
+        Opcode.INVOKE_VIRTUAL, // CharSequence atomic reference
+        Opcode.MOVE_RESULT_OBJECT,
+        Opcode.CHECK_CAST,
+        Opcode.MOVE_OBJECT, // CharSequence reference, and control flow label.  Insert code here.
```
oSumAtrIX, line 23:
```suggestion
        Opcode.MOVE_OBJECT, // CharSequence reference, and control flow label. Insert code here.
```

src/main/kotlin/app/revanced/patches/youtube/layout/returnyoutubedislike/fingerprints/TextComponentContextFingerprint.kt
```diff
@@ -0,0 +1,24 @@
+package app.revanced.patches.youtube.layout.returnyoutubedislike.fingerprints
+
+import app.revanced.patcher.extensions.or
+import app.revanced.patcher.fingerprint.method.impl.MethodFingerprint
+import org.jf.dexlib2.AccessFlags
+import org.jf.dexlib2.Opcode
+
+/**
+ * Matches to the same method as [TextComponentAtomicReferenceFingerprint].
```
oSumAtrIX, line 9:
```suggestion
 * Resolves against the same method that [TextComponentContextFingerprint] resolves to.
```

src/main/kotlin/app/revanced/patches/youtube/layout/returnyoutubedislike/patch/ReturnYouTubeDislikePatch.kt
```diff
@@ -123,22 +141,23 @@ class ReturnYouTubeDislikePatch : BytecodePatch(
...
+                        insertShorts(insertIndex, spannedParameterRegister)
                     }
                 }
+
+            // Additional hook, called after user dislikes.
+            with(it.mutableMethod) {
+                val insertIndex = it.scanResult.patternScanResult!!.startIndex + 2
+                val insertRegister = (implementation!!.instructions.elementAt(insertIndex - 1)
+                        as OneRegisterInstruction).registerA
+                insertShorts(insertIndex, insertRegister)
```
oSumAtrIX, line 159:
```suggestion
                val overwriteRegister = (implementation!!.instructions.elementAt(insertIndex - 1)
                        as OneRegisterInstruction).registerA
                insertShorts(insertIndex, overwriteRegister)
```

oSumAtrIX (approved):

PR #1876, opened 2023-04-09
feat!:make debugging patch universal

andrisas:
add shared debugging patch and rename the original to `enable-debugging`


src/main/kotlin/app/revanced/patches/youtube/misc/debugging/patch/DebuggingPatch.kt
```diff
@@ -11,16 +11,17 @@ import app.revanced.patches.shared.settings.preference.impl.StringResource
...
 import app.revanced.patches.youtube.misc.integrations.patch.IntegrationsPatch
+import app.revanced.patches.all.misc.debugging.patch.SharedDebuggingPatch
 import app.revanced.patches.youtube.misc.settings.bytecode.patch.SettingsPatch
 import org.w3c.dom.Element
 
 @Patch
-@Name("debugging")
-@DependsOn([IntegrationsPatch::class, SettingsPatch::class])
-@Description("Adds debugging options.")
+@Name("debugging-extra")
```
oSumAtrIX, line 19: Both patch should be called "debugging", or "enable-debugging" if we want to convey what the patch does. The issue would be, that the shared patch would be selected by name as well. Any ideas to that?

andrisas: Maybe we can add an incompatiblePackages annotioation for that [like how the compatiblePackages work](https://github.com/revanced/revanced-patcher/blob/main/src/main/kotlin/app/revanced/patcher/annotation/CompatibilityAnnotation.kt#L11) but i still do not understand what is the problem with renaming the patches

oSumAtrIX: I believe naming one patch "enable-android-debugging" and the other "enable-debugging" wouldn't be too off?

andrisas: Thats fine too but i think `debugging` and `debugging-extra` describes its functions better and do not forget that both of them would show up in the manager or when listing patches with revanced-cli

oSumAtrIX: It is unclear what "extra" means wheread "enable-android-debugging" explicitly explains, that the debugging capabilities of Android are enabled for the app. For that reason, I'd say it would be better than an arbitrary suffix "extra".

andrisas:
Renamed debugging patches

src/main/kotlin/app/revanced/patches/all/misc/debugging/patch/DebuggingPatc.kt
```diff
@@ -0,0 +1,43 @@
...
+import app.revanced.patcher.patch.*
+import app.revanced.patcher.patch.annotations.DependsOn
+import app.revanced.patcher.patch.annotations.Patch
+import org.w3c.dom.Element
+
+@Patch
+@Name("enable-android-debugging")
+@Description("Makes the app debuggable.")
+@Version("0.0.1")
+class SharedDebuggingPatch : ResourcePatch {
```
oSumAtrIX, line 16: Rename this EnableAndroidDebuggingPatch

src/main/kotlin/app/revanced/patches/youtube/misc/debugging/patch/DebuggingPatch.kt
```diff
@@ -11,16 +11,17 @@ import app.revanced.patches.shared.settings.preference.impl.StringResource
...
 import app.revanced.patches.youtube.misc.settings.bytecode.patch.SettingsPatch
 import org.w3c.dom.Element
 
 @Patch
-@Name("debugging")
-@DependsOn([IntegrationsPatch::class, SettingsPatch::class])
-@Description("Adds debugging options.")
+@Name("enable-debugging")
+@DependsOn([IntegrationsPatch::class, SettingsPatch::class, SharedDebuggingPatch::class])
+@Description("Adds extra debugging options.")
```
oSumAtrIX, line 21:
```suggestion
@Description("Adds debugging options.")
```

andrisas:
changed class name and description of debugging patches

src/main/kotlin/app/revanced/patches/all/misc/debugging/patch/DebuggingPatc.kt
```diff
@@ -0,0 +1,43 @@
...
+import app.revanced.patcher.annotation.Version
+import app.revanced.patcher.data.ResourceContext
+import app.revanced.patcher.patch.*
+import app.revanced.patcher.patch.annotations.DependsOn
+import app.revanced.patcher.patch.annotations.Patch
+import org.w3c.dom.Element
+
+@Patch
+@Name("enable-android-debugging")
+@Description("Makes the app debuggable.")
```
oSumAtrIX, line 14:
```suggestion
@Description("Enables Android debugging capabilities.")
```

gradle.properties
```diff
@@ -1,2 +1,2 @@
 kotlin.code.style = official
-version = 2.168.0-dev.1
+version = 2.168.0-dev.1
```
oSumAtrIX, line 2:
```suggestion
version = 2.168.0-dev.1

```

oSumAtrIX (approved):

oSumAtrIX:
Thanks

Issue #1863, opened 2023-04-06
bug: `Hide create, clip and thanks button` messes up layout 


shivanshvermak:
### Type

Cosmetic

### Bug description

It's random but occurs very frequently. The text in the share button doesn't appear. Sometimes the text appears but the icon doesn't. And despite enabling this option, remix button is their, although the icon and text isn't showing. Please refer to the screenshot for context.

### Steps to reproduce

Normally patch the YouTube app (I am using 18.05.40) and have the required patch.

### Relevant log output

```shell
NA
```

### Screenshots or videos

![Screenshot_2023-04-06-16-13-30-412_app revanced android youtube-edit](https://user-images.githubusercontent.com/15206857/230355064-a6c8a27f-6dc0-4a38-83f3-ec10517cff1a.jpg)

### Acknowledgements

- [X] I have searched the existing issues and this is a new and no duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I filled out all of the requested information in this issue properly.

EhabxAlaa:
This happenes to me too , but not only the remix button starts popping up randomly, the Thanks and clip buttons as well, and the other buttons starts to disappear

LAMBORGUINI:
It's always happening to me too, you need to correct this error

RileyXX:
+1 for this issue. Same problem over here. Clean install on latest versions. Seems to be a bug. It looks like there is something weird going on behind the scenes with the buttons. Like actual button elements are being swapped or edited somehow by the yt app as you interact. So revanced ends up targeting the wrong button element. Sounds like someone needs to take another look at how these buttons are targeted in a future patch.

oSumAtrIX:
I agree

oSumAtrIX:
I believe it breaks because the identifier is checked instead of the path, might be wrong though, i forgot

LAMBORGUINI:
It depends on which device you are talking about that the layout is broken

LAMBORGUINI:
would you like to make a command for each icon , example , hide clip button , hide create button , hide thank you button , hide share button

RileyXX:
> Fixed

Great news thank you! 

This may be a dumb question but is there a place where we can download nightly versions or a dev branch apk of the revanced manager with the most recent updates? Or do we simply have to wait until there is a new release?

RileyXX:
> 

Good to know thank you! 

[...]

PR #1924, opened 2023-04-19
fix(youtube/hide-video-action-buttons): fix 'hide share button'

### LINKED ISSUE #1178: bug(youtube/hide-video-action-buttons): 'Hide create, clip and thanks button' sometimes shows broken layout
### Type

Cosmetic

### Bug description

It's random but occurs very frequently. The text in the share button doesn't appear. Sometimes the text appears but the icon doesn't. And despite enabling this option, remix button is there, although the icon and text isn't showing. Please refer to the screenshot for context.

### Steps to reproduce

Normally patch the YouTube app (I am using 18.05.40) and have the required patch.

### Screenshots or videos

Screenshot_2023-04-06-16-13-30-412_app revanced android youtube-edit

### Solution

_No response_

### Additional context

_No response_

### Acknowledgements

LisoUseInAIKyrios:
Consolidated "hide share button" setting into the existing "hide action buttons" (change required to fix broken layout issues)

Consolidated "hide like" and "hide dislike" into a single setting.  Since most users have segmented like/dislikes and both must be hidden together.

hide buttons

This PR appears to fix https://github.com/revanced/revanced-patches/issues/1863

[integration changes](https://github.com/revanced/revanced-integrations/pull/360)

### DIFF
--- a/src/main/kotlin/app/revanced/patches/youtube/layout/buttons/action/patch/HideButtonsPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/youtube/layout/buttons/action/patch/HideButtonsPatch.kt
@@ -64,13 +64,6 @@ class HideButtonsPatch : ResourcePatch {
                         false,
                         StringResource("revanced_hide_action_button_summary_on", "Buttons are hidden"),
                         StringResource("revanced_hide_action_button_summary_off", "Buttons are shown")
-                    ),
-                    SwitchPreference(
-                        "revanced_hide_share_button",
-                        StringResource("revanced_hide_share_button_title", "Hide share button"),
-                        false,
-                        StringResource("revanced_hide_share_button_summary_on", "Share button is hidden"),
-                        StringResource("revanced_hide_share_button_summaryoff", "Share button is shown")
                     )
                 ),
                 StringResource("revanced_hide_buttons_summary", "Hide or show buttons under videos")
--- a/src/main/kotlin/app/revanced/patches/youtube/layout/buttons/action/patch/HideButtonsPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/youtube/layout/buttons/action/patch/HideButtonsPatch.kt
@@ -60,10 +60,10 @@ class HideButtonsPatch : ResourcePatch {
                     ),
                     SwitchPreference(
                         "revanced_hide_action_button",
-                        StringResource("revanced_hide_action_button_title", "Hide create, clip and thanks buttons"),
+                        StringResource("revanced_hide_action_button_title", "Hide all other action buttons"),
                         false,
-                        StringResource("revanced_hide_action_button_summary_on", "Buttons are hidden"),
-                        StringResource("revanced_hide_action_button_summary_off", "Buttons are shown")
+                        StringResource("revanced_hide_action_button_summary_on", "Share, remix, clip, thanks, shop, live chat buttons are hidden"),
+                        StringResource("revanced_hide_action_button_summary_off", "Share, remix, clip, thanks, shop, live chat buttons are shown")
                     )
                 ),
                 StringResource("revanced_hide_buttons_summary", "Hide or show buttons under videos")
--- a/src/main/kotlin/app/revanced/patches/youtube/layout/buttons/action/patch/HideButtonsPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/youtube/layout/buttons/action/patch/HideButtonsPatch.kt
@@ -52,11 +52,11 @@ class HideButtonsPatch : ResourcePatch {
                         StringResource("revanced_hide_playlist_button_summary_off", "Playlist button is shown")
                     ),
                     SwitchPreference(
-                        "revanced_hide_action_button",
-                        StringResource("revanced_hide_action_button_title", "Hide all other action buttons"),
+                        "revanced_hide_action_buttons",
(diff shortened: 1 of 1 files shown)


0xrxL:
Probably, an alternative way to accomplish this goal, is to search between protobuf (or other type of web) request and change the actionbuttons visibility.

 Cheers!

oSumAtrIX:
So how I remember it was that the log spits out multiple of the same path so it's not differentiable. A hack I used was to count the amount of how often I see the action button path. Since the share button comes right after the dislike/like button, we can identify it among the other action buttons. So to allow hiding the share menu, simply check if it is the first occurrence of action button. If it is and the option to hide it is on, then hide it. If the option to hide the rest of the action buttons is on, simply hide them as well, otherwise show them.

0xrxL:
> So how I remember it was that the log spits out multiple of the same path so it's not differentiable. A hack I used was to count the amount of how often I see the action button path. Since the share button comes right after the dislike/like button, we can identify it among the other action buttons. So to allow hiding the share menu, simply check if it is the first occurrence of action button. If it is and the option to hide it is on, then hide it. If the option to hide the rest of the action buttons is on, simply hide them as well, otherwise show them.

Your solution only had one inconvenience: it hid Share Button instead Live Chat action button (on normal videos), because there wasn't (and still isn't) a way to determine if a video contains the Live Chat action button or not.

oSumAtrIX:
I believe that can be fixed. You can check for the first occurrence of action button. First occurrence will be share button. The remaining occurrences are the rest of the action buttons. It should be possible to hide share and other buttons separately that way.

0xrxL:
> I believe that can be fixed. You can check for the first occurrence of action button. First occurrence will be share button. The remaining occurrences are the rest of the action buttons. It should be possible to hide share and other buttons separately that way.

First occurrence could be Live Chat too, how you determine this?

oSumAtrIX:
> > So how I remember it was that the log spits out multiple of the same path so it's not differentiable. A hack I used was to count the amount of how often I see the action button path. Since the share button comes right after the dislike/like button, we can identify it among the other action buttons. So to allow hiding the share menu, simply check if it is the first occurrence of action button. If it is and the option to hide it is on, then hide it. If the option to hide the rest of the action buttons is on, simply hide them as well, otherwise show them.
> 
> Your solution only had one inconvenience: it hid Share Button instead Live Chat action button (on normal videos), because there wasn't (and still isn't) a way to determine if a video contains the Live Chat action button or not.

@0xrxL yes. There is no proper way currently to determine it. The only way would be to dig in the conversioncontext object (and possibly children) to check, if the action button is a share button for example

oSumAtrIX:
Does the order of the buttons change?

0xrxL:
> > > So how I remember it was that the log spits out multiple of the same path so it's not differentiable. A hack I used was to count the amount of how often I see the action button path. Since the share button comes right after the dislike/like button, we can identify it among the other action buttons. So to allow hiding the share menu, simply check if it is the first occurrence of action button. If it is and the option to hide it is on, then hide it. If the option to hide the rest of the action buttons is on, simply hide them as well, otherwise show them.
> > 
> > 
> > Your solution only had one inconvenience: it hid Share Button instead Live Chat action button (on normal videos), because there wasn't (and still isn't) a way to determine if a video contains the Live Chat action button or not.
> 
> @0xrxL yes. There is no proper way currently to determine it. The only way would be to dig in the conversioncontext object (and possibly children) to check, if the action button is a share button for example

The problem is (from the first time I suggest this mod): the new Layout.

All the aforementioned three/four buttons have all the same infos, so you can't identify them individually.

oSumAtrIX:
If we figure out how to traverse componentcontext children recursively we can very quickly figure out every components structure. Question is if at the time we figure it out, it can still be hidden, the place we hook is a recursive call already thar builds the component and its children recursively by streaming the buffer. So at the time we know about the conversion context children, it is already too late to return an EmptyComponent to end the recursion and hide the component. An external hook outside of the recursion would be required. That way on calling of that hook, a second hook in the recursion can figure out the components children and structure and callback to the first hook, so that the first hook returns an EmptyComponent and effectively hides the original component.

oSumAtrIX:
The componentcontext object overrides ToString and therefor its fields and members should be trivial to reverse.

oSumAtrIX:
Reversing the necessary methods to for example read a childrens text components text is what would be the difficult part.

oSumAtrIX:
> It's happens the most when the action buttons are scrolled to the right

I was not able to trigger the bug by scrolling to the left or right, have you tried blocking the first occurrence of the action button path builder and checking if it hides the share button?

oSumAtrIX:
What is the explanation for the buttons to be partially being hidden? Have you tried only hiding the share button instead of only showing it? When you block an action button, the recursion stops and children stop rendering, then the next action button is rendered. It is odd that the action button is partially visible. It might be possible, that the children are rerendered without the button itself, so it would at first block at a child and therefore hide the buttons content only, but you were able to hide all buttons properly, so something along these lines there should be an explanation.

oSumAtrIX:
Concurrency can't be a problem. The buffer is read sequentially for each component context. And the component context object is built recursively. While it might build components in multiple threads, each component itself is built sequentially/recursively, so it doesn't explain why the buttons are partially hidden.

oSumAtrIX:
Makes sense, in that case the PR is a solution for now as is.

oSumAtrIX (approved):

PR #1926, opened 2023-04-19
chore(photomath): Move compatibility annotation

badawoll:


oSumAtrIX:
This compatibility file should be removed entirely as the patch is merely a dependency of another patch for which the compatibility is checked only.

badawoll:
Done :)

oSumAtrIX (approved):

PR #1948, opened 2023-04-21
docs(readme): replace non-existing Play Store URL

AnonymousWP:
Since Vanced was obviously never released on the Play Store, replace the URL with the GitHub repository (where Vanced _used_ to be available) as the source.

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,7 +167,7 @@ The official Patch bundle provided by ReVanced and the community.
 | `premium-icon-reddit` | Unlocks premium Reddit app icons. | all |
 </details>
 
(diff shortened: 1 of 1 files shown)


reisxd:
The README is generated automatically, please do not PR to the README.

AnonymousWP:
> The README is generated automatically, please do not PR to the README.

Based on what is it automatically generated? Can't you make sure it replaces only that line? Should be doable I'd say.

oSumAtrIX:
Breaks with app/ patch agnostics. The ONLY place where we hardcode package names is in patches compatibility annotation. No other place should hardcode packages. Thanks for the PR non the less.

PR #1952, opened 2023-04-22
feat(id-austria): remove version constraint

1fexd:
Just tested the patch for version `2.6.1`, and it still works.


oSumAtrIX:
Probably dropping the version constraint is a good idea.

1fexd:
But what if they change something which breaks the patch in the future?

oSumAtrIX:
That is obviously a risk, but it always depends on the patch's resilience. I believe in this case it should be fairly resilient considering the implementation.

1fexd:
Alright, I have removed the versions array

oSumAtrIX (approved):

PR #1961, opened 2023-04-23
fix minor spelling mistake in patches.json

parknich081:
change "Blocks embedded steam ads using services like TTV.lol or PurpleAdBlocker." to "Blocks embedded stream ads using services like TTV.lol or PurpleAdBlocker.


oSumAtrIX:
This file is autogenerated, do not commit to it, instead commit to the patches "Description" annotation.

PR #2010, opened 2023-04-28
feat(youtube): move video settings to `Video` settings category

johnconner122:


oSumAtrIX:
On another note, do you consider these patches miscellaneous or interaction?

johnconner122:
> On another note, do you consider these patches miscellaneous or interaction?

We are not constantly interacting, we just set the quality or speed, and forget about it.

oSumAtrIX:
@LisoUseInAIKyrios The four menus are currently meant to be pages to hold a list of patch menus/settings. If we add another menu, just for a single patch, then this would break that convention, because now a fifth menu would soley be dedicated for the speed/quality patch. Keep in mind, the Settings patch is used the following way: "SettingsPatch.MISC.add(setting)". If you now add a quality/speed menu, now other patches would for some reason see: "SettingsPatch.QUALITY.add(setting)". Therefore it should be added to an existing menu and not in a dedicated one.

oSumAtrIX:
If you introduce this menu, other video related settings should be moved into it.

johnconner122:
I guess we are fine with current PR. 🤷

johnconner122:
Looks good.

oSumAtrIX:
Things like video timestamp don't belong in video playback menu.

oSumAtrIX:
The same applies to them, the menu is labeled as "Settings related to [...] Playback" which is where the conflict happens

oSumAtrIX:
The "root" menu is meant as categories. Is "speed" and "quality" a category? Are there patches that can be put inside the category in the future? If these menus are just made to host the current speed and quality settings, it loses it's meaning, which is why it should be inside the misc category instead.

src/main/kotlin/app/revanced/patches/youtube/misc/settings/bytecode/patch/SettingsPatch.kt
```diff
@@ -137,6 +137,7 @@ class SettingsPatch : BytecodePatch(
         val ADS = Screen("ads", "Ads", "Ad related settings")
         val INTERACTIONS = Screen("interactions", "Interaction", "Settings related to interactions")
         val LAYOUT = Screen("layout", "Layout", "Settings related to the layout")
+        val VIDEO = Screen("video", "Video", "Settings related to video playback")
```
oSumAtrIX, line 140:
```suggestion
        val VIDEO = Screen("video", "Video", "Settings related to the video player")
```

oSumAtrIX (changes requested):
A new package should be created. Currently, the category has been changed from `misc` to `video`, but no package `video` has been introduced.

oSumAtrIX (approved):

src/main/kotlin/app/revanced/patches/youtube/misc/video/speed/VideoSpeeds.kt
```diff
@@ -1,24 +0,0 @@
-package app.revanced.patches.youtube.misc.video.speed
```
johnconner122, line 1: @oSumAtrIX This file is empty due to some reasons!

oSumAtrIX: Please open a follow up PR, seems like this was caused by the merge conflict resolution.

PR #2021, opened 2023-04-29
feat(youtube/general-ads): hide multiple audio track button on video overlay

LisoUseInAIKyrios:
Hides the audio track button that appears on [some videos](https://youtu.be/1otF0N6surM).

The audio track is still selectable, by going into the playback settings (video overlay gear icon).

[integration changes](https://github.com/revanced/revanced-integrations/pull/377)


oSumAtrIX (changes requested):
This patch should be removed, and the preference should be added to the general ads patch preferences. This is because if the default value is true, regardless of whether this patch is included or not, you wouldn't be able to disable it. In this current case it is false, but for consistency reasons, it should be moved to where the rest of the litho toggles are; that way, refactoring in the future will be easier.

oSumAtrIX (approved):

PR #2031, opened 2023-04-30
fix(youtube/return-youtube-dislike): support older UI layouts

LisoUseInAIKyrios:
ReturnYouTubeDislike now supports the old layouts used with [spoofed old versions ](https://github.com/revanced/revanced-patches/pull/2013)

[integration changes](https://github.com/revanced/revanced-integrations/pull/378)


0xrxL:
As confirmation: after a new installation, the old layout is shown, before the new one appears after restarting the app.

So...a cleanup at level resource won't will happens soon.

oSumAtrIX (approved):

PR #2051, opened 2023-05-02
fix(youtube/sponsorblock): fix skip button in wrong location when full screen and comments visible

LisoUseInAIKyrios:
Found a different view group to use for the skip buttons, and it adjusts correctly when full screen comments are visible.

fixes https://github.com/revanced/revanced-patches/issues/1960

[integration changes](https://github.com/revanced/revanced-integrations/pull/387)

### DIFF
--- a/src/main/kotlin/app/revanced/patches/youtube/layout/sponsorblock/bytecode/fingerprints/ControlsOverlayFingerprint.kt
+++ b/src/main/kotlin/app/revanced/patches/youtube/layout/sponsorblock/bytecode/fingerprints/ControlsOverlayFingerprint.kt
@@ -5,7 +5,7 @@ import app.revanced.patcher.fingerprint.method.impl.MethodFingerprint
 import org.jf.dexlib2.AccessFlags
 import org.jf.dexlib2.Opcode
 
-object PlayerOverlaysLayoutInitFingerprint : MethodFingerprint(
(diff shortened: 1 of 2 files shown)


src/main/kotlin/app/revanced/patches/youtube/layout/sponsorblock/bytecode/fingerprints/PlayerOverlaysLayoutInitFingerprint.kt
```diff
@@ -1,9 +1,27 @@
...
+        Opcode.INVOKE_VIRTUAL,
+        Opcode.CONST,
+        Opcode.INVOKE_VIRTUAL,
+        Opcode.MOVE_RESULT_OBJECT,
+        Opcode.CHECK_CAST,
+        Opcode.NEW_INSTANCE,
+        ),
+    customFingerprint = { methodDef ->
+        methodDef.definingClass == "Lcom/google/android/apps/youtube/app/player/overlay/YouTubeControlsOverlay;"
+    }
```
oSumAtrIX, line 26: Since the fingerprint has been changed, have you considered if the existing version compatibility of this patch is still valid? If not, you can drop the old versions or test if it works.

oSumAtrIX (approved):

PR #2059, opened 2023-05-03
fix(readme-generator): attempt to sort according to semantic version specification

rstular:
Closes ReVanced/revanced-patches-template#1067

### DIFF
--- a/src/main/kotlin/app/revanced/meta/ReadmeGenerator.kt
+++ b/src/main/kotlin/app/revanced/meta/ReadmeGenerator.kt
@@ -6,12 +6,23 @@ import app.revanced.patcher.extensions.PatchExtensions.description
 import app.revanced.patcher.extensions.PatchExtensions.patchName
 import app.revanced.patcher.patch.Patch
 import java.io.File
+import java.lang.module.ModuleDescriptor.Version
 
 internal class ReadmeGenerator : PatchesFileGenerator {
     private companion object {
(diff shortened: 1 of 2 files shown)


oSumAtrIX (changes requested):
Looks good. [This](https://github.com/revanced/revanced-patches/issues/2019#issuecomment-1535949314) was suggested and is a better solution than abusing `ModuleDescriptor.Version`.

oSumAtrIX:
In that case please refactor this PR so that it uses the suggested solution.

src/main/kotlin/app/revanced/meta/ReadmeGenerator.kt
```diff
@@ -37,9 +38,9 @@ internal class ReadmeGenerator : PatchesFileGenerator {
                     }
                 }.let { commonMap ->
                     commonMap.maxByOrNull { it.value }?.value?.let {
-                        // This is not foolproof, because for example v1.0.0-dev.0 will be returned instead of v1.0.0-release.
-                        // Unfortunately this can not be solved easily because versioning can be complex.
-                        commonMap.entries.filter { mostCommon -> mostCommon.value == it }.maxBy { it.key }.key
+                        // Uses FlexVer (https://github.com/unascribed/FlexVer/) to perform version comparison
```
oSumAtrIX, line 41:
```suggestion
```

oSumAtrIX (approved):

PR #2060, opened 2023-05-03
feat(patch): Bump patch compatibility

badawoll:
Bump patch compatibility for PhotoMath & MemeGenerator.


src/main/kotlin/app/revanced/patches/photomath/misc/unlockplus/annotations/UnlockPlusCompatibilty.kt
```diff
@@ -23,6 +23,8 @@ import app.revanced.patcher.annotation.Package
             "8.19.0",
             "8.20.0",
             "8.21.0",
+            "8.21.1",
```
oSumAtrIX, line 26: It might be time to remove the constraint on versions as it seems that the patch works fine for future versions.

oSumAtrIX (approved):

oSumAtrIX (approved):

PR #2077, opened 2023-05-04
feat(youtube): import / export of revanced settings

### LINKED ISSUE #1863: feat(youtube): import / export revanced specific settings

LisoUseInAIKyrios:
Added option to import/export settings.
export settings

Only non default setting values are exported, which makes it simple to see what settings are changed.  Clearing any settings in the text dialog and then importing will restore the removed settings to the ReVanced default.

Settings imported / exported include both ReturnYouTubeDislike and SponsorBlock.  The existing SB import/export feature remains in the SB settings and remains unchanged (the feature must remain separate, as SB uses a different and more complex import/export scheme).

Since the SB user id is a sensitive piece of information, if the user has a SB user id then the import/export feature informs the user not to share their user id:
sponsorblock private user id warning
(random fun fact: your SB private user id can be a string of emojis 👀)

Internally SB was changed so it does not create a user id until it's actually needed (when a user votes, or creates a segment, or they've imported SB settings from the desktop extension).  For users that do not create segments and never vote for SB segments, they will never see the import/export SB user id warning as they will not have a user id to worry about leaking.

This change also removes the API calls for vip check and user stats for these passive SB users (since they cannot have VIP status or any stats to show).

[integration changes](https://github.com/revanced/revanced-integrations/pull/388)


src/main/resources/settings/host/values/strings.xml
```diff
@@ -2,4 +2,12 @@
 <resources>
     <string name="revanced_settings_confirm_user_dialog_title">Do you wish to proceed?</string>
     <string name="revanced_settings_reset">Reset</string>
+    <string name="revanced_settings_import">Import</string>
+    <string name="revanced_settings_import_sb_userid_present">Your settings contain a private SponsorBlock user id.\n\nYour user id is like a password and it should never be shared.\n</string>
```
oSumAtrIX, line 6: SponsorBlock is it's own patch. Settings patch is unaware about other patches. For that reason it should not mention SponsorBlock here.

0xrxL:
@LisoUseInAIKyrios Wouldn't it be easier to create buttons that save and load a json file than copying and importing through the preferences clipboard?

0xrxL:
> The dialog could have two additional options, "save to file" and "load from file.
> 
> But it should have a way to copy paste text. It's easier to share a config with others for diagnosing or just showing off their preferred configuration. Works on Reddit, Discord, or anywhere really. No file upload/download.

Well so...in this case keep both solutions. The first with a file load/save and the second for debugging.

0xrxL:
> If import/export to a file is added, it cannot be later removed.
> 
> Adding file import/export options makes the UI more complicated and would need consideration since additional actions are possible. The dialog box would need a custom view to support extra buttons (only 3 buttons are built in and all are used: copy, cancel, import). Or additional import/export preferences would be needed. Maybe something like: "Import / Export clipboard", "Import / Export file"
> 
> SponsorBlock supports importing / exporting by copy paste only, and so far nobody has asked for file support for that.
> 
> I think adding file support can be added later, if users demand it.

About the complexity:

Can't you (like on apps such AdGuard, when a user try to import/export its settings) use the built-in Android loader/writer manager?

It called just "File" and is a system app.

src/main/kotlin/app/revanced/patches/youtube/misc/video/quality/patch/RememberVideoQualityPatch.kt
```diff
@@ -43,18 +43,17 @@ class RememberVideoQualityPatch : BytecodePatch(
     override fun execute(context: BytecodeContext): PatchResult {
         SettingsPatch.PreferenceScreen.MISC.addPreferences(
             SwitchPreference(
-                "revanced_remember_video_quality_last_selected",
+                "revanced_video_remember_last_selected",
```
oSumAtrIX, line 46: This new key is not really suitable, considering that there is a remember playback and quality preference, they would conflict.

src/main/resources/sponsorblock/host/values/strings.xml
```diff
@@ -42,17 +42,21 @@
...
     <string name="sb_api_url_changed">API URL changed</string>
     <string name="sb_settings_ie">Import/Export settings</string>
+    <string name="sb_settings_copy">Copy</string>
     <string name="sb_settings_ie_sum">Your SponsorBlock JSON configuration that can be imported/exported to ReVanced and other SponsorBlock platforms. This includes your private user id. Be sure to share this wisely</string>
     <string name="sb_settings_import_successful">Settings imported successfully</string>
     <string name="sb_settings_import_failed">Failed to import: %s</string>
-    <string name="sb_settings_export_failed">Failed to export settings (try clearing app data)</string>
+    <string name="sb_settings_export_failed">Failed to export: %s</string>
+
+    <string name="sb_settings_revanced_export_user_id_warning">Your settings contain a private SponsorBlock user id.\n\nYour user id is like a password and it should never be shared.\n</string>
```
oSumAtrIX, line 58: Is exporting a feature, that the SponsorBlock patch is aware of? Otherwise it would be breaking modularity as SponsorBlock technically would be unaware of the feature.

oSumAtrIX (commented):
Quite a lot of changes. A convention has been used for all keys, but this PR breaks it in couple places. Please suggest a convention for the keys. For example "revanced_<patch>_...". For example in the case of the remember video quality patch you'd use "revanced_remember_video_quality_enabled". This convention will allow in the future to refactor the code so that the key is generated automatically by SettingsPatch by depending on this convention. This would drastically improve the API by not having the patch user to think about any keys, if not necessary.

oSumAtrIX:
Also, currently ReVanced integrations have to mirror the preferences due to limitations. An enum is used for the seperate preferences. The enum name of a preference should match it's key.

oSumAtrIX:
What do you mean by patch declarations?

src/main/kotlin/app/revanced/patches/youtube/misc/video/quality/patch/RememberVideoQualityPatch.kt
oSumAtrIX: "quality" was omitted, similarly "speed" would be. Therefor both keys equal.
Simply don't omit them to fix the issue: "revanced_remember_video_quality_last_selected".

src/main/kotlin/app/revanced/patches/shared/settings/preference/impl/ListPreference.kt
```diff
@@ -14,16 +14,16 @@ import org.w3c.dom.Element
...
  */
 internal class ListPreference(
     key: String,
     title: StringResource,
     val entries: ArrayResource,
     val entryValues: ArrayResource,
-    val default: String? = null,
-    val summary: StringResource? = null
+    val summary: StringResource? = null,
+    val default: String? = null, // Default does not need to be set if key is present in YouTube or TikTok SettingsEnum
```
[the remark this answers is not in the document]
oSumAtrIX: I think the default should not be nullable just because Twitch does not utilize it, instead it should map the structure of the XML preferences. SwitchPreference doesn't necessarily need "default" to be present in its args, thus can be nullable.

oSumAtrIX:
> But any localized strings (ie: revanced_debug_toast_on_error_user_dialog_message) should still use "revanced_"

Alright, in that case, we can decide on the convention of the SettingsPatch:

1. Omit the prefix "revanced_" for every string not present in `strings.xml` or similar
2. Keep it for every string for higher consistency.

If we settle on either, `SettingsPatch` can be refactored so that the user does not have to use `StringResource` but instead the strings, which would significantly simplify using the patch.

src/main/kotlin/app/revanced/patches/shared/settings/preference/BasePreference.kt
```diff
@@ -13,6 +13,7 @@ import org.w3c.dom.Element
 internal abstract class BasePreference(
     override val key: String,
     override val title: StringResource,
+    val summary: StringResource? = null,
```
oSumAtrIX, line 16: Each preference seems to have a `summary`. Thus, it should be present in `IPreference` whereas it should be serialized in this classes implementation of the method `serialize`. Each preference either extends from `BasePreference` or `IPreference`, thus the declaration of `serialize` should be present in `IPreference`

oSumAtrIX:
```kotlin

internal interface IPreference {
    /**
     * Key of the preference.
     */
    val key: String

    /**
     * Title of the preference.
     */
    val title: StringResource

    /**
     * Tag name of the preference.
     */
    val tag: String

    /**
     * Summary of the preference.
     */
    val summary: StringResource?

    /**
     * Serialize preference element to XML.
     *
     * @param ownerDocument Target document to create elements from.
     * @param resourceCallback Called when a resource has been processed.
     * @return The serialized element.
     */
    fun serialize(ownerDocument: Document, resourceCallback: ((IResource) -> Unit)? = null): Element
}

```

src/main/kotlin/app/revanced/patches/shared/settings/preference/impl/TextPreference.kt
```diff
@@ -13,17 +13,17 @@ import org.w3c.dom.Element
...
  * @param inputType The input type of the text preference.
- * @param default The default value of the text preference.
  * @param summary The summary of the text preference.
+ * @param default The default value of the text preference.
  */
 internal class TextPreference(
     key: String,
     title: StringResource,
-    var inputType: InputType = InputType.STRING,
+    summary: StringResource,
```
oSumAtrIX, line 22: Seems like some classes that extend of `BasePreference` do not propagate the correct type. Some classes have the `summary` of type `StringResource` and some of type `StringResource?` (nullable), whereas you might not always accompany a `TextPreference` with a summary for example.

src/main/kotlin/app/revanced/patches/shared/settings/preference/impl/SwitchPreference.kt
```diff
@@ -10,17 +10,17 @@ import org.w3c.dom.Element
...
+ * @param default The default value of the switch.
  */
 internal class SwitchPreference(
     key: String, title: StringResource,
-    val default: Boolean = false,
     val summaryOn: StringResource? = null,
     val summaryOff: StringResource? = null,
-    val userDialogMessage: StringResource? = null
+    val userDialogMessage: StringResource? = null,
+    val default: Boolean? = null,
```
oSumAtrIX, line 23: Why not:

```kotlin
val default: Boolean = false,
```

src/main/kotlin/app/revanced/patches/youtube/ad/general/resource/patch/GeneralAdsResourcePatch.kt
```diff
@@ -31,263 +31,234 @@ class GeneralAdsResourcePatch : ResourcePatch {
...
-                        "",
+                        "revanced_custom_filter_strings",
+                        StringResource("revanced_custom_filter_strings_title", "Custom filter"),
                         StringResource(
-                            "revanced_adremover_custom_strings_summary",
+                            "revanced_custom_filter_strings_summary",
                             "Filter components by their name separated by a comma"
-                        )
+                        ),
+                        InputType.TEXT
```
oSumAtrIX, line 261: This parameter is default and can be removed:
```kotlin
```

src/main/kotlin/app/revanced/patches/shared/settings/preference/impl/TextPreference.kt
oSumAtrIX: > Some existing code doesn't have a summary for the items.

Can you clarify what you mean by that?

src/main/kotlin/app/revanced/patches/shared/settings/preference/impl/SwitchPreference.kt
oSumAtrIX: I may not understand this correctly. I believe, Android treats preferences that do not specify the `default` attribute as false by default. This would mean, an if check on this parameter would suffice, to decide weather to write the attribute to the XML document, or not.

src/main/kotlin/app/revanced/patches/shared/settings/preference/impl/TextPreference.kt
oSumAtrIX: > SwitchPreference has no single summary and uses two (summary switch on, summary switch off)

That seems to me that we have a little issue with our abstraction. 
`SwitchPreference` extends to `BasePreference`, which implements `IPreference` with the `summary` property, but `SwitchPreference` does not have `summary` thus violating Liskov substitution.

oSumAtrIX: I'll push some changes in a minute.

oSumAtrIX:
Commit 7d24acf5c0fa71a26701ec61459cdfb33dd6f610 introduces a new base class that includes a `default` property so that some classes that have defaults can extend off it. Additionally, it gets rid of the useless interfaces.

PR #2094, opened 2023-05-07
 feat(youtube/android-debugging): exclude by default

trmdi:
According to @oSumAtrIX, it should be excluded by default, but currently it isn't: https://github.com/revanced/revanced-discussions/discussions/979#discussioncomment-5814187


oSumAtrIX:
`EnableAndroidDebuggingPatch` should be excluded by default. The `enable-debugging` patch for YouTube can be enabled by default.

PR #2096, opened 2023-05-07
Line no 13 (| `client-spoof` adding this -> (Also fix issues "the fol…

Parth-root:
Line no 13 (| `client-spoof` adding this -> (Also fix issues "the following content is not available") )

I had a problem while patching app can't start the video, it says the following content is not available, so adding it will make more users to solve the issues in YouTube revanced.


Parth-root:
`client-spoof` | Spoofs a patched client to allow playback.(Also fix issues "the following content is not available")

I had a problem while patching app can't start the video, it says the following content is not available, so adding it will make more users to solve the issues in YouTube revanced.

oSumAtrIX:
The file is autogenerated and should not be committed to. The existing description already says that it fixes playback issues, that includes the message you added afterwards.

PR #2100, opened 2023-05-07
docs: remove "the" in front of ReVanced names  

oSumAtrIX:
This PR does the following:

```diff
- the ReVanced Patcher
+ ReVanced Patcher
```

### DIFF
--- a/docs/0_preparation.md
+++ b/docs/0_preparation.md
@@ -10,7 +10,7 @@ To develop ReVanced patches, a certain development environment is required.
 
 ## 🏃 Prepare the environment
 
-For this guide, the [ReVanced Patches](https://github.com/revanced/revanced-patches) will be used as a base.
+For this guide, [ReVanced Patches](https://github.com/revanced/revanced-patches) will be used as a base.
 
 1. Clone the repository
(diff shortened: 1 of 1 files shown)


Sculas:
How about we include removing all of the unnecessary "ReVanced" mentions as well in this PR? It's fine if you put "ReVanced ..." once, but doing it everywhere is very excessive and doesn't read nicely.

oSumAtrIX:
`ReVanced [...]` is a full product name. If we take `ReVanced Patches` for example, we refer to the patches from our repository, whereas `patches` refers to the patches from `ReVanced Patches`. Even when the context is fully clear, every time you refer to `ReVanced Patches` you should use `ReVanced Patches` instead of any abbreviation or alternative form such as:

- the ReVanced patches
- the patches
- the ReVanced Patches
- the Patches
- patches
- RV patches
- ReV Patches
- ...

Always use the fully qualified name `ReVanced Patches`. An exemplary sentence for this would be:

`Are you using ReVanced Patches` instead of `Are you using the ReVanced patches`, `Are you using the patches` or similar.

PR #2119, opened 2023-05-10
feat(youtube): add `hide-shorts-components` patch

oSumAtrIX:
## About

This PR moves some classes and renames some members, most notably Shorts-related code, to its dedicated patch. It also introduces the capability to filter for the Protobuf buffer for future patches. It also introduces the new patch `hide-shorts-components` to hide shorts components and the navigation bar when playing Shorts.

The PR should be rebased and merged.

## RE: 

- https://github.com/revanced/revanced-integrations/pull/395


0xrxL:
Didn't you (in a relative near past) consider the Protobuf Buffer (should be ByteBuffer) to be as excessive heavy solution? This will open doors for patches that go much beyond shorts at this point.

oSumAtrIX:
It's the bytearray stream of ProtoBuf, hence ProtoBuf Buffer. Knuth Morris Prat is used to check for occurrences of byte patterns as an efficient way to search through it. It is possible to hook all methods of the class which reads of the buffer and find out which one returns useful strings so that it can be used instead of reading the entire array.

0xrxL:
> It's the bytearray stream of ProtoBuf, hence ProtoBuf Buffer. Knuth Morris Prat is used to check for occurrences of byte patterns as an efficient way to search through it. It is possible to hook all methods of the class which reads of the buffer and find out which one returns useful strings so that it can be used instead of reading the entire array.

And the same solution could be applied to litho filters (of general ads patch)? There are some mods who would benefit from this.

oSumAtrIX:
No, indexOf already implements KMP.

LisoUseInAIKyrios (approved):

src/main/kotlin/app/revanced/patches/youtube/layout/hide/shorts/resource/patch/HideShortsComponentsResourcePatch.kt
```diff
@@ -0,0 +1,109 @@
...
+                    ),
+                    SwitchPreference(
+                        "revanced_hide_shorts_share_button",
+                        StringResource("revanced_hide_shorts_share_button_title", "Hide share button"),
+                        false,
+                        StringResource("revanced_hide_shorts_share_button_on", "Share button is hidden"),
+                        StringResource("revanced_hide_shorts_share_button_off", "Share button is shown")
+                    ),
+                    SwitchPreference(
+                        "revanced_hide_shorts_like_button",
```
johnconner122, line 74: Hiding `like` & `dislike` button aren't implemented in [integrations](https://github.com/revanced/revanced-integrations/pull/395)!

LisoUseInAIKyrios (approved):

PR #2127, opened 2023-05-10
feat(youtube/theme): use default seekbar color when option is empty

aliernfrog:
Currently there's no way to have stock YT seekbar color without excluding `theme` patch. This PR makes the patch use stock YT seekbar color when `Seekbar color` option is empty (`""`).

Depends on https://github.com/revanced/revanced-integrations/pull/397


aliernfrog:
I'll close my PRs to not cause conflicts with the other one.
I am not able to test the other PR right now. But checking if the option is empty might be better than checking if it is red, since stock YT uses a white color for seekbar in dark mode.

PR #2133, opened 2023-05-11
Add expand button for josn example

sagarkpanda:
Adds an expand button for the josn example at the end of the readme


oSumAtrIX:
The file is auto generated, do not commit to it, instead to the template.

PR #2151, opened 2023-05-14
feat: add universal ad blocking patch

andrisas:
this pr adds universal patch to block ads

### TODOs

- [ ] code cleanups
- [ ] make block list changeable by user (move it to a json file)
- [ ] expand the block list
- [ ] testing

### DIFF
--- a/src/main/kotlin/app/revanced/patches/all/ads/blocklist/AndroidAdsBlocklist.kt
+++ b/src/main/kotlin/app/revanced/patches/all/ads/blocklist/AndroidAdsBlocklist.kt
@@ -1,0 +1,31 @@
+package app.revanced.patches.all.ads.blocklist
+
+import org.jf.dexlib2.Opcode
+
+val blockUrls: List<String> = listOf(
+    "adservice.google.com",
+    "g.doubleclick.net",
+    "googleadservices.com",
+    "amazon-adsystem.com",
+)
+
+val replaceUrlsWith: String = "https://example.com"
+
+val stringOpcodes: List<Opcode> = listOf(
+    Opcode.CONST_STRING
+)
+
+val blockInvokes: List<String> = listOf(
+    "loadAd",
+    "AdListener",
+    "AdLoad",
+    "AdView",
+)
+
+val invokeOpcodes: List<Opcode> = listOf(
+    Opcode.INVOKE_VIRTUAL,
+    Opcode.INVOKE_SUPER,
+    Opcode.INVOKE_DIRECT,
+    Opcode.INVOKE_STATIC,
+    Opcode.INVOKE_INTERFACE
+)
--- a/src/main/kotlin/app/revanced/patches/all/ads/bytecode/AdsBytecodePatch.kt
+++ b/src/main/kotlin/app/revanced/patches/all/ads/bytecode/AdsBytecodePatch.kt
@@ -1,0 +1,87 @@
+package app.revanced.patches.all.ads.bytecode
+
+import app.revanced.patcher.annotation.Description
+import app.revanced.patcher.annotation.Name
+import app.revanced.patcher.annotation.Version
+import app.revanced.patcher.data.BytecodeContext
+import app.revanced.patcher.patch.BytecodePatch
+import app.revanced.patcher.patch.PatchResult
+import app.revanced.patcher.patch.PatchResultSuccess
+import app.revanced.patcher.patch.annotations.DependsOn
+import app.revanced.patcher.patch.annotations.Patch
+import app.revanced.util.bytecode.*
+import app.revanced.patches.all.ads.blocklist.*
+import app.revanced.patcher.util.proxy.mutableTypes.MutableClass
+import app.revanced.patcher.util.proxy.mutableTypes.MutableMethod
+import org.jf.dexlib2.Opcode
+import org.jf.dexlib2.iface.instruction.formats.Instruction21c
+import org.jf.dexlib2.iface.instruction.formats.Instruction35c
+import org.jf.dexlib2.iface.instruction.formats.Instruction3rc
+import org.jf.dexlib2.iface.reference.MethodReference
+import org.jf.dexlib2.iface.reference.StringReference
+import org.jf.dexlib2.immutable.reference.ImmutableStringReference
+import org.jf.dexlib2.iface.ClassDef
+import org.jf.dexlib2.iface.Method
+import org.jf.dexlib2.builder.instruction.BuilderInstruction21c
+import org.jf.dexlib2.builder.instruction.BuilderInstruction10x
+
+
+@Patch
+@Name("ads")
+@Description("Removes all ads from the selected application.")
+@Version("0.0.1")
+class AdsBytecodePatch : BytecodePatch() {
+    override fun execute(context: BytecodeContext): PatchResult {
+        BytecodeUtils.transformIns(context, ::invokesTransformFunction, invokeOpcodes)
+        BytecodeUtils.transformIns(context, ::urlTransformFunction, stringOpcodes)
+        return PatchResultSuccess()
+    }
+
+    private fun invokesTransformFunction(Ins: Any, index: Int, methodDef: Method, classDef: ClassDef, context: BytecodeContext) {
+        if (Ins is Instruction35c) {
+            val instructionMethodReference = Ins.getReference() as MethodReference
+
+            if (blockInvokes.any{blockInvokes.contains(instructionMethodReference.getName())}) {
+                
+                // make class and method mutable, if not already
+                var mutableClass: MutableClass? = context.proxy(classDef).mutableClass
+                var mutableMethod: MutableMethod? =  mutableClass!!.methods.first {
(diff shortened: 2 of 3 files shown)


src/main/kotlin/app/revanced/patches/all/ads/bytecode/AdsBytecodePatch.kt
```diff
@@ -0,0 +1,77 @@
...
+import org.jf.dexlib2.iface.reference.StringReference
+import org.jf.dexlib2.immutable.reference.ImmutableStringReference
+import org.jf.dexlib2.iface.ClassDef
+import org.jf.dexlib2.iface.Method
+import org.jf.dexlib2.builder.instruction.BuilderInstruction21c
+import org.jf.dexlib2.builder.instruction.BuilderInstruction10x
+
+@Patch(false)
+@Name("ads")
+@Description("Removes all ads from the selected application.")
```
oSumAtrIX, line 26:
```suggestion
@Description("Attempts to remove ads.")
```

src/main/kotlin/app/revanced/patches/all/ads/bytecode/AdsBytecodePatch.kt
```diff
@@ -0,0 +1,77 @@
...
+@Version("0.0.1")
+class AdsBytecodePatch : BytecodePatch() {
```
oSumAtrIX, line 28: How about `RemoveAdsPatch`?

src/main/kotlin/app/revanced/patches/all/ads/blocklist/AndroidAdsBlocklist.kt
```diff
@@ -0,0 +1,31 @@
...
+    "AdView",
+)
+
+val invokeOpcodes: List<Opcode> = listOf(
+    Opcode.INVOKE_VIRTUAL,
+    Opcode.INVOKE_SUPER,
+    Opcode.INVOKE_DIRECT,
+    Opcode.INVOKE_STATIC,
+    Opcode.INVOKE_INTERFACE
+)
```
oSumAtrIX, line 31: You can check if the current instruction is a ReferenceInstruction and if the reference is a method reference if you want to take a shortcut, may be less desired though than just checking the opcode.

src/main/kotlin/app/revanced/patches/all/ads/blocklist/AndroidAdsBlocklist.kt
```diff
@@ -0,0 +1,31 @@
...
+val stringOpcodes: List<Opcode> = listOf(
+    Opcode.CONST_STRING
+)
+
+val blockInvokes: List<String> = listOf(
+    "loadAd",
+    "AdListener",
+    "AdLoad",
+    "AdView",
+)
```
oSumAtrIX, line 23: A comment explaining how you got the list of method names or what they correspond to would be nice. For example "Google ads" or "Amazon ads" etc.

src/main/kotlin/app/revanced/patches/all/ads/blocklist/AndroidAdsBlocklist.kt
```diff
@@ -0,0 +1,31 @@
+package app.revanced.patches.all.ads.blocklist
```
oSumAtrIX, line 1: This class isn't necessarily required; you can put it in the companion object of the patch that uses this class.

src/main/kotlin/app/revanced/patches/all/ads/bytecode/AdsBytecodePatch.kt
```diff
@@ -0,0 +1,77 @@
...
+    override fun execute(context: BytecodeContext): PatchResult {
+        BytecodeUtils.transformIns(context, ::invokesTransformFunction, invokeOpcodes)
+        BytecodeUtils.transformIns(context, ::urlTransformFunction, stringOpcodes)
+
+        return PatchResultSuccess()
+    }
+
+    private fun invokesTransformFunction(Ins: Any, index: Int, methodDef: Method, classDef: ClassDef, context: BytecodeContext) {
```
oSumAtrIX, line 36: Parameters start with a small letter; preferably, this repository avoids abbreviations like this. Instead of using `Any` as the type, you can use `Instruction` that is the lowest common type between all instructions.

src/main/kotlin/app/revanced/patches/all/ads/bytecode/AdsBytecodePatch.kt
```diff
@@ -0,0 +1,77 @@
...
+        if (Ins is Instruction35c) {
+            val instructionMethodReference = Ins.getReference() as MethodReference
+
+            if (blockInvokes.any{blockInvokes.contains(instructionMethodReference.getName())}) {
```
oSumAtrIX, line 40: This seems odd to me. Why would you use `contains` in `any`? Probably you wanted:

```suggestion
            if (blockInvokes.any{ it == instructionMethodReference.getName() }) {
```

src/main/kotlin/app/revanced/util/bytecode/BytecodeUtils.kt
```diff
@@ -0,0 +1,43 @@
...
+import org.jf.dexlib2.iface.Method
+
+
+internal object BytecodeUtils{
+    public fun transformIns(context: BytecodeContext, transformFn: (Ins: Any, index: Int, methodDef: Method, classDef: ClassDef, context: BytecodeContext) -> Any, opcodes: List<Opcode>) {
+        context.transformInstructions(transformFn, opcodes, context)
+    }
+
+
+    private fun BytecodeContext.transformInstructions(transformFn: (Ins: Any, index: Int, methodDef: Method, classDef: ClassDef, context: BytecodeContext) -> Any , opcodes: List<Opcode>, context:BytecodeContext) {
```
oSumAtrIX, line 18: Iirc such a method exists in the utils package.

oSumAtrIX, line 18: You can create a type alias for the big lambda so you don't have to type it twice.

oSumAtrIX (approved):

src/main/kotlin/app/revanced/util/bytecode/BytecodeUtils.kt
andrisas: where?

oSumAtrIX: In `app.revanced.util.patch`.

andrisas: if you are talking about [this](https://github.com/andrisas/revanced-patches/blob/20b2612efdcf08e6dacfced87bc0e6b580e85146/src/main/kotlin/app/revanced/util/patch/AbstractTransformInstructionsPatch.kt) it can not filter for opcodes.  Should i move/rename the one that i made?

Linus789: It is possible to filter for opcodes. Here is an example: https://github.com/revanced/revanced-patches/blob/43d22014c75e0845b506f50eb1e14e996e7d9272/src/main/kotlin/app/revanced/util/patch/MethodCall.kt#L83-L85

It is used by the `filterMap` function as you can see here:
https://github.com/revanced/revanced-patches/blob/43d22014c75e0845b506f50eb1e14e996e7d9272/src/main/kotlin/app/revanced/patches/all/connectivity/wifi/spoof/patch/SpoofWifiPatch.kt#L192-L202

src/main/kotlin/app/revanced/patches/all/ads/blocklist/AndroidAdsBlocklist.kt
```diff
@@ -0,0 +1,32 @@
...
+import org.jf.dexlib2.Opcode
+
+val blockUrls: List<String> = listOf(
+    "adservice.google.com",
+    "g.doubleclick.net",
+    "googleadservices.com",
+    "amazon-adsystem.com",
+)
+
+val replaceUrlsWith: String = "https://example.com"
```
Linus789, line 12: Would it be possible to use something like `http://127.0.0.1` to avoid the extra data traffic?

oSumAtrIX: I believe so, in the end it depends to the app. An app could crash if it can't reach a URL if it was previously sure, that it would reach it for example.

Axelen123: I believe 0.0.0.0 is preferred over 127.0.0.1 because there is no timeout for `0.0.0.0`. Most `/etc/hosts`-based adblockers use it for this reason. Wouldn't receiving unexpected content be more likely to crash apps compared to not getting anything at all?

kazimmt:
No improvement 😮‍💨

oSumAtrIX:
Is this still being worked on?

quyleanh:
The patch is promised, but it seems that author is inactivating now...

PR #2168, opened 2023-05-16
feat(vsco): add `unlock-pro` patch

redphx:
Unlock Pro features in [VSCO app](https://play.google.com/store/apps/details?id=com.vsco.cam).

### DIFF
--- a/src/main/kotlin/app/revanced/patches/vsco/misc/pro/fingerprints/RevCatSubscriptionFingerprint.kt
+++ b/src/main/kotlin/app/revanced/patches/vsco/misc/pro/fingerprints/RevCatSubscriptionFingerprint.kt
@@ -6,7 +6,6 @@ import app.revanced.patcher.fingerprint.method.impl.MethodFingerprint
 
 object RevCatSubscriptionFingerprint : MethodFingerprint(
     returnType = "V",
-    parameters = listOf("Z"),
     strings = listOf("use_debug_subscription_settings"),
     customFingerprint = { methodDef, _ ->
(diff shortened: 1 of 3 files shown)


src/main/kotlin/app/revanced/patches/vsco/misc/pro/fingerprints/RevCatSubscriptionFingerprint.kt
```diff
@@ -0,0 +1,15 @@
+package app.revanced.patches.vsco.misc.pro.fingerprints
+
+import org.jf.dexlib2.AccessFlags
+import app.revanced.patcher.extensions.or
+import app.revanced.patcher.fingerprint.method.impl.MethodFingerprint
+
+object RevCatSubscriptionFingerprint : MethodFingerprint(
```
oSumAtrIX, line 7: Probably the strings property is enough to locate the method, the rest might be removed.

redphx: No, because it might match this method:  

image

src/main/kotlin/app/revanced/patches/vsco/misc/pro/fingerprints/RevCatSubscriptionFingerprint.kt
```diff
@@ -0,0 +1,14 @@
...
+    returnType = "V",
+    parameters = listOf("Z"),
```
oSumAtrIX, line 9:
```suggestion
```

Powerm4KK (commented):
Duplicate of #

PR #2209, opened 2023-05-20
chore: remove scopes from issue templates

AnonymousWP:
# Added

- Add scope to `title` field. Alternatively, use `(patch)` instead of `(scope)`.

# Fixed

-  Fix typo in description of Patch Request

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-issue.yml
+++ b/.github/ISSUE_TEMPLATE/bug-issue.yml
@@ -1,6 +1,6 @@
 name: 🐞 Bug report
 description: Report a very clearly broken issue.
-title: 'bug: <title>'
+title: 'bug(scope): <title>'
 labels: [bug]
 body:
   - type: markdown
--- a/.github/ISSUE_TEMPLATE/bug-issue.yml
+++ b/.github/ISSUE_TEMPLATE/bug-issue.yml
@@ -1,6 +1,6 @@
 name: 🐞 Bug report
 description: Report a very clearly broken issue.
-title: 'bug(scope): <title>'
+title: 'bug: '
 labels: [bug]
 body:
   - type: markdown
--- a/.github/ISSUE_TEMPLATE/patch-request.yml
+++ b/.github/ISSUE_TEMPLATE/patch-request.yml
@@ -24,7 +24,7 @@ body:
   - type: textarea
     attributes:
       label: Motivation
-      description: Why should your patch request should be considered? What makes it valuable to the community?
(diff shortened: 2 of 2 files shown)


.github/ISSUE_TEMPLATE/bug-issue.yml
```diff
@@ -1,6 +1,6 @@
 name: 🐞 Bug report
 description: Report a very clearly broken issue.
-title: 'bug: <title>'
+title: 'bug(scope): <title>'
 labels: [bug]
```
oSumAtrIX, line 4: Most people opening issues get confused by the title syntax already as is. Do you think it's a good idea to add this?

AnonymousWP: In that case I think using `(patch)` fits better, as mostly, bugs exist in patches. In fact, the same theory is applied to the Patch Request template where its default is `(patch)` as well.

oSumAtrIX: usually it's feat(<patch app>): add <patch name>. For bugs I still think using "bug:" soley would be better than what is done currently as some people put their title in the <> placeholder brackets

AnonymousWP: I get your point, but the same would have happened to the Patch Request issue template then, yeah? Perhaps trying `bug(patch): <title>` is an idea. If you notice people indeed don't comply, you can undo it again.

oSumAtrIX: > but the same would have happened to the Patch Request issue template then, yeah

Yeah, it would be "feat:".

> Perhaps trying `bug(patch): <title>`

This will cause the following titles:

-  bug(hide ads in app): <hide ads in app>
- bug(patch): <hide ads>
- bug(hide ads in app): <title>
- bug(patch): hide ads

Instead, if we go with `bug:` the most likely title will be `bug: hide ads for app` which is correct.

AnonymousWP: > Yeah, it would be "feat:".

Sooo, if I understand correctly, my PR should also remove the scope (`(patch)`) for consistency. If you want it to be generic, I'll push a change that makes it effectively

- `bug: `
- `feat: `

Agreed?

oSumAtrIX: That sounds good

oSumAtrIX (approved):

PR #2252, opened 2023-05-25
feat(youtube/hide-seekbar): more fine grained hiding of seekbar

LisoUseInAIKyrios:
Changes:
- added hide video thumbnail seekbar
- moved seekbar settings into ReVanced -> Layout -> Seekbar

seekbar

[integration changes](https://github.com/revanced/revanced-integrations/pull/409)


oSumAtrIX:
A seekbar PreferenceScreen should be added so that both seekbar settings can be moved into it.

oSumAtrIX:
The new switch looks arbitrary to me. It shouldn't exist for simplicity reasons. The user should simply be able to chose any color to set the seekbar color.

oSumAtrIX:
Second point is too nieche to give up on the simplicity of the patch. Enabling and disabling is reasonable, but it is a neche case, where it is colored only when clicked. YouTube has changed to this behaviour recently, though I don't see anyone who really wants it. Coloring the seekbar to any color, or disabling this feature should suffice, what do you think?

oSumAtrIX:
If it is decided to keep the option. Remove on/off toggle and this one, instead use a list preference with the items:

- Off
- Always color
- Color on focus

oSumAtrIX:
"Color on player focus only"

oSumAtrIX:
A custom event handler onItemSelected can be added. Either that, or simply don't disable the settings. Inconsistent behavior towards switches thought, but doesn't cause any issues. Imo, it is more annoying:

If you want to change seekbar color while it is disabled you have to enable the patch, change the color and disable it again.

oSumAtrIX:
Sounds good, I'll leave a review in timely manner later this day.

src/main/kotlin/app/revanced/patches/shared/settings/preference/impl/PreferenceScreen.kt
```diff
@@ -26,4 +26,15 @@ internal open class PreferenceScreen(
...
+    fun addPreferences(vararg preferencesToAdd: BasePreference) {
+        val list: MutableList<BasePreference>
+        if (preferences is MutableList) {
+            list = preferences as MutableList<BasePreference>
+        } else {
+            list = preferences.toMutableList()
+            preferences = list
+        }
+        list.addAll(preferencesToAdd)
+    }
```
oSumAtrIX, line 39: That's an exciting solution, but I don't think it's pretty because this method arbitrarily exists in `PreferenceScreen` solely to solve the issue in this PR. `PreferenceScreen` has a property `preferences` where you can add preferences already. Currently, ReVanced has a different way of solving the problem that is being solved with this method. Create an intermediary "Seekbar" patch and make it implement "Closeable." Add an "addPreference" function to its companion object. The function collects added preferences. Now depend on the "Seekbar" patch everywhere where necessary. When all patches depend on it, the close function implemented from "Closeable" will be called. You can access the list of preferences previously added through the "addPreference" function and add them to SettingsPreference.PreferenceScreen.LAYOUT.

src/main/kotlin/app/revanced/patches/youtube/layout/seekbar/bytecode/patch/SeekbarColorBytecodePatch.kt
```diff
@@ -47,7 +47,7 @@ class SeekbarColorBytecodePatch : BytecodePatch(
             addInstructions(
                 registerIndex + 1,
                 """
-                        invoke-static { v$colorRegister }, $INTEGRATIONS_CLASS_DESCRIPTOR->getSeekbarColorValue(I)I
+                        invoke-static { v$colorRegister }, $INTEGRATIONS_CLASS_DESCRIPTOR->getVideoPlayerSeekbarColorOverride(I)I
```
oSumAtrIX, line 50: Are we sure to use "Override" as a suffix? Currently, this is not conventionally populated to all similar methods from other patches.

src/main/kotlin/app/revanced/patches/youtube/layout/theme/resource/ThemeResourcePatch.kt
```diff
@@ -18,16 +19,13 @@ import org.w3c.dom.Element
 @DependsOn([SettingsPatch::class, ResourceMappingPatch::class])
 class ThemeResourcePatch : ResourcePatch {
     override fun execute(context: ResourceContext): PatchResult {
-        SettingsPatch.PreferenceScreen.LAYOUT.addPreferences(
+        SeekbarColorResourcePatch.addSeekbarPreferences(
```
oSumAtrIX, line 22: `ThemeResourcePatch` does not depend on `SeekbarColorResourcePatch`. For that reason, `close` will never be called, unless another patch depends on it and is included. Add a dependency to `SeekbarColorResourcePatch` for patch `ThemeResourcePatch`.

src/main/kotlin/app/revanced/patches/youtube/layout/hide/seekbar/patch/HideSeekbarPatch.kt
```diff
@@ -34,12 +35,18 @@ class HideSeekbarPatch : BytecodePatch(
     listOf(SeekbarFingerprint)
 ) {
     override fun execute(context: BytecodeContext): PatchResult {
-        SettingsPatch.PreferenceScreen.LAYOUT.addPreferences(
+        SeekbarColorResourcePatch.addSeekbarPreferences(
```
oSumAtrIX, line 38: I don't remember well, but to me, it sounds weird that `HideSeekbarPatch` depends on `SeekbarColorBytecodePatch` (and thus transitively on `SeekbarColorResourcePatch`).  `HideSeekbarPatch` should not depend on `SeekbarColorBytecodePatch` because one hides the seek bar, and the other colors it, therefor they are separate patches. If you want a "general" seek bar patch presented for the user, it can depend on the separate patches, such as hiding or coloring it. But the coloring or hiding patches should not depend on each other structurally. 

You tried to solve the problem of including the preferences for hiding and coloring under the seek bar preference screen, but I suggest creating a separate closeable patch (without @Patch) called `SeebarPreferencesPatch` as an example. Then, all other seek bar patches can depend on it and call `SeebarPreferencesPatch.addSeekbarPreferences`. Because the patch would be closeable and patches depend on it, ReVanced Patcher will call `SeebarPreferencesPatch.close` and write the preferences after all patches that depend on its finish executing.

The reason for this change is to properly separate the patches from each other because currently, they depend on each other when not to, semantically. 

If you need help or have a question, please let me know.

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/youtube/layout/hide/seekbar/patch/HideSeekbarPatch.kt
```diff
@@ -17,14 +17,16 @@ import app.revanced.patches.shared.settings.preference.impl.StringResource
...
+import app.revanced.patches.youtube.layout.seekbar.resource.SeekbarPreferencesPatch
 import app.revanced.patches.youtube.misc.integrations.patch.IntegrationsPatch
 import app.revanced.patches.youtube.misc.settings.bytecode.patch.SettingsPatch
 
 @Patch
 @DependsOn([
     IntegrationsPatch::class,
     SettingsPatch::class,
-    SeekbarColorBytecodePatch::class // Used to hide the seekbar in the feed and watch history
+    SeekbarColorBytecodePatch::class, // Used to hide the seekbar in the feed and watch history
```
oSumAtrIX, line 28: If `hide-seekbar` depends on this patch, will the seek bar be colored red by default? If so, including `hide-seekbar` patch also causes the unintentional side effect of affecting the color of the seek bar when not desired.

oSumAtrIX: I have a hard time inferring this. How come the `theme` patch is responsible for enabling coloring? I thought, that `SeekbarColorBytecodePatch` injects hooks, that color the seekbar by default.

oSumAtrIX: > But then again, the only thing SeekbarColorBytecodePatch does is inject calls to change the seekbar color. So maybe it's name is correct.

Which method is the hook calling? If it injects calls, don't these calls cause the seek bar that is responsible to change the color? So if those methods set the color to red by default (which is iirc what the seek bar color patch does) doesn't that mean that the `SeekbarColorBytecodePatch` patch would change the color?

src/main/kotlin/app/revanced/patches/youtube/layout/hide/seekbar/patch/HideSeekbarPatch.kt
```diff
@@ -25,8 +25,10 @@ import app.revanced.patches.youtube.misc.settings.bytecode.patch.SettingsPatch
 @DependsOn([
     IntegrationsPatch::class,
     SettingsPatch::class,
-    SeekbarColorBytecodePatch::class, // Used to hide the seekbar in the feed and watch history
-    SeekbarPreferencesPatch::class    // Used to add preferences to seekbar settings menu
+    // Used to hide the seekbar in the feed and watch history by setting the seekbar color to transparent.
+    SeekbarColorBytecodePatch::class,
```
oSumAtrIX, line 29: Is transparent the default color it patches the seek bar to?

src/main/kotlin/app/revanced/patches/youtube/layout/hide/seekbar/patch/HideSeekbarPatch.kt
oSumAtrIX: > if the custom seekbar color settings controls are available

How is this determined?

oSumAtrIX: I think I understand, so basically, `SeekbarColorBytecodePatch` does not change behavior if included because, by default, it does not color the seek bar (completely red) as previously assumed but instead follows YouTube's default behavior of coloring only when the player is focused.

src/main/kotlin/app/revanced/patches/youtube/layout/hide/seekbar/patch/HideSeekbarPatch.kt
```diff
@@ -17,14 +17,18 @@ import app.revanced.patches.shared.settings.preference.impl.StringResource
...
 import app.revanced.patches.youtube.misc.integrations.patch.IntegrationsPatch
 import app.revanced.patches.youtube.misc.settings.bytecode.patch.SettingsPatch
 
 @Patch
 @DependsOn([
     IntegrationsPatch::class,
     SettingsPatch::class,
-    SeekbarColorBytecodePatch::class // Used to hide the seekbar in the feed and watch history
+    // Used to hide the seekbar in the feed and watch history by setting the seekbar color to transparent.
+    SeekbarColorBytecodePatch::class,
```
oSumAtrIX, line 29:
```suggestion
    // Does not alter the behavior of the seek bar by default.
    SeekbarColorBytecodePatch::class,
```

oSumAtrIX (approved):

src/main/kotlin/app/revanced/patches/youtube/layout/hide/seekbar/patch/HideSeekbarPatch.kt
oSumAtrIX: Thanks for clarifying. When I added the patch, the default behavior of it was to change the seek bar color to red completely by default, so I assumed this was still the case and hence questioned if hide-seekbar would cause this unintended change.

PR #2258, opened 2023-05-25
feat: universal `remove-screen-capture-restriction` patch

timschneeb:
This patch disables screen/audio capture restrictions often used for DRM-related reasons. It supersedes my old patch that only worked with Spotify.

Depends on https://github.com/revanced/revanced-integrations/pull/410.


ThePBone:
Oops, just noticed I accidentally set the main branch as the target branch. I'll create a new PR.

Ushie:
In the future, you can change the target branch by doing this https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-base-branch-of-a-pull-request

ThePBone:
Thanks, good to know. I wasn't aware of that.

PR #2287, opened 2023-05-27
feat(scbeasy): add `remove-debugging-detection` patch

ssense1337:


oSumAtrIX (approved):
What is the use of this patch, and why should this patch exist for end users and be included by default?

ssense1337:
I'm using Tasker and Shizuku
They uses wireless debugging for adb privileges

src/main/kotlin/app/revanced/patches/scbeasy/detection/debugging/patch/RemoveDebuggingDetectionPatch.kt
```diff
@@ -0,0 +1,31 @@
...
+import app.revanced.patcher.data.BytecodeContext
+import app.revanced.patcher.extensions.addInstructions
+import app.revanced.patcher.patch.BytecodePatch
+import app.revanced.patcher.patch.PatchResult
+import app.revanced.patcher.patch.PatchResultSuccess
+import app.revanced.patcher.patch.annotations.Patch
+import app.revanced.patches.scbeasy.detection.debugging.annotations.RemoveDebuggingDetectionCompatibility
+import app.revanced.patches.scbeasy.detection.debugging.fingerprints.DebuggingDetectionFingerprint
+
+@Patch
```
oSumAtrIX, line 13:
```suggestion
@Patch(false)
```

Incog5:
@ssense1337 surprised to see this patch come up here, very nice. It's very annoying all the restrictions these banking apps do it's a cat and mouse game vs malware and doesn't help in the long run. 

Would be nice to also have a patch that stops the accessible check, both kplus and scb have it and since android 13 cannot use accessibility apps that don't come from the Play store. 

Screenshot_20230602_110731

PR #2347, opened 2023-06-03
feat(finanzonline): support any version

1fexd:
This patch adds support for version 3.0.1, and makes the patch more robust


oSumAtrIX (approved):
Looks good so far. Is there any reason an opcode pattern is used instead of the class and method names? The patterns can change and break compatibility. Also, do the previous versions still work?

1fexd:
> Looks good so far. Is there any reason an opcode pattern is used instead of the class and method names? The patterns can change and break compatibility.

Switched to opcodes because the classnames changed between `2.2.0` and `3.0.1` while the actual method body stayed the same. 

> Also, do the previous versions still work?

Yes

oSumAtrIX:
In this case, I think a more future-proof way would be to drop the old and, from now on, support the new version by using the new class names. The reason for this is, that the class name is unlikely to change, unlike the method bodies.

oSumAtrIX:
This would work, but there is not much benefit in supporting the past version.

1fexd:
Should I change it? In my opinion it makes more sense to keep the opcodes since these methods are not very likely to change, while the classname may change, right (because R8/proguard is used)?

oSumAtrIX (approved):

oSumAtrIX (approved):

PR #2369, opened 2023-06-06
feat(songpal): Add patches for Sony's SongPal

1fexd:
Adds two patches to remove the "Badge" sub-tab in the "Activity" tab (see attached screenshot) as well as the red badge in the bottom navbar on the "Activity" tab since they may be perceived as annoying.

screen1

### DIFF
--- a/src/main/kotlin/app/revanced/patches/songpal/badge/annotations/BadgeCompatibility.kt
+++ b/src/main/kotlin/app/revanced/patches/songpal/badge/annotations/BadgeCompatibility.kt
@@ -1,0 +1,8 @@
+package app.revanced.patches.songpal.badge.annotations
+
+import app.revanced.patcher.annotation.Compatibility
+import app.revanced.patcher.annotation.Package
+
+@Compatibility([Package("com.sony.songpal.mdr")])
+@Target(AnnotationTarget.CLASS)
+internal annotation class BadgeCompatibility
--- a/src/main/kotlin/app/revanced/patches/songpal/badge/fingerprints/CreateTabsFingerprint.kt
+++ b/src/main/kotlin/app/revanced/patches/songpal/badge/fingerprints/CreateTabsFingerprint.kt
@@ -1,0 +1,31 @@
+package app.revanced.patches.songpal.badge.fingerprints
+
+import app.revanced.patcher.fingerprint.method.annotation.FuzzyPatternScanMethod
+import app.revanced.patcher.fingerprint.method.impl.MethodFingerprint
+import org.jf.dexlib2.AccessFlags
+import org.jf.dexlib2.Opcode
+
+// Located @ ub.i0.h#p (9.5.0)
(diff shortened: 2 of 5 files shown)


oSumAtrIX (commented):
The fingerprints may break on future versions due to their opcode array sizes. Is this considered?

oSumAtrIX:
Pushed some changes, please review if it works correctly.

1fexd:
> Pushed some changes, please review if it works correctly.

Seems to work fine, thanks

oSumAtrIX (approved):

1fexd:
> The fingerprints may break on future versions due to their opcode array sizes. Is this considered?

I don't quite get what you mean by that?

oSumAtrIX:
The opcode patterns are large, so if the dev changes the implementation of these methods, the pattern might break and the fingerprint won't resolve anymore.

1fexd:
That's true yes - does ReVanced support only partial matching (if so, I didn't know that, I just assumed I have to put all opcodes of the method into the list to make it work)?

The method the badge tab fingerprint matches looks like this:
Screenshot from 2023-06-10 18-57-54

I don't think this is going to change, if Sony decides to add a new tab they are probably going to just extend the Tab enum; The Tab enum is also used in other places, therefore I don't think it will be removed in the future.

The other fringerprint matches this fingerprint:
Screenshot from 2023-06-10 18-58-13

Here I am not so sure if it is going to stay like that - is it possible to check the type/return type an instruction is called on?

oSumAtrIX:
You can use partial opcode patterns; it only has to match. 

> is it possible to check the type/return type an instruction is called on?

I don't understand what you mean. An instruction can not be called on.

1fexd:
What I mean is if it is possible, to only match invoke-virtual's which call the method `f` on an instance of `com/google/android/material/bottomnavigation/BottomNavigationView` like the third instruction does:
Screenshot from 2023-06-10 19-08-16

oSumAtrIX:
You can utilize the `customFingerprint` property of a `MethodFingerprint`. There you can check if the list contains a `ReferenceInstruction` which reference is a `MethodReference` and check the references method name.

1fexd:
> You can utilize the `customFingerprint` property of a `MethodFingerprint`. There you can check if the list contains a `ReferenceInstruction` which reference is a `MethodReference` and check the references method name.

Oh ok, thanks a lot, I will update the patch.

1fexd:
> The fingerprint can specify the full return type (Ljava/util/List;) instead of just "L"

You can? I was under the impression that this is not possible and only the actual return type like L, V, etc. is supported

> Can add a custom fingerprint and match the method name (run)

I am trying to avoid that since I think this approach is very fragile (however, in this case it probably fine since `run` is an overriden method from `Runnable`)

oSumAtrIX:
> You can?

Yeah this was a recent change, docs should probably reflect this

PR #2370, opened 2023-06-06
feat: Fallback to local.properties if needed

1fexd:
While `gradle.properties` is listed in `.gitignore`, it is still checked in to the repository since it contains the project's version. 

This may be problematic since `gradle.properties` can also be used to store the Github token required by ReVanced (one could argue that accidentally committing the Github token can be avoided by carefully choosing the files/lines that should be included in the commit, but mistakes happen and I think it's better to play it safe). Therefore, I have added a fallback that looks for the Github user/token in `local.properties` if it is not found in `gradle.properties`.

In my opinion, support for storing the token in `gradle.properties` should be removed altogether, but that is a decision for the ReVanced team to make.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -122,5 +122,6 @@ gradle-app.setting
 # Dependency directories
 node_modules/
 
-# gradle properties, due to Github token
+# gradle properties and local.properties, due to Github token
 ./gradle.properties
+./local.properties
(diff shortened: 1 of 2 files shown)


oSumAtrIX:
Tokens should be added to the local users gradle properties file, not the local repository

1fexd:
Ohhhhhhh, that makes a lot of sense lmao. I guess I can close this PR then

PR #2384, opened 2023-06-09
feat: remove citra patches

Aunali321:
Citra is open-source project and available at https://github.com/citra-emu/citra.


oSumAtrIX (approved):

Quetzalcoatless:
I disagree with removing the Citra patch. While Citra is open-source, the non-Play builds provided on their GitHub and website still have billing code, just with no way to unlock premium due to there being no billing provider. A user would have to modify and compile the source themselves to remove billing and unlock dark mode.

I also think having patches for an open-source program in the repo is a good thing as it gives any future documentation an easy example to point to of how patches hook into existing methods.

oSumAtrIX:
ReVanced leverages the nature of open source by providing open source patches for apps that aren't and builds the effective bridge to modify apps that are not open source. If the app is open source, this bridge is not necessary. The user or someone else is encouraged to fork the application as they are supposed to and provide the modifications that way.

PR #2385, opened 2023-06-10
feat: remove patches for 'com.wakdev.apps.nfctools.se'

TheAabedKhan:
Removed 'com.wakdev.apps.nfctools.se' as there is no apk with that package name.


reisxd:
Please do not PR to the `patches.json` file, because it's automatically generated. Also, it does [exist](https://www.wakdev.com/en/apps/nfc-tools-se-android.html).

TheAabedKhan:
Oh! My bad.  I thought the patches were just for the apps available on the Google Play Store.

Thank you so much for letting me know.

Issue #189, opened 2022-07-17
Swipe control in wall panel issue


Singh345:

TheJeterLP:
Please open a proper issue.

PR #2391, opened 2023-06-11
New utility methods

1fexd:
As discussed over at [#189 @ revanced-patcher](https://github.com/revanced/revanced-patcher/pull/189), I have moved some extension methods here. 

Additionally, this adds a `MethodWithReferenceInstructionFingerprint` wrapper to even further simplify searching for methods with a certain instruction (in this case only `ReferenceInstruction`). Not sure if this level of abstraction is something you want to have in ReVanced though.


src/main/kotlin/app/revanced/util/fingerprint/MethodWithReferenceInstructionFingerprint.kt
```diff
@@ -0,0 +1,30 @@
...
+
+abstract class MethodWithReferenceInstructionFingerprint(
+    internal val returnType: String? = null,
+    internal val accessFlags: Int? = null,
+    internal val parameters: Iterable<String>? = null,
+    internal val opcodes: Iterable<Opcode?>? = null,
+    internal val strings: Iterable<String>? = null,
+    instructionOpcode: Opcode,
+    methodReference: MethodReference,
+    matchSettings: EnumSet<InstructionExtensions.MethodReferenceMatch> = InstructionExtensions.MethodReferenceMatch.all
```
oSumAtrIX, line 20: Any reason these are not internal?

src/main/kotlin/app/revanced/util/fingerprint/FingerprintUtils.kt
```diff
@@ -0,0 +1,21 @@
+package app.revanced.util.fingerprint
+
+import app.revanced.patcher.fingerprint.method.impl.CustomFingerprint
+import org.jf.dexlib2.Opcode
+import org.jf.dexlib2.iface.instruction.Instruction
+
+object FingerprintUtils {
+    inline fun <reified T : Instruction> hasInstruction(
```
oSumAtrIX, line 8: Following the naming convention:

```suggestion
    inline fun <reified T : Instruction> any(
```

src/main/kotlin/app/revanced/util/fingerprint/FingerprintUtils.kt
```diff
@@ -0,0 +1,21 @@
+package app.revanced.util.fingerprint
+
+import app.revanced.patcher.fingerprint.method.impl.CustomFingerprint
```
oSumAtrIX, line 3: This reference does not exist; merging this PR will cause a build error. Either don't use the typealias here yet, or push a commit to bump the version of ReVanced Patcher once the typealias is merged.

oSumAtrIX (changes requested):

oSumAtrIX:
This PR only adds extension functions but does not add usage for them. Why are they required?

src/main/kotlin/app/revanced/util/fingerprint/MethodWithReferenceInstructionFingerprint.kt
1fexd: Parameters can only be `internal` if they are also marked as `val`, but we don't need them as value because they are never used outside of the superclass invokation; But if you want I can make them `internal val`

1fexd:
Added an example https://github.com/revanced/revanced-patches/pull/2391/commits/c5ea054b41ff250644c54950a967939e3577ac85

src/main/kotlin/app/revanced/util/fingerprint/FingerprintUtils.kt
```diff
@@ -0,0 +1,23 @@
...
+
+import org.jf.dexlib2.Opcode
+import org.jf.dexlib2.iface.ClassDef
+import org.jf.dexlib2.iface.Method
+import org.jf.dexlib2.iface.instruction.Instruction
+
+object FingerprintUtils {
+    // TODO: Use CustomFingerprint typealias once https://github.com/revanced/revanced-patcher/pull/189 is merged
+    inline fun <reified T : Instruction> any(
+        opcode: Opcode,
```
oSumAtrIX, line 11: Any reason why this is required?

1fexd: For example, both `invoke virtual` and `invoke static` can be a method reference instruction.

oSumAtrIX: You could use `reified T : ReferenceInstruction`

1fexd: Yes but then the method wouldn't work for other `Instruction`s - shouldn't this be generic? I just used method reference instructions because I had a usecase for them (and I don't really know much about the other instructions), but ideally, there would be an extension method for each instruction in `InstructionExtensions` (and a dedicated fingerprint class for each instruction)

oSumAtrIX:
Hey, just letting you know that we have renamed repositories. Please close this PR and re-open it [here](https://github.com/ReVanced/revanced-patches/pulls). PS: No additional changes are needed to your remotes.

PR #2402, opened 2023-06-12
fix(syncforreddit/change-oauth-client-id): trim read file contents

PalmDevs:
Fixes lexer errors if read file contents has a new-line.


PalmDevs:
I've never used Kotlin before so if I do something wrong, please tell me 😅.

oSumAtrIX (approved):

PR #2434, opened 2023-06-15
feat(boostforreddit): add `change-oauth-client-id` patch

Aunali321:


oSumAtrIX (changes requested):
Instead of duplicating the patch for Sync, you should create an abstract patch for both. The abstract patch should display the same errors and execute the same patch. You could use separate fingerprints for Sync and Boost and pass them to the abstract class. There are other patches, such as the `settings` patch, you can take an example of.

src/main/kotlin/app/revanced/patches/reddit/customclients/AbstractChangeOAuthClientIdPatch.kt
```diff
@@ -0,0 +1,57 @@
...
+            File(Environment.getExternalStorageDirectory(), "reddit_client_id_revanced.txt").also {
+                if (it.exists()) return@also
+
+                val error = """
+                    In order to use this patch, you need to provide a client ID.
+                    You can do this by creating a file at ${it.absolutePath} with the client ID as its content.
+                    Alternatively, you can provide the client ID using patch options.
+                    
+                    You can get your client ID from https://www.reddit.com/prefs/apps.
+                    The application type has to be "installed app" and the redirect URI has to be set to "$redirectUri""
```
oSumAtrIX, line 33:
```suggestion
                    The application type has to be "installed app" and the redirect URI has to be set to "$redirectUri"
```

oSumAtrIX (changes requested):

Issue #2424, opened 2023-06-14
feat: Reddit is Fun patch


mtswilnd3865:
### Application

Reddit is Fun

### Issue

The current Revanced patcher does not include Reddit is Fun in the list of supported apps. 

### Patch

Is it possible to implement Reddit is Fun support in the same way that Revanced did for Sync for Reddit?

### Motivation

The looming API changes for third-party Reddit apps has basically killed off third-party apps and Reddit is Fun, along with Apollo, has announced its intention to shut down by the end of the month. By providing a revanced Sync for Reddit, it should be possible for other third-party Reddit apps to be included in Revanced in order to ensure they can still be used beyond the API deadline. 

### Acknowledgements

- [X] I have searched the existing issues and this is a new and no duplicate or related to another open issue.
- [ ] I have written a short but informative title.
- [x] I filled out all of the requested information in this issue properly.

oSumAtrIX:
Any specific reason why it should be added when Sync is already present?

snvoid:
Not OP, but Its down to preference. Some of us prefer RIF because its what we have used since it first came out and it works exceptionally well.

profucius:
> Any specific reason why it should be added when Sync is already present?

Sync is decently customizable, but pales in comparison to Boost for Reddit, which is what I prefer.

Revanced team will hopefully make patches for each 3rd party app, not only for the sake of the users who prefer their own, but also for diversity against Reddit busting one app and then everyone breaks.

Diversity, diversity, diversity.

Aunali321:
> Revanced team will hopefully make patches for each 3rd party Reddit app, or at least prioritize the big 4 (Sync, Boost, RIF, Infinity).


Why waste time making the same thing for each app? Reverse engineering and making a patch takes time. The efforts will increase significantly to support other apps. You can't just copy paste patch. Unless there is a reason to besides user's preference. It better to ask the app developers to support changing api key.

profucius:
I'll admit that I'm not aware of it being a time-consuming process to patch/replace an API key within an app. It seems that it would be more of a minor thing to change. If it is time consuming, I can understand not patching every single Reddit app without the demand for it.

However, what I can guarantee is that Boost certainly has more customization options than Sync does. It's difficult to prove that to you unless you download it yourself to look. Sync has a few theme/color choices and a couple font size choices, but by-and-large you're stuck with one of the main formats (e.g. Card, List, Compact, etc) and you can minorly tweak those aspects. 

[...]

PR #2452, opened 2023-06-18
feat(infinityforreddit): add `change-oauth-client-id` patch

johnconner122:


oSumAtrIX:
Infinity is open source, why would a patch be necessary?

johnconner122:
> Infinity is open source, why would a patch be necessary?

Not everyone is Tech Savy to change the `client-id` in source code.

johnconner122:
And also it was requested here:
https://github.com/revanced/revanced-patches/issues/2424#issuecomment-1591367220

oSumAtrIX:
> > Infinity is open source, why would a patch be necessary?
> 
> Not everyone is Tech Savy to change the `client-id` in source code.

Then create and release a fork (already exists)

oSumAtrIX:
> And also it was requested here: [#2424 (comment)](https://github.com/revanced/revanced-patches/issues/2424#issuecomment-1591367220)

The issue is for RiF, not Infinity

oSumAtrIX:
I'll keep this open until the other Reddit client PR is merged, then you need to rebase

Sculas:
ReVanced is about QoL things and/or fixes, no? So then why does it matter whether it's opensource or not? ReVanced is a trusted source for most, so they rather use that than some random fork they found online. Or am I wrong?

oSumAtrIX:
> ReVanced is about QoL things and/or fixes, no?

Yes, but there is no. benefit in doing patches, when you can also just fork an open source project.

Sculas:
ReVanced has like 80k members who don't even have a clue what a fork is, so they can't do it themselves. This means they'll have to use other people's fork. I'll repeat what I said above:
> ReVanced is a trusted source for most, so they rather use that than some random fork they found online.

Sculas:
ReVanced has reputation, a random fork has none. It would have to be verified by people who understand how it works (the 80k people I mentioned **don't**), but then how are they supposed to share that with the rest of the world in a way people trust the verifier?

Also, saying that every open source project is a trusted source is one big fat lie and you know that.

But, you seem to insist that you find this useless, so I'm going to leave it at this because this is just a waste of time for me.

oSumAtrIX:
> ReVanced has reputation, a random fork has none. It would have to be verified by people who understand how it works (the 80k people I mentioned don't), but then how are they supposed to share that with the rest of the world in a way people trust the verifier?

Point remains that that number of people trust a team of <15.  The same can properly and without any problem be done on a fork:

> Infinity has an obvious big user base and there will be people that will take care of such forks 

>Also, saying that every open source project is a trusted source is one big fat lie and you know that. 

Stop putting words in my mouth that I have not said:

> and can be verified by people.

> But, you seem to insist that you find this useless

Yes I insist into it, due to you not mentioning any point against the justification I made above for it.

johnconner122:
Developer [announced](https://www.reddit.com/r/Infinity_For_Reddit/comments/147bhsg/the_future_of_infinity/) that he might make it a paid app, it's not decoded yet. I think this patch is worth considering.

oSumAtrIX:
Paid or not, it remains open source:

image

ponces:
Much easier solution: [Google Colab Script](https://colab.research.google.com/drive/1zZ3cUJWGDjT4mtq2rohfHID9ffONhqfL?usp=sharing)

Agree with @Sculas anyway on including this patch :)

oSumAtrIX:
A proper solution is not a patch, but a fork which should also already exist, for that reason a patch is not required

ixfd64:
Would it work to use the API key from the official app?

https://news.ycombinator.com/item?id=36086240

I suspect it might not work because third-party apps don't seem to use HMAC like the official app does, but correct me if this is not the case.

johnconner122:
Do we also need to change `redirect-uri` & `user-agent`, or changing `client-id` is enough?

Turhvjbufv:
I think also change url redirect and client id as stated by the developer  https://www.reddit.com/r/Infinity_For_Reddit/comments/14c7v84/if_you_want_to_use_your_own_api_key/

mvevitsis:
You should change all 3. The fact that the sync patch doesn't change the redirect and user agent means it is only a matter of time before they start blocking it.

src/main/kotlin/app/revanced/patches/reddit/customclients/infinityforreddit/api/patch/ChangeOAuthClientIdPatch.kt
```diff
@@ -0,0 +1,47 @@
...
+import app.revanced.patcher.patch.PatchResultSuccess
+import app.revanced.patcher.patch.annotations.Patch
+import app.revanced.patches.reddit.customclients.AbstractChangeOAuthClientIdPatch
+import app.revanced.patches.reddit.customclients.infinityforreddit.api.fingerprints.GetHTTPBasicAuthHeaderFingerprint
+import app.revanced.patches.reddit.customclients.infinityforreddit.api.fingerprints.LoginActivityOnCreateFingerprint
+import org.jf.dexlib2.iface.instruction.OneRegisterInstruction
+
+@Patch
+@Name("change-oauth-client-id")
+@Description("Changes the OAuth client ID in Infinity for Reddit.")
```
oSumAtrIX, line 19:
```suggestion
@Description("Changes the OAuth client ID.")
```

mqwec43as:
How are you going to tell when they will be running those checks though? Do they make prior announcements?

oSumAtrIX:
By checking if the requests fail

jameshilliard:
> Do we also need to change `redirect-uri` & `user-agent`, or changing `client-id` is enough?

Probably a good idea to allow changing these as well, in fact the patching process could be simplified a bit for users this was as you could then avoid the requirement for them to generate an API key entirely by simply copying one from an app that's except from any API limits.

For example this is the redreader redirect uri(you can see it just by clicking "Add Reddit Account" in redreader and copying the login url you get redirected to), using it in other apps would require changing the `redirect-uri` from my understanding:

https://www.reddit.com/api/v1/authorize.compact?response_type=code&duration=permanent&state=Texas&redirect_uri=redreader%3A%2F%2Frr_oauth_redir&client_id=yH0aTnJEt6qUgGn835B4vg&scope=identity%20edit%20flair%20history%20modconfig%20modflair%20modlog%20modposts%20modwiki%20mysubreddits%20privatemessages%20read%20report%20save%20submit%20subscribe%20vote%20wikiedit%20wikiread

jameshilliard:
> If Reddit decides to place some checks for the redirect uri or user agent, then we can change them later on

FYI looks like this will probably be needed:
https://old.reddit.com/r/AfterVanced/comments/14nqz2s/we_are_beginning_to_see_reports_of_api_keys/

oSumAtrIX:
@jameshilliard If your OAuth client id gets revoked, ask Reddit why first.

jameshilliard:
> ask Reddit why

Isn't reddit's policy to just [lie](https://old.reddit.com/r/apolloapp/comments/14dkqrw/i_want_to_debunk_reddits_claims_and_talk_about/) about anything related to the API these days? I'm not sure how asking them about uses cases they aren't interested in supporting would be all that productive.

oSumAtrIX:
I don't know what they want to lie about here. Ask them about if, why and when the client id is revoked.

Issue #1752, opened 2023-03-14
bug(YouTube): Spoof app signature - various side effects


JUANHDA-CX:
### Type

Cosmetic

### Description

- [x] ~~Video is not added to the history properly~~ Fixed in latest patch
- [x] ~~Audio playback forced in feed~~ Fixed
- [x] ~~Subtitles sometimes don't appear in correct location~~ Fixed
- [x] ~~Playback issues with some videos: home feed autoplay, playlists, opening videos from push notifications, opening some external video links~~ Fixed 
- [x] ~~Endscreen cards are always hidden~~ Fixed
- [x] ~~Built in YouTube download feature does not work for long videos or downloading multiple videos at once~~ Fixed
- [x] ~~Download button is hidden in old layout~~ Fixed
- [x] ~~Ambient mode does not work~~ Fixed
- [x] ~~Shared clips do not playback from the clipped time region~~ Fixed
- [x] ~~Autoplayed videos from feed are added to history~~ Fixed
- [x] ~~Seekbar preview thumbnails are in low quality~~ Fixed
- [x] ~~Home feed autoplay does not show close caption text~~ Fixed
- [x] ~~Zoom to fill sometimes does not work~~ Fixed

oSumAtrIX:
Try with stock YouTube.

joyskmathew:
It was removed from the stock app also

But if somehow, revanced team can reproduce it, it would be great

SodaWithoutSparkles:
Cannot reproduce, I still have it in revanced. Maybe google is doing A/B testing?

![Screenshot_app.revanced.android.youtube-edit.jpg](https://user-images.githubusercontent.com/64138578/225214815-9a7b7b01-62f3-405f-ae1b-340d419afe7e.jpg)

oSumAtrIX:
No, if you include microg-support patch, the spoof-signature-verification also gets included.

kazimmt:
This issue arrived after this spoof-signature-verification patch
inotia00 also discussed about it on his group. He add a settings to toggle it on/off

oSumAtrIX:
Not really a good solution, ideally the patch works around the playback as cleanly as possible

kazimmt:
It works to fix buffer issue, 
But see already 3 issue arrived after this patch

Ya it's not a good solution. But maybe temporary

Settings for this patch, Refer to: 
https://github.com/inotia00/revanced-patches/commit/18747410769cd7723616d2155474e06850305024
[...]

PR #2476, opened 2023-06-23
fix(youtube/client-spoof): update settings text for known side effects of spoof signature

LisoUseInAIKyrios:
- Updated settings summary to reflect the latest patch changes.
- Turned off the subtitle fix (not needed for the current spoof)

[integration changes](https://github.com/revanced/revanced-integrations/pull/424)


inotia00:
Spoofing with `CgIQBg==` parameter resolves issue where end screen cards were always hidden
(Of course, you have to manually change `Hide end screen cards` to off in Layout settings)

Also, the issue where the download button was hidden in the old layout or tablet layout is also resolved
(But I can't guarantee that offline downloads will work either. Someone test it and let me know the results)

Instead, there will be side effects such as:
- [Ambient mode](https://support.google.com/youtube/answer/12827017) may not work
- Seekbar thumbnails are hidden

src/main/kotlin/app/revanced/patches/youtube/misc/fix/playback/patch/SpoofSignatureVerificationPatch.kt
```diff
@@ -96,6 +97,7 @@ class SpoofSignatureVerificationPatch : BytecodePatch(
         } ?: return OpenCronetDataSourceFingerprint.toErrorResult()
 
         // hook override subtitles
+        if (false) { // Subtitle fix is not required for current protobuf parameter spoofing
```
oSumAtrIX, line 100: The following code should be removed, including the integrations. Git saves this in history if we need this anytime again.

oSumAtrIX (approved):

oSumAtrIX:
> Spoofing with CgIQBg== parameter resolves issue where end screen cards were always hidden

This should be reflected in #1752

PR #2485, opened 2023-06-24
feat(ytmusic): shorten the app name more

katyusha128:
Some launchers only show 1 line for app label which cut the app label to 'YT Musi...'


oSumAtrIX:
This is an issue with your launcher. The label has to change otherwise it's not distinguishable between the original app

PR #2508, opened 2023-06-27
fix(relayforreddit/change-oauth-client-id): change client id instead of developer key

gudmunduro:
Changes the patch to change the correct client id.

I somehow naively thought that `DeveloperKey` was the client id, but that's actually just for YouTube.
I tested the app better now to make sure it was actually using my client id.

Also fixes the patching error on the `@ChangeOAuthClientIdPatchAnnotation` by moving `@Patch` out of it.

### DIFF
--- a/src/main/kotlin/app/revanced/patches/reddit/customclients/relayforreddit/api/fingerprints/GetLoggedOutBearerTokenFingerprint.kt
+++ b/src/main/kotlin/app/revanced/patches/reddit/customclients/relayforreddit/api/fingerprints/GetLoggedOutBearerTokenFingerprint.kt
@@ -1,0 +1,3 @@
+package app.revanced.patches.reddit.customclients.relayforreddit.api.fingerprints
+
+object GetLoggedOutBearerTokenFingerprint : AbstractClientIdFingerprint("https://oauth.reddit.com/grants/installed_client")
--- a/src/main/kotlin/app/revanced/patches/reddit/customclients/relayforreddit/api/fingerprints/AbstractClientIdFingerprint.kt
+++ b/src/main/kotlin/app/revanced/patches/reddit/customclients/relayforreddit/api/fingerprints/AbstractClientIdFingerprint.kt
@@ -2,6 +2,6 @@ package app.revanced.patches.reddit.customclients.relayforreddit.api.fingerprint
 
 import app.revanced.patcher.fingerprint.method.impl.MethodFingerprint
 
-object GetClientIdFingerprint : MethodFingerprint(
-    strings = listOf("AIzaSyCTQfRx9fHnDpfcfiI5pmwyGUBjDVTNvX8")
+abstract class AbstractClientIdFingerprint(string: String) : MethodFingerprint(
+    strings = listOfNotNull("dj-xCIZQYiLbEg", string),
 )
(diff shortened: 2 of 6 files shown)


src/main/kotlin/app/revanced/patches/reddit/customclients/boostforreddit/api/patch/ChangeOAuthClientIdPatch.kt
```diff
@@ -1,35 +1,31 @@
...
 import app.revanced.patcher.extensions.InstructionExtensions.addInstructions
 import app.revanced.patcher.fingerprint.method.impl.MethodFingerprintResult
 import app.revanced.patcher.patch.PatchResult
 import app.revanced.patcher.patch.PatchResultSuccess
 import app.revanced.patcher.patch.annotations.Patch
 import app.revanced.patches.reddit.customclients.AbstractChangeOAuthClientIdPatch
+import app.revanced.patches.reddit.customclients.ChangeOAuthClientIdPatchAnnotation
 import app.revanced.patches.reddit.customclients.boostforreddit.api.fingerprints.GetClientIdFingerprint
 
 @Patch
```
oSumAtrIX, line 15: I think moving @Patch out of it shouldn't be necessary. Instead this issue should be fixed in ReVanced Patcher.

src/main/kotlin/app/revanced/patches/reddit/customclients/relayforreddit/api/fingerprints/GetLoggedOutBearerTokenFingerprint.kt
```diff
@@ -0,0 +1,7 @@
+package app.revanced.patches.reddit.customclients.relayforreddit.api.fingerprints
+
+object GetLoggedOutBearerTokenFingerprint :
+    AbstractClientIdFingerprint(
+        "RedditAccountManager;",
+        "https://oauth.reddit.com/grants/installed_client"
```
oSumAtrIX, line 6: Are really both strings required?

src/main/kotlin/app/revanced/patches/reddit/customclients/boostforreddit/api/patch/ChangeOAuthClientIdPatch.kt
gudmunduro: Yeah I guess that would be a better fix

src/main/kotlin/app/revanced/patches/reddit/customclients/relayforreddit/api/fingerprints/GetLoggedOutBearerTokenFingerprint.kt
gudmunduro: Yes, one is the name of the class, the other is a string from the method, and there are multiple methods in the same class that need to be patched

gudmunduro: I mean it's likley possible to skip the class name, since it's unlikley that any other method contains https://oauth.reddit.com/grants/installed_client. But I still think it makes more sense to keep it, so that it is consistent with the the other fingerprints (which most likely need both strings)

oSumAtrIX (approved):

PR #2571, opened 2023-07-02
feat(slideforreddit): add `change-oauth-client-id` patch

vlemazza:
This patch, like others for the various reddit clients, replace the developer's client id with a personal one.


patches.json
oSumAtrIX, line 1: Please do not commit to this file, it is autogenerated

src/main/kotlin/app/revanced/patches/reddit/customclients/slide/api/fingerprints/GetClientIdFingerprint.kt
```diff
@@ -0,0 +1,11 @@
+package app.revanced.patches.reddit.customclients.slide.api.fingerprints
+
+import app.revanced.patcher.fingerprint.method.impl.MethodFingerprint
+
+object GetClientIdFingerprint : MethodFingerprint(
+    customFingerprint = custom@{ methodDef, classDef ->
+        if (!classDef.type.endsWith("Credentials;")) return@custom false
+
+        methodDef.name == "getClientId"
```
oSumAtrIX, line 9: Likely this is the only method that has this name. Is it necessary to check for the classes type?

oSumAtrIX (approved):
Resolve pending comments, rest lgtm

vlemazza:
a conflict in the file patches.json needs to be resolved, what is missing ?

oSumAtrIX:
Simply do not commit to that file

src/main/kotlin/app/revanced/patches/reddit/customclients/slide/api/fingerprints/GetClientIdFingerprint.kt
oSumAtrIX: This is less readable than before

oSumAtrIX: Consider the following example:

```
if (!a) return false;
if (b) return false;

return c;
```

This example is easy to understand and follow. Now try to convert it to a single statement. Already the conversion requires a certain amount of thought to process. The reverse will be required when trying to understand the statement.

I'd use && to prevent nesting for example, or when there is a lot of expressions which you can group together such as:

`if (a && b && c && d) return false;`

Here, I prefer to seperate the two statements, because the newline before the `&&` causes an indentation, which looks weird. Indentations are useful to group blocks of code that are semantically related such as an if statements body. Here the indentation makes less sense in this regard. Removing the newline causes the line of code to be very long which is why separating into two expressions is a good solution that is simple to follow and read.

vlemazza:
> > a conflict in the file patches.json needs to be resolved, what is missing ?
> 
> You need to revert the commit that edited patches.json
 done reverted

src/main/kotlin/app/revanced/patches/reddit/customclients/slide/api/fingerprints/GetClientIdFingerprint.kt
oSumAtrIX: > If its an issue with multiple lines it could be

Similar lines of code exceed the maximum recommended amount of characters per line, hence auto refactoring will insert the newline back again, but yeah minor problem :)

oSumAtrIX (approved):

PR #2578, opened 2023-07-04
feat(pixiv): add `hide-ads` patch

Sculas:
It's a straightforward patch and does the job perfectly. It's better than those DNS blockers since the layout element is never added meaning you don't see a weird transparent box.

As for the FIXME, it doesn't matter for now. However, if the fingerprint does break in the future, the first thing that should be done is to look at the new string value or even better, remove it and find it from somewhere else. But unless Pixiv goes on the offensive, this should work just fine.


oSumAtrIX:
Why is the client-side premium status not spoofed? While server-side features won't work, this patch would unlock client-side features:

image

Issue #2432, opened 2023-06-15
bug(youtube): videos only buffer to 1:10, client-spoofing fails


suptower:
### Type

Error at runtime

### Bug description

When watching any video, it will stop loading exactly at the 1:09 / 1:10 mark. The issue seems to be related to client-spoofing.

Additionally, spoofing the version to 16.08.35 seems to resolve the problem, as well as activating Incognito.

Theory was this is the result of A/B testing, however, according to this discord thread, it is not only a certain region impacted.
https://discord.com/channels/952946952348270622/1118066503779098684

### Steps to reproduce

1. Have an impacted account.
2. Play any video longer than 70 seconds
3. Try to pass the 1:09 / 1:10 mark. Either by letting the video play or by scrolling (scrolling past this point results in an infinite buffer loop as well)

### Relevant log output

```shell
NA
```

### Screenshots or videos

![image](https://github.com/revanced/revanced-patches/assets/70473844/2ddc54d5-7401-4919-8481-1ca50c86e3ac)

### Solution

 - Spoofing the version to 16.08.35 seems to resolve the problem.
 -  Activate incognito mode

### Additional context

https://discord.com/channels/952946952348270622/1118066503779098684

Spoof App Signature was initially enabled. Tested it with both off and on after the issue has happened, did not alter the outcome.

### Acknowledgements

- [X] I have searched the existing issues and this is a new and no duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I filled out all of the requested information in this issue properly.

oSumAtrIX:
> Spoofing the version to 16.08.35 seems to resolve the problem.

Can you try any later versions?

oSumAtrIX:
It may also be that the server reacts differently to the versions. If the server receives one version it does allow playback, if another, it doesn't.

suptower:
I've also tried it with different versions of YouTube Revanced (18.22.36 and 18.19.35). Currently running 18.22.36 with spoofed 16.08.35, which seems to work for now.

Julius8900:
May I know how you did try 18.22.36? The `client-spoof` patch is supported only on version 18.19.35.

suptower:
I've enabled experimental patches support in the settings of Revanced Manager. 

I was keen on having the new UI while simultaneously being able to use SponsorBlock, that's why.
Never ran into issues regarding that and I've been using it for over a month or so I think.

kevinbbbb:
[...]

PR #2607, opened 2023-07-09
fix(youtube/spoof-signature-verification): show video time and chapters while using seekbar

LisoUseInAIKyrios:
Partial fix for the seekbar behaving differently when spoofing.

Video time and chapters are now visible when dragging the seekbar.  

Seekbar preview thumbnails are still missing.  This might be fixable with the right patch.  I made a few patch attempts but could not force the thumbnails to generate (still not sure if this is possible at all).

I may try adding an imitation of the seekbar thumbnails, where if the seekbar is dragged then temporarily pause the video and manually call `VideoInformation.seekTo()` so the video matches wherever the seekbar time currently is.  Then after dragging is complete unpause the video.   It would not be ideal, but might work better than no seekbar thumbnails at all.  Still need to try this.

[integration changes](https://github.com/revanced/revanced-integrations/pull/435)


oSumAtrIX:
>  then temporarily pause the video

This is something that should be off by default. I personally, for example often drag while watching the video play. This will be the case for the majority.

src/main/kotlin/app/revanced/patches/youtube/misc/fix/playback/patch/SpoofSignatureVerificationPatch.kt
```diff
@@ -68,6 +56,47 @@ class SpoofSignatureVerificationPatch : BytecodePatch(
             }
         } ?: return ProtobufParameterBuilderFingerprint.toErrorResult()
 
+
+        // Force the video time and chapters to show up when using the seekbar.
```
oSumAtrIX, line 60: Document the strategy of this patch, its purpose, why it is necessary, and how it works. Someone passing by this will question why this patch exists and how it works. Take an example from [here](https://github.com/revanced/revanced-patches/blob/main/src/main/kotlin/app/revanced/patches/youtube/misc/litho/filter/patch/LithoFilterPatch.kt#L38-L72).

oSumAtrIX:
I am unsure this is an easy problem to solve. We already change the client parameters that stop requesting thumbnails from YouTube, meaning, that if we want to request thumbnails we need to spoof the parameter back to what it was before and if we do that, playback may once again be affected.

I have received reports that the old parameter was working again, it may be worth a shot trying if it still happens to work again. 

The alternative but definitive solution to all related problems is to properly spoof the clients integrity status and not workaround it by spoofing player parameters. This requires finding and patching any places that are used to calculate the integrity. Android only provides a handsful of APIs to calculate the integrity meaning that if we successfully spoof all references to these APIs, integrity can be restored. Though, a scenerio remains that prevents this from working which is the use of reflection. If the API is called reflectively, references can not be found directly.  Another scenario would be that the integrity is calculated by GMS, outside of the YouTube app, meaning that references to GMS have to be spoofed. 

So far I have tried replacing all references to the Android API that are used to fetch the signature of the app as well as the package name and spoofed them to point to the original YouTube APK, so that instead of calculating the integrity of the current app it calculates the integrity of the official app, though that did not work. Cause can be, that I missed spots, other APIs or reflection is used to calculate the integrity or integrity is calculated outside of the app.

Once we figure out a PoC patch that spoofs the integrity successfully, we can narrow down the patch to the minimum necessary patches and understand how integrity is calculated and where it is used and sent.

On another note, a different but technically working solution would be to spoof the client to iOS. iOS does not feature such extensive APIs to calculate the integrity of the app, playback issues do not happen there. If we are able to spoof YouTube so that it is an iOS app, it would not only resolve #1752 but also act as a way better work around than the current one.

oSumAtrIX:
Related to [this](https://github.com/yt-dlp/yt-dlp/commit/1e75d97db21152acc764b30a688e516f04b8a142).

oSumAtrIX (approved):

YT-Advanced:
> I found the video seeking idea did work, but it does not update very quickly unless the video quality is very low (such as 240p or 144p). The video quality could be temporarily changed to a lower quality while scrubbing, but I gave up on this idea for now as it was getting too complicated.

@LisoUseInAIKyrios I think it may possible. Can you share the code, I want to try.

YT-Advanced:
@LisoUseInAIKyrios @oSumAtrIX I think I have a good new for playback issues. As you can see in the this line we have Youtube developer key. So it mean youtube use YT Data API v3 to limit modded app. For easier understand, I will give an example how it limit.
In Youtube, when posting the request to Youtube, it send something like this:
```
GET/ https://www.googleapis.com/videos/getblabla
x-android-package: com.google.android.youtube
x-android-cert: Maybe the hash code from package name and version code
```
BUT in non-root users we must use different package name, so Google will return 4xx code

image

YT-Advanced:
For Solution, I think we have 2 solution:
1. Hook the shell_Url_Activity.getPackageName() for always return Youtube default package name
2. Use different Youtube API Key (Not recommended)

YT-Advanced:
> I think the package name is already spoofed in most places. Check what that code looks like with a patched apk.
Edit: the code is the same when patched.

So it mean the package name is not spoofed ?

oSumAtrIX:
I jave once tried spoofing the package names globally on the entire app, but I have accidentally skipped over this particular api (ShellUrlActivity). I'll give spoofing this particular a shot today, or if anyone wants to do it, simply report back in #1752.

oSumAtrIX:
Spoofing this specific location does not resolve the issue.

YT-Advanced:
> Spoofing this specific location does not resolve the issue.

What is the parameter that you spoof. The correct ones is `com.google.android.apps.youtube.app.application.Shell_UrlActivity`

oSumAtrIX:
What I have tried so far is spoof every reference to `pm.getPackageInfo(String, int)`. I checked if the first parameter matches the app or GMS and spoofed it to the correct package. There are of course, more ways to get the package name and signature, but I have not tried spoofing them.

YT-Advanced:
As I read from JADX, we are on the right way. `StandalonePlayerActivity` is a class used to perform operations such as playing videos. When start, it call `Shell_UrlActivity` to get `developer_key` (stringExtra3), `app_version` (stringExtra4), `client_library_version` (stringExtra5) and `app_package` (stringExtra6). We only need to focus on `app_package`
In OnCreate() of `StandalonePlayerActivity`, it have a check if package name match youtube original package name (you can see the picture below)
image
Also in `d`, we have a hash code func which is used for `x-android-cert`

YT-Advanced:
I think we should hook these parameter:
- `this.a = str;` in `d` file to `this.a = "com.google.android.youtube";`
- `String str6 = str2;` in `ngd` file to `String str6 = "com.google.android.youtube";`
- Hook `stringExtra6` in `StandalonePlayerActivity` to "com.google.android.youtube"

YT-Advanced:
> Try patching the method and add some logging to find out. Also add logging to the other method I mentioned above.

@johnconner122 Now I am not in PC so can test it :( But I think it called by video because of the hash code in `d` file (it used to verify the youtube signature)

YT-Advanced:
@oSumAtrIX @LisoUseInAIKyrios I have tested and look like it worked on my devices

YT-Advanced:
@LisoUseInAIKyrios I mannual patch Revanced Apks

YT-Advanced:
> Did you also change the hash function value, or did you only change the package name?

Here is my changes:
- Hook shell_Url_Activity.getPackageName() (the parameter above)
- this.a = str; in d file to this.a = "com.google.android.youtube";
- String str6 = str2; in ngd file to String str6 = "com.google.android.youtube";
- Hook stringExtra6 in StandalonePlayerActivity to "com.google.android.youtube"

YT-Advanced:
@LisoUseInAIKyrios I have updated new changes, with only 2 changes

YT-Advanced:
@LisoUseInAIKyrios New Update: Look like it didn't work like excepted, an user reported about playback issues, but it decrease the number of user have this problem

YT-Advanced:
> The idea needed a bit more work than just dynamically changing video quality, because everytime the seekbar was moved it reset loading the video preview. So the user had to move the seekbar and then stop for a half second to reliably update the preview. That could be fixed by seeking the video and then not seeking again until the preview loads the last seeked time, but it would require additional patch changes to detect that.

@LisoUseInAIKyrios I think use `VideoInfomation.Seekto` is better. And also the bug you are mentioning can easily fixed by add a func. If you can please share your changes with the way above.

YT-Advanced:
> I was using VideoInformation.Seekto. The issue is when seeking multiple times (such as, when the user is sliding the seekbar), is on every call to seekTo it resets the loading of the video. So the only way to reliably get the video to show the current seek time, is to first seek and then don't seek again until the paused video is updated with that time. There is no existing way to detect this, and an additional patch point would be needed to detect the paused video has finished loading the last requested seek time. Even if that is patched, the preview does not update very quickly unless the video quality is very low (such as 240p).

I know, so I think we can add an integrations change to check if user are scrolling. Here is an example:
- At the first, user seek to 1:02, the patches send to integrations and integrations perform the seek then save 1:02 to a variable.
- Then it send 1:04, the integration compare with the old variable (1:02) then skip it
- When user stop the seek, reset variable

oSumAtrIX:
A seekEnd event can be fired when the time stops changing after a specific interval. Every time you seek, the time changes which resets the interval, once you stop seeking, the interval would fite the seekEnd event. A delay of the length of the interval would be an issue

YT-Advanced:
> PS: A seekEnd event can be fired when the time stops changing after a specific interval. Every time you seek, the time changes which resets the interval, once you stop seeking, the interval would fite the seekEnd event. A delay of the length of the interval would be an issue

Right, but is is better than no seekbar thumbnail

> It would still be slow updating the preview unless the video quality was very low

I know it will be slow but it should be a temporary solution while finding a way to force the thumbnail create

YT-Advanced:
Nobody plan to do this ?

YT-Advanced:
> I looked thru my old comments regarding this, and remembered there also an issue with pausing/resuming playback. To do the emulated seeking preview it has to first pause and then after scrubbing it needs to resume video playback. I was able to get the video to pause, but could not find a way to reliably resume playback. Forcing it to resume would break the UI (it would think the video was still paused) or it was ugly because it needed to force the resume button to show before forcing playback to resume. And even if all this solved, its impossible to scrub the video unless playback is paused (there is absolutely no way around that limitation). With all of these limitations I gave up on this idea.

Okay, I understand.

YT-Advanced:
@LisoUseInAIKyrios As #2432, the Short paramter worked when spoof to old version. So I think spoof all `versioncode` and `version name` on `ayhv`, `wsl`, `btp`, `now` and `com/google/android/apps/youtube/embeddedplayer/service/clientinfo/service/e`

YT-Advanced:
> Try it if you think it might work.

Tested and it not worked. And can you share me the path that you have changed to perform a seekTo idea ?

Issue #2426, opened 2023-06-14
bug(tiktok): version 30.0.2 fails to patch


xscd:
### Type

Error while patching

### Bug description

tiktok bumped a major version and that one fails to patch

### Steps to reproduce

1) get the latest (30.0.2) tiktok apk
2) try to patch (java -jar revanced-cli-2.21.3-all.jar -a tiktok.apk -c -o re-tiktok.apk -b revanced-patches-2.177.1-dev.2.jar -m revanced-integrations-0.110.1-dev.2.apk)
3) see fail

### Relevant log output

```shell
INFO: Reading dex files 
INFO: Decoding AndroidManifest.xml only, because resources are not needed 
WARNING: Could not decode attr value, using undecoded value instead: ns=android, name=resource, value=0x6a020000 
INFO: Skipping change-package-name: Excluded by default 
INFO: Skipping enable-android-debugging: Excluded by default 
INFO: Skipping export-all-activities: Excluded by default 
INFO: Skipping predictive-back-gesture: Excluded by default 
INFO: Skipping remove-screen-capture-restriction: Excluded by default 
INFO: Skipping remove-screenshot-restriction: Excluded by default 
INFO: Skipping sim-spoof: Excluded by default 
INFO: Skipping spoof-wifi-connection: Excluded by default 
INFO: Merging revanced-integrations-0.110.1-dev.2.apk 
INFO: disable-login-requirement succeeded 
SEVERE: downloads failed: 
app.revanced.patcher.patch.PatchResultError: 'downloads' depends on 'settings' but the following error was raised: Hardcoded offset changed.
	at app.revanced.patcher.Patcher.executePatches$executePatch(Patcher.kt:304)
	at app.revanced.patcher.Patcher.access$executePatches$executePatch(Patcher.kt:36)
	at app.revanced.patcher.Patcher$executePatches$1.invokeSuspend(Patcher.kt:350)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlin.sequences.SequenceBuilderIterator.hasNext(SequenceBuilder.kt:129)
	at app.revanced.utils.patcher.PatcherKt.applyPatchesVerbose(Patcher.kt:84)
	at app.revanced.cli.patcher.Patcher.start$revanced_cli(Patcher.kt:19)
	at app.revanced.cli.command.MainCommand.run(MainCommand.kt:166)
[...]

PR #2656, opened 2023-07-18
fix(tiktok/settings): fix tiktok settings patch

d4rkk3y:
Instead of replacing content of an exist settings entry in tiktok settings, create a settings entry for revanced settings and insert it to tiktok settings view.
https://github.com/ReVanced/revanced-integrations/pull/440
This PR should solve #2426.


src/main/kotlin/app/revanced/patches/tiktok/misc/settings/fingerprints/SettingsEntryFingerprint.kt
```diff
@@ -0,0 +1,9 @@
+package app.revanced.patches.tiktok.misc.settings.fingerprints
+
+import app.revanced.patcher.fingerprint.method.impl.MethodFingerprint
+
+object SettingsEntryFingerprint : MethodFingerprint(
+    strings = listOf(
+        "pls pass item or extends the EventUnit"
```
oSumAtrIX, line 7: Is this an actual string?

d4rkk3y: Yes. It is an exception message.

oSumAtrIX (approved):

PR #2714, opened 2023-07-29
feat(Instagram): Add patch to disable screenshot detection

badawoll:


oSumAtrIX:
A patch that is not constrained to any package is present to disable screenshot restrictions. Any reason Instagram needs special treatment? If so, can the other patch be modified in a way that allows to remove the screenshot detection from Instagram without the patch needing special code for Instagram

badawoll:
Screenshot detection works by registering a content observer and then monitoring it for new images in the screenshots folder. Functionally very different from screenshot prevention.

I think the patch functionality can be generalized to all apps by patching/hooking all calls to the content observer API, but it would probably belong into its own patch. If the unconstrained patch makes sense, I can start working on it, and this MR can be closed.

oSumAtrIX:
> But it would probably belong in its patch

The current patch to lift screenshot prevention hooks calls to APIs to prevent screenshots. A separate patch can manage a hook for the content observer API calls and filter those that monitor for new images in the screenshots folder. It may be considerable to create a base implementation of the content observer hook in a separate patch but in a way that allows the screenshot prevention patch to implement it (detecting new screenshots) because, in terms of abstraction, the patch your PR would introduce has the same purpose as the patch for screenshot prevention. 

Depending on the complexity we can also merge this PR.

oSumAtrIX:
> Could bypassing content observation break other apps that use it for non-screenshot purposes?

Yes, I have come across the same thought. An example scenario would be that some app uses the content observer to detect new screenshots for some random purpose but uses the dedicated system API to prevent screenshots. Executing the patch on this app would break functionality. Adding patch options that give necessary control to the extent of the unified screenshot detection patch may be considerable for that reason.

> The universal patch would ignore if the IG fingerprint fails to resolve, making IG-specific modifications only if IG is the target.

If the patch is considered "universal," it would break the design to include fingerprints for specific apps. Consider the scenario above, where any sample application may break.

oSumAtrIX:
Is this PR currently worked on?

fillwithjoy1:
@badawoll seems to just fell off the platform - no activity, possibly forgot about account?

PR #2722, opened 2023-08-02
feat(YouTube Music): Add `Permanent repeat` patch

### LINKED ISSUE #1295: Permanent Shuffle & Permanent Repeat (Discontinued Vanced YouTube Music feature)
### Application

YouTube Music Revanced Extended

### Issue

Permanent Shuffle and Permanent Repeat were features added by the Vanced team for the discontinued YouTube Music Vanced.

### Patch

Permanent Shuffle and Permanent Repeat were features added by the Vanced team for the discontinued YouTube Music Vanced. 

Screenshot_20221217-221100183 (1)

You don't have to tap those icons over and over again, because it remembers with every new song/video you play even when closing the app or rebooting phone. I have YTM Vanced installed with German language so I don't know the exact English translation. But basically once taping any of those icons it would remember the next time you play a song/video in YTM Vanced.

### Motivation

Less taping those icons and therefore much more enjoyable YTM experience.

### Acknowledgements

MatthewCash:
To partially resolve ReVanced/revanced-patches-template#1576 for YouTube Music this adds permanent repeat to YouTube music. When the patch is enabled, the repeat button will stay enabled until it is manually disabled.

Because the repeat button is part of the layout, I put the patch in the `layout/` directory, but I am not sure if that is actually the best place for it.

Goes along with ReVanced/revanced-patches#2730, since both patches are nearly identical I'll apply any requested changes to both.
No integrations for this patch.

### DIFF
--- a/src/main/kotlin/app/revanced/patches/music/interaction/permanentrepeat/fingerprints/PermanentRepeatFingerprint.kt
+++ b/src/main/kotlin/app/revanced/patches/music/interaction/permanentrepeat/fingerprints/PermanentRepeatFingerprint.kt
@@ -1,0 +1,22 @@
+package app.revanced.patches.music.interaction.permanentrepeat.fingerprints
+
+import app.revanced.patcher.extensions.or
+import app.revanced.patcher.fingerprint.method.impl.MethodFingerprint
+import org.jf.dexlib2.AccessFlags
+import org.jf.dexlib2.Opcode
+
+object PermanentRepeatFingerprint : MethodFingerprint(
+    "V",
+    AccessFlags.PUBLIC or AccessFlags.FINAL,
+    listOf("L", "L"),
+    listOf(
+        Opcode.CHECK_CAST,
+        Opcode.INVOKE_INTERFACE,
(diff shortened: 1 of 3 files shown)


oSumAtrIX (changes requested):
This duplicates code quite a lot. Please abstract in a way that does not duplicate it. Also, a patch for settings should be added in another pull request. For the scope of this PR, adding the patch that enables the patches permanently suffices.

oSumAtrIX:
Please also separate this PR into two, introducing each patch.

oSumAtrIX:
> When the patch is enabled, the repeat button will stay enabled until it is manually disabled.

Isn't it okay to include the patch by default because you can enable or disable it?

MatthewCash:
> > When the patch is enabled, the repeat button will stay enabled until it is manually disabled.
> 
> Isn't it okay to include the patch by default because you can enable or disable it?

The default behavior is for repeat to be automatically disabled when a new song is selected. This patch prevents that behavior. I think it should be opt-in since some users may not prefer having to manually disable song repeating, but I don't think its a big deal either way.

Issue #2727, opened 2023-08-02
feat: reddit is fun user agent change 


DaviddH:
### Application

Reddit is fun

### Issue

Reddit is fun is now blocked based on the user agent so it seems, just using oauth client id change patch is not enough anymore 

### Patch

Patch to change the user agent reddit is fun is using, if possible mirroring something like a recent chrome version so it can't be detected by redddit

### Motivation

Patching this would make the app work again and not force users to use the terrible first party app

### Acknowledgements

- [X] I have searched the existing issues and this is a new and no duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I filled out all of the requested information in this issue properly.

oSumAtrIX:
The official client's user agent could be tried.

Ma-ve:
The info Reddit shows:

>>> whoa there, pardner!
reddit's awesome and all, but you may have a bit of a problem.

Make sure your User-Agent is not empty, is something unique and descriptive and try again. if you're supplying an alternate User-Agent string, try changing back to default as that can sometimes result in a block.

You can read Reddit's Terms of Service here.

if you think that we've incorrectly blocked you or you would like to discuss easier ways to get the data you want, please contact us at this email address.

when contacting us, please include your ip address which is: x.x.x.x and reddit account

saveriol:
We need to be able the redirect url of the apis too.

Otherwise "redditisfun://auth" is the next thing they'll block.

oSumAtrIX:
No, this is a local protocol handler.

calexil:
following this to see if this gets worked out, rif is bae

LogicalError007:
For me reddit is down on browsers too but it happened gradually, first Rif then infinity and then on browsers.

Sculas:
> No, this is a local protocol handler.

Still identifiable data you're giving to Reddit. Believe me, if they made a tool to delete all API keys with the RIF UA, it's really easy to do the same for the redirect URI.

oSumAtrIX:
It would be unconventional protocol to blacklist redirect URIs whereas checking for the user agent is not.

Sculas:
Convention this, convention that. No person follows all of the laws perfectly. If they're gonna block you from using RIF, they'll use every solution they have at hand, and that includes blocking a redirect URI.

oSumAtrIX:
[...]

PR #2743, opened 2023-08-03
fix(redditisfun): Change User-Agent to generic Android browser UA

nandryshak:
As mentioned in my comment in #2727 ([here](https://github.com/ReVanced/revanced-patches/issues/2727#issuecomment-1664391102)), I believe that changing the user agent to something that clearly identifies the app as being modified by revanced presents a security risk:

1) It risks the user of the app being banned by reddit for ToS violations.

2) It risks revanced getting a cease & desist from reddit. 

3) It makes it super easy for reddit to ban these user agents again in the future.

This is because the new way of generating a "more convincing" UA no longer has any plausible deniability. Before, the random number did not distinctly identify the app as being modded by revanced. Now, it does. Personally, I will not be using the patch as-is, because I do not want my account to be banned.

Clearly reddit is doing some type of "detective work" in order to block apps by their user agents. Therefore, a more thoughtful solution than identifying as "revanced" is needed.

~~The first solution that came to my mind is to use a generic browser user agent together with the random number. This prevents reddit from doing a simple search on all user agents containing the phrase "revanced" and blocking them.~~ edit: spoofing the UA is also a ToS violation, so we definitely need a custom UA here.

Let me know your thoughts on this, and let me know if I'm missing any conventions or PR etiquette for this particular project.

### DIFF
--- a/src/main/kotlin/app/revanced/patches/reddit/customclients/redditisfun/api/patch/SpoofClientPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/reddit/customclients/redditisfun/api/patch/SpoofClientPatch.kt
@@ -60,7 +60,7 @@ class SpoofClientPatch : AbstractSpoofClientPatch(
     override fun List<MethodFingerprintResult>.patchUserAgent(context: BytecodeContext): PatchResult {
         // Use a random user agent.
         val randomName = (0..100000).random()
-        val userAgent = "android:app.revanced.$randomName:v1.0.0 (by /u/revanced)"
+        val userAgent = "Mozilla/5.0 (Linux; Android 13) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/115.0.$randomName Mobile Safari/537.36"
 
         first().mutableMethod.addInstructions(
             0,


oSumAtrIX (changes requested):
No detective work is present, as explained in #2727 and the user agent proposed in this PR violates the one Reddit expects:

image

Use a schema that follows this format.

nandryshak:
> No detective work is present, as explained in #2727 and the user agent proposed in this PR violates the one Reddit expects:
> 
> image
> 
> Use a schema that follows this format.

That's useful to know, thanks. In this case I wonder if it's worth it to allow each user to add their own UA in the same way that we add our client id token.

dgw:
> In this case I wonder if it's worth it to allow each user to add their own UA in the same way that we add our client id token.

I'd say it is worthwhile. According to the prescribed format, whoever owns the application (generated the client ID) should be the `/u/name` that's included in the UA string, as I interpret the guidelines.

oSumAtrIX:
You could add an option but the default one should be randomized

Forceu:
I would suggest that each user has to provide an agent string, similar to how the API ID has to be provided by the user. That way it is ensured that a unique ID is used, has the username and this project would not violate the t&c

DaviddH:
It seems broken again now as i just mentioned in my original issue, why not mirror the UA of the official reddit app to prevent it from being blocked?

nandryshak:
> It seems broken again now as i just mentioned in my original issue, why not mirror the UA of the official reddit app to prevent it from being blocked?

Mine is still working okay. 

As mentioned by oSumAtrIX above and by someone else in 2727, spoofing the official app's UA would be a ToS violation, and they "will ban liars with extreme prejudice". 

So it definitely seems like we need the ability to use a custom UA. I can work on adding that tomorrow.

oSumAtrIX:
Is this still being worked on?

nandryshak:
> Is this still being worked on?

Yes, but I haven't put time into it recently. I'll see if I can get to it this week.

oSumAtrIX:
Reopen once ready.

PR #2750, opened 2023-08-04
feat(all - Change package name): Add default package name

Pro-Tweaker:
Small modification to the patch to change the package name. If no name is specified, the name of the new pakckage will be "original-package-name.revanced". 
I think this change is particularly useful for mobile users of the manager, as they can't use the patch options (yet).


oSumAtrIX:
Sounds like an interesting idea.

oSumAtrIX (approved):

src/main/kotlin/app/revanced/patches/all/misc/packagename/patch/ChangePackageNamePatch.kt
```diff
@@ -12,36 +12,40 @@ import org.w3c.dom.Element
 @Description("Changes the package name.")
```
oSumAtrIX, line 12:
```suggestion
@Description("Changes the package name. Appends \".revanced\" to the package name by default.")
```

PR #2874, opened 2023-08-25
feat(YouTube): Support version `18.32.39`

oSumAtrIX:
Depends on https://github.com/ReVanced/revanced-integrations/pull/464

# TODO

- [x] Fix resolution of `TextComponentAtomicReferenceFingerprint` fingerprint
- [x] Bump compatibility (as a separate commit)


oSumAtrIX:
Those patches support 18.20.39, but compatibility with older versions of 2-3 other patches is dropped with this bump. Fixing the issue with tablet devices is not a problem, I simply haven't looked into it yet. ReVanced Manager suggests the version to use that has the most patches, but the user can still choose to patch 18.20.39  with those 2-3 patches missing.

oSumAtrIX:
So far, everything works, 4 eyes are better than two though @LisoUseInAIKyrios.

PR #2931, opened 2023-09-04
feat(Grindr): Add `Unlock unlimited` patch

Slenderman00:
This patch hooks into the following methods and modifies their output as listed:
- HasFeatureFingerprint: TRUE
- IsFreeFingerprint: FALSE
- IsNoPlusUpsellFingerprint: TRUE
- IsNoXtraUpsellFingerprint: TRUE
- IsPlusFingerprint: FALSE
- IsUnlimitedFingerprint: TRUE
- IsXtraFingerprint: FALSE
- InnaccessibleProfileManagerbFingerprint: TRUE
- InnaccessibleProfileManagerdFingerprint : FALSE

Even though many of these methods typically return 'false', they still need to be hooked. Otherwise, instabilities may arise when Grindr runs promotions.

Due to a lack of distinction i am not able to reference the following functions without using their names:
- IsPlusFingerprint
- IsUnlimitedFingerprint
- IsXtraFingerprint

Despite some limitations, this patch remains effective for the three latest versions of Grindr.


Slenderman00:
I am not planning on abandoning it.

Slenderman00:
Could you explain what makes the app `Gay`?

Slenderman00:
Exactly, It certainly does! It does not however say "This app is gay". Grindr is a dating app serving the LGBTQ+ community.

Slenderman00:
Is tinder a straight app? maybe it`s bi? As i have already said, i am not planning on abandoning the patch.

PalmDevs:
> The Android app store titles the app as "Grindr - Gay chat": [play.google.com/store/apps/details?id=com.grindrapp.android](https://play.google.com/store/apps/details?id=com.grindrapp.android) But you insist the app is not "gay"
> 
> Arguing over which LGBTQ words are used is overlooking the question I asked.

Why does it matter?

PalmDevs:
> > Why does it matter?
> 
> Why does it matter which LGBTQ word is used? I don't understand either.

What's your problem with words? As long as the opposite sides understand, there's no need to do literal-speaking. Plus, it's not even related to the issue.

alexandreteles:
I wonder if this line of questioning would arise if we were talking about Tinder instead... I doubt it. In any case, the application being targeted at part of the LGBTQIA+ has nothing to do with patches being maintained or not. And the author said themselves that they will keep maintaining it, so as far as maintenance goes, this patch should be fine.

validcube:
> > I am not the one who brought up this issue.
> 
> Oh really? Where is that onslaught of people that starts raging because ReVanced has a patch for Grindr available?
> You're grasping for straws here, as if every patch is maintained by team at any given moment? Hilarious.
> You're clearly not aware how open source projects work, nor aware of the people that are in team. 

Please refrain from fueling the conversation.

Ushie:
No, let them confront the moron 🙂 eventually all these comments will be marked as offtopic, might aswell say what we want to say to the dumbass

validcube:
> No, let them confront the moron 🙂 eventually all these comments will be marked as offtopic, might aswell say what we want to say to the dumbass

Go to sleep, this is dad speaking.

oSumAtrIX:
It sounds odd to me that numerous fingerprints are necessary instead of one. What does the architecture look like? I imagine a server response that is deserialized into a user object with flags about the upgrade status of the user, which should be a single hook

Slenderman00:
The architecture of the app is a bit chaotic. This patch heavily relies on the UserSession storage class (storage.s0) which is the class that all of the grindr features interact directly with. Instances of this class are created all over the place, most notable by the api class. This class contains one role object and one featureFlags object, this class also contains the SessionId, AuthToken and xmppToken(chat).

Hooking these functions somehow enables all local premium features without getting the api class to send any "premium" requests.

I will try to hook the UserSession's clear role method and the UpdateRole method, replace the role from "FREE" to "UNLIMITED" and see what happens, This approach still requires hooking HasFeature and InnaccessibleProfileManager b and d

src/main/kotlin/app/revanced/patches/grindr/patch/FirebaseGetCertPatch.kt
oSumAtrIX, line 1: This looks very similar to another firebase patch. Introduce abstraction to reduce duplicate code.
Also move the patch to a subpackage, not just under "grindr".

src/main/kotlin/app/revanced/patches/grindr/unlimited/fingerprints/IsPlusFingerprint.kt
```diff
@@ -0,0 +1,30 @@
...
+        Opcode.APUT_OBJECT,
+        Opcode.CONST_4,
+        Opcode.SGET_OBJECT,
+        Opcode.APUT_OBJECT,
+        Opcode.INVOKE_STATIC,
+        Opcode.MOVE_RESULT,
+        Opcode.RETURN
+    ),
+    customFingerprint = { methodDef, _ ->
+        methodDef.name == "y"
```
oSumAtrIX, line 28: This has to be removed, otherwise the fingerprint will fail to resolve on different versions.

src/main/kotlin/app/revanced/patches/grindr/unlimited/fingerprints/IsUnlimitedFingerprint.kt
```diff
@@ -0,0 +1,30 @@
...
+        Opcode.APUT_OBJECT,
+        Opcode.CONST_4,
+        Opcode.SGET_OBJECT,
+        Opcode.APUT_OBJECT,
+        Opcode.INVOKE_STATIC,
+        Opcode.MOVE_RESULT,
+        Opcode.RETURN
+    ),
+    customFingerprint = { methodDef, _ ->
+        methodDef.name == "x"
```
oSumAtrIX, line 28: This has to be removed, otherwise the fingerprint will fail to resolve on different versions.

src/main/kotlin/app/revanced/patches/grindr/unlimited/fingerprints/IsXtraFingerprint.kt
```diff
@@ -0,0 +1,30 @@
...
+        Opcode.APUT_OBJECT,
+        Opcode.CONST_4,
+        Opcode.SGET_OBJECT,
+        Opcode.APUT_OBJECT,
+        Opcode.INVOKE_STATIC,
+        Opcode.MOVE_RESULT,
+        Opcode.RETURN
+    ),
+    customFingerprint = { methodDef, _ ->
+        methodDef.name == "p"
```
oSumAtrIX, line 28: This has to be removed, otherwise the fingerprint will fail to resolve on different versions.

src/main/kotlin/app/revanced/patches/grindr/unlimited/patch/setUnlimited.kt
```diff
@@ -0,0 +1,91 @@
...
+        IsNoXtraUpsellFingerprint,
+        IsPlusFingerprint,
+        IsUnlimitedFingerprint,
+        IsXtraFingerprint,
+        InnaccessibleProfileManagerbFingerprint,
+        InnaccessibleProfileManagerdFingerprint
+    )
+) {
+
+    class Feature(val feature: MethodFingerprint, val bytecode: String, val offset: Int = 0) {
```
oSumAtrIX, line 45: This structure looks odd. Why is a method fingerprint a "feature"? This edges on code obfuscation.

Call the class "Patch", call the fingerprint "target".

src/main/kotlin/app/revanced/patches/grindr/unlimited/patch/setUnlimited.kt
```diff
@@ -0,0 +1,91 @@
...
+
+        val TRUE_BYTECODE = """
+            const/4 v0, 0x1
+            return v0
+        """.trimIndent()
+
+        val FALSE_BYTECODE = """
+            const/4 v0, 0x0
+            return v0
+        """.trimIndent()
```
oSumAtrIX, line 68: Trimming is not required.
```suggestion
        val TRUE_BYTECODE = """
            const/4 v0, 0x1
            return v0
        """
        val FALSE_BYTECODE = """
            const/4 v0, 0x0
            return v0
        """
```

src/main/kotlin/app/revanced/patches/grindr/unlimited/patch/setUnlimited.kt
```diff
@@ -0,0 +1,91 @@
...
+                }
+            } ?: throw feature.exception
+        }
+
+    }
+
+
+    override fun execute(context: BytecodeContext) {
+
+        val TRUE_BYTECODE = """
```
oSumAtrIX, line 60: This is not a constant, use the correct naming convention for variables.

oSumAtrIX, line 60: Instead of passing the SMALI for compilation to every "Feature" class, compile it once. There are instruction list overloads for replaceInstructions.

src/main/kotlin/app/revanced/patches/grindr/unlimited/patch/setUnlimited.kt
```diff
@@ -0,0 +1,91 @@
...
+import app.revanced.patcher.fingerprint.method.impl.MethodFingerprint
+
+import app.revanced.patches.grindr.patch.FirebaseGetCertPatch
+
+@Patch
+@Name("Unlock unlimited")
+@Description("Unlocks unlimited features.")
+@UnlockUnlimitedCompatibility
+@DependsOn([FirebaseGetCertPatch::class])
+
```
oSumAtrIX, line 30:
```suggestion
```

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/grindr/patch/FirebaseGetCertPatch.kt
Slenderman00: I would have to rewrite the warnwetter FirebaseGetCertPatch to something more reusable.

Slenderman00:
@LisoUseInAIKyrios modifying the application hardly counts as an exploit.

Slenderman00:
@LisoUseInAIKyrios, their rest API and XMPP setup are extremely bad. If you want to find some exploits you should reverse engineer their API and probe it a bit. They do almost no message verification for example, it used to be possible to sign xmpp messages with fake names and fake user ids to send untraceable messages.

src/main/kotlin/app/revanced/patches/grindr/patch/FirebaseGetCertPatch.kt
oSumAtrIX: That works

kazimmt:
Is it necessary to made a ReVanced patch for an app which is banned in many countries?
Thumbs down from me.

RobbieTechie:
> It sounds odd to me that numerous fingerprints are necessary instead of one. What does the architecture look like? I imagine a server response that is deserialized into a user object with flags about the upgrade status of the user, which should be a single hook

The sessions from their server set the features.  Those sessions are base64 encoded when they arrive at the app.  The features granted by the session also get placed into jwt_fallback_preferences.xml in shared_prefs.  A while back, one could edit the jwt_fallback_preferences.xml file and relaunch the app to see what changes would happen.  The sessions last one hour on Android and 1/2 hour on iOS

Here is an example from about one year ago of the jwt_fallback_preferences.xml file:

<?xml version='1.0' encoding='utf-8' standalone='yes' ?>
<map>
    <long name="session_expiration_time" value="[redacted]" />
    <string name="session_features_fallback">AlbumsOnlyFilter,,BlockRemoteProfiles,,ChatDeliveryStatus,,ChatReadStatus,,ChatRemoteProfiles,,CreateLiveStream,,CreateVideoCall,,FaceOnlyFilter,,FavoriteRemoteProfiles,,HaveChattedHighlight,,HaventChattedFilter,,HideBannerAds,,HideFullScreenAds,,HideUnlimitedStore,,HideXtraStore,,Incognito,,LastChattedTimestamp,,LongAudioCall,,MoreChatGroups,,MultiTribes,,MyTypeFilters,,OnlineNowFilter,,PhotosOnlyFilter,,PriorityPhotoModeration,,PushPreview,,PushSnoozing,,SavedPhrases,,SendMultiplePics,,Subscriber,,TapFilters,,TapRemoteProfiles,,TranslateMessage,,TypingIndicator,,UnlimitedBlocklist,,UnlimitedBlocks,,UnlimitedCascade,,UnlimitedExpiringPhotos,,UnlimitedFavorites,,UnlimitedTagCascade,,UnsentMessage,,VideoCallMonthlyCalculateLengthHigh,,ViewedMeCascade,,ViewedMeTimestamp</string>
    <string name="session_roles_fallback">Unlimited</string>
</map>

Some more things have been added with updates to the app such as some of the experiments they run etc...They are currently running around 31 experiments in the app and some can be toggled on and off such as "profile-redesign-20230214" which changes the look of a profile when someone views it.

Not ALL features are local to the app, a few are more interactive and controlled server side such as the ability to view albums or see the entire list of viewers, but most things are local.  I just integrated the Grindr Plus mod into the apk using Shizuku and it worked, but Shizuku has to be running.  I feel having this patch as an option for those that want it is a better option than Shizuku.  I'm not sure if this has any weight with anyone here, but I've seen Slenderman00 work on other things dealing with Grindr in the past so I don't see him abandoning this project.  For what it's worth I vote yes to support this.

src/main/kotlin/app/revanced/patches/grindr/unlimited/patch/setUnlimited.kt
Slenderman00: I can't seem to find the instruction list overloads that you are describing.

Slenderman00:
@LisoUseInAIKyrios with descriptive sentences like "Totally, this social web-based application assists you with tracking down new individuals as your fellowship". I don't see how this can be anything other than the same application!

Issue #2570, opened 2023-07-02
bug: Remove Screenshot restriction global patch doesnt work on Telegram.


gthu34:
### Type

Other

### Bug description

I tried to patch three different version of telegram.([Playstore](https://play.google.com/store/apps/details?id=org.telegram.messenger) , [web](https://telegram.org/android)  and [forkgram](https://f-droid.org/en/packages/org.forkgram.messenger/)) 
but none worked after patching the app. the apk installs fine but it either freezes on login screen or nothing happens when you put login details. 
the forkgram installed succesfully but i couldnt take screenshot. it said "cant take screenshot due to security policy". 
forkgram is a third party open source client for telegram on fdroid.

### Steps to reproduce

Step 1. Download telegram or forkgram form any of the sources mentioned above.
Step 2. Get the apk
Step 3. load the apk from storage using Revanced Manager
Step 3. patch it using two patches named _Remove Screenshot restriction_ and _Remove screen Capture restriction_
Step 5. after patching is done , install it.
Step 6. Now open the app you will have login screen.
Step 7. Now the screen will freeze or if not then type login detail and nothing will happen, no otp screen comes up.
if it is Forkgram then login will be successful. now go to any channel/group or make one your self with _restrict content_ enable. 
now take screenshot. you cant take one.

### Relevant log output

```shell
Initializing installer
Creating working directory
Copying original apk
Unpacking input apk
Reading dex files
Decoding AndroidManifest.xml only, because resources are not needed
Merging integrations
Deleting existing resource cache directory
Decoding resources
Executing patches
Applied remove-screen-capture-restriction
Applied remove-screenshot-restriction
Repacking patched apk
Compiling resources
Writing modified dex files
Finished!
Installing patched file using nonroot method
Installed!
```

### Solution

please make separate patch for telegram or just fix the global patch for telegram.

### Acknowledgements

- [X] I have searched the existing issues and this is a new and no duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I filled out all of the requested information in this issue properly.

typingbeaver:
Telegram/Forkgram sets a secure flag ([`WindowManager.LayoutParams.FLAG_SECURE`](https://github.com/search?q=repo%3Aforkgram/TelegramAndroid%20FLAG_SECURE&type=code)) preventing the system to take screenshots/recordings.

They should work if this flag is removed.

balajinoob:
[...]

PR #2938, opened 2023-09-05
fix(Remove screenshot restriction): Improve reliability

Linus789:
issue: ReVanced/revanced-patches-template#651
integrations: https://github.com/ReVanced/revanced-integrations/pull/471

@gthu34 could you check if this fixes your issue?


Linus789:
@LisoUseInAIKyrios I'm not able to test if this PR fixes the problem, because my ROM ignores all screenshot restrictions by default.

oSumAtrIX:
Is this PR still active? If so, mark as ready for review.

Linus789:
The PR is still active, I just have to find some time to test it.

Linus789:
Okay, I just tested it with Telegram and it seems to work. So it should close #2570.

oSumAtrIX (approved):

PR #2963, opened 2023-09-09
build: migrate some YouTube patches

TheAabedKhan:


src/main/kotlin/app/revanced/patches/youtube/interaction/copyvideourl/bytecode/CopyVideoUrlBytecodePatch.kt
```diff
@@ -0,0 +1,49 @@
+package app.revanced.patches.youtube.interaction.copyvideourl.bytecode
```
oSumAtrIX, line 1: Move this out of the bytecode package

src/main/kotlin/app/revanced/patches/youtube/interaction/copyvideourl/resource/CopyVideoUrlResourcePatch.kt
```diff
@@ -1,8 +1,8 @@
-package app.revanced.patches.youtube.interaction.copyvideourl.resource.patch
+package app.revanced.patches.youtube.interaction.copyvideourl.resource
```
oSumAtrIX, line 1: Move this out of the resource package

oSumAtrIX (changes requested):
From skimming, some patches are still kept in resource or bytecode packages. You can move them out to the parent package: some/bytecode/patch.kt -> some/patch.kt.

The rest looked solid, I'll do a proper review later.

PR #2983, opened 2023-09-11
refactor: Fix grammar in toggle description

KobeW50:
closes https://github.com/ReVanced/revanced-patches-template/issues/365

I have no idea what I am doing. I did a global search for the description text and then edited it.

Sorry if I missed something


src/main/kotlin/app/revanced/patches/youtube/layout/hide/general/patch/HideLayoutComponentsPatch.kt
```diff
@@ -55,7 +55,7 @@ class HideLayoutComponentsPatch : BytecodePatch(
                 "revanced_hide_expandable_chip",
                 StringResource(
                     "revanced_hide_expandable_chip_title",
-                    "Hide the expandable chip under videos"
+                    "Hide expandable chip under videos"
```
oSumAtrIX, line 58: This is incorrect grammar. Either you include "the" or say "chips". If you do the latter other references such as the key have to be renamed to "chips", including on the integrations side.

oSumAtrIX (approved):

PR #2979, opened 2023-09-11
feat(Tumblr): Add `Disable dashboard ads` patch

leumasme:
This is both the first revanced patch for tumblr and my first revanced patch.
Related to ReVanced/revanced-patches-template#439, ReVanced/revanced-patches-template#1477 but doesn't really close them since they both request many more patches at once.  

All this patch does is replace the string "client_side_ad_waterfall" anywhere in the app code with "nope".  
This string is used by the tumblr api as the object ID for an ad in the User Dashboard.  
By replacing it in the app code, we make the tumblr app fail to recognize what type of dashboard object the ads should be, leading to them simply not being shown. This is also documented in code comments.  

Tested in app version 31.1.0.110, but marked for compatability with all versions since this patch is very simple and very unlikely to break.

PR #2987, opened 2023-09-12
feat(Tumblr): Add `Disable Tumblr Live` patch

leumasme:
The DisableTumblrLivePatch
- Changes the expected ID of the tumblr live dashboard/timeline carousel element to a dummy so the app will not recognize it (see #2979)
- Depends on FeatureOverridePatch to overwrite the value of an A/B testing feature flag to hide the Live tab button in the bottom navbar

The FeatureOverridePatch 
- Adds a new helper function to `com.tumblr.configuration.Configuration` which accepts a Feature object and returns the overwrite value if one is present, or null if the feature should not be overwritten
- Hooks a method on `com.tumblr.configuration.Configuration` responsible for getting the string value of a Feature object to call the helper function and return its value instead if it doesn't return null
- Presents an `addOverride(String name, String value)` delegate method which adds the equivalent of  
 `if (currFeatureName == {name}) return {value};` in the middle of the helper function. This is currently only used once, but has been tested to work with multiple feature overrides.

The FeatureOverridePatch is not registered as a @ Patch since it is not usable by itself. A patch options screen could be added in the future to let the user set custom feature overrides, making the Patch indepentently usable.


src/main/kotlin/app/revanced/patches/tumblr/featureflags/patch/OverrideFeatureFlagsPatch.kt
```diff
@@ -0,0 +1,118 @@
...
+                    
+                    # !!! If you add more instructions above this line, update helperInsertIndex below!
+                    # Here we will insert one compare & return for every registered Feature override
+                    # This is done below in the addOverride function
+                    
+                    # If none of the overrides returned a value, we should return null
+                    const/4 v0, 0x0
+                    return-object v0
+                """
+            )
```
oSumAtrIX, line 59: It is suggested to move such implementations to [ReVanced Integrations](https://github.com/revanced/revanced-integrations). How it works is that you implement your method there, ReVanced Patcher merges the compiled method and you can use it in the patches. If the method references code from the app itself, you can add classes to the "dummy" module. Since your patch, later on, adds instructions to that helper method, you can use a fingerprint to locate the method you added to ReVanced Integrations.

leumasme: > such implementations

What makes this an implementation to move over there? Simply because it's a new method?  
It seems a bit silly to separate it into the integrations repo since it's just a toString call and a return, but if you say so then I'll do it this way.

oSumAtrIX: Basically readability. It is much easier to maintain and read a java method instead of bytecode.

leumasme: (as discussed in discord) moving the function into integrations doesn't really work since it would compile away the unused variable and get complied with only one register. we need at least two registers for the later injected instructions, which just moves the problem of too few registers which was the reason for creating the helper function in the first place.  
I've updated comments to also explain this and to provide the java-equivalent of the injected smali instructions.

TeaOwO:
Yes I need this!

PR #2990, opened 2023-09-13
feat(tumblr): Add anti-annoyance patches

leumasme:
DisableGiftMessagePopupPatch:
Disables the "You can now give back to the people you love" popup when visiting a blog for the first time.
Adds a `return false` at the start of the `BlogNotifyCtaDialog.isEnabled` method

DisableBlogNotificationReminderPatch:
Disables the "Want to get a notification when this person posts?" popup when visiting a specific blog for the second time
Adds a `return` at the start of the `GiftMessagePopup.(showPopup)` method


oSumAtrIX:
Please open a separate pull request for each patch so it can be squashed.

PR #2997, opened 2023-09-16
feat(Strava): Add `Hide upselling` patch

xehpuk:
Hides annoying suggestions to subscribe for premium features.

Mostly prominent on the progress tab and activity pages.

### DIFF
--- a/src/main/kotlin/app/revanced/patches/strava/annotations/StravaCompatibility.kt
+++ b/src/main/kotlin/app/revanced/patches/strava/annotations/StravaCompatibility.kt
@@ -1,0 +1,8 @@
+package app.revanced.patches.strava.annotations
+
+import app.revanced.patcher.annotation.Compatibility
+import app.revanced.patcher.annotation.Package
+
+@Compatibility([Package("com.strava", ["324.10"])])
+@Target(AnnotationTarget.CLASS)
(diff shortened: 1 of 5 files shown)


xehpuk:
They're totally independent of each other. The only thing they have in common is their uselessness when truly being subscribed.

oSumAtrIX:
Isn't the subscription suggestion normally not shown if you are subscribed which the subscription unlock patch is mimicking?

xehpuk:
The app seems to be a thin client most of the time, so it just renders what the server tells it to. And the server obviously doesn't know about our patched app.

The subscription unlock patch works just for a small subset of all features. That's where the server sends all the needed data regardless of the subscription status and only the client checks the flag.

xehpuk:
The intended behavior of the other patch is just to enable what's possible.

The thing about this patch is: The sections which get hidden are not 100 % useless. There's some very superficial data. As soon as you tap them to see more, it just tells you to subscribe.

So while the other patch doesn't seem to have a downside, this one may be a matter of taste to some.

oSumAtrIX (approved):

emafed:
hello, is possibile to update the app suggested version (320.12) to the last app version (336.11)? thanks

PR #3001, opened 2023-09-17
feat(youtube - spoof app version): add version 18.19.35

Julienraptor01:
this makes so the video speed and quality menu take all screen instead of just the small square in the middle of the screen

example :
not spoofed
new
vs
spoofed
old


Julienraptor01:
> Version 18.20.39 is the most recent version that uses the old quality/speed flyout menu.  But could use 18.20.35 to keep the version numbers more consistent (all the other targets are x.x.35)
> 
> It might be better keeping targets sorted with the newest first and oldest last.  That matches the video quality picker that has highest quality first and lowest quality last.

The thing is i've used .35 because it was a supported version for revanced
I don't think 18.20.39 was ever recommended by revanced
Or at least, i missed it

Julienraptor01:
> The thing is i've used .35 because it was a supported version for revanced
> I don't think 18.20.39 was ever recommended by revanced
> Or at least, i missed it  

From reading different patch supported versions it seems like 18.20.35 was replaced by .39 so i'll change it
Having the latest version that support the feature is a good idea

Julienraptor01:
PR should now be tested again to make sure it work (can't do that RN so if someone else can test, would be kind)

Julienraptor01:
OK i'll invert the order

Julienraptor01:
OK i'll invert the order

Julienraptor01:
I've just done the changes you asked for

Julienraptor01:
i did just rebase right now so it can be easily merged

src/main/kotlin/app/revanced/patches/youtube/layout/spoofappversion/SpoofAppVersionPatch.kt
```diff
@@ -60,17 +60,19 @@ object SpoofAppVersionPatch : BytecodePatch(
                 ArrayResource(
                     "revanced_spoof_app_version_target_entries",
                     listOf(
-                        StringResource("revanced_spoof_app_version_target_entry_1", "17.08.35 - Restore old UI layout"),
-                        StringResource("revanced_spoof_app_version_target_entry_2", "16.08.35 - Restore explore tab"),
-                        StringResource("revanced_spoof_app_version_target_entry_3", "16.01.35 - Restore old Shorts player"),
+                        StringResource("revanced_spoof_app_version_target_entry_1", "18.20.39 - Restore old video speed & quality UI"),
```
oSumAtrIX, line 63:
```suggestion
                        StringResource("revanced_spoof_app_version_target_entry_1", "18.20.39 - Restore wide video speed & quality menu"),
```

oSumAtrIX (changes requested):

oSumAtrIX (approved):

src/main/kotlin/app/revanced/patches/youtube/layout/spoofappversion/SpoofAppVersionPatch.kt
Julienraptor01: yeah better wording 👍

PR #3011, opened 2023-09-22
feat(CieID): Add `bypass root check` patch

GioPan04:
CieID is the official Android app created by the Italian government for logging in government websites, and it blocks every device with root or a custom rom.
This PR patches the onResume callback which is used to check if the device has been rooted or has a custom rom.


GioPan04:
I just saw a possible bug. I have to investigate, in the meantime I'll fix the errors with the current version

GioPan04:
Okay, so I looked into it, seems fine, the PR is ready.

PR #3012, opened 2023-09-22
feat: Publicize abstract settings resource patch

Aunali321:
This PR adds the functions to the public abstract settings resource patch.


Aunali321:
Patches will soon be published on maven central so other bundles can depend on it. I was trying to implement custom Instagram settings in my fork which requires this methods. I currently have to duplicate code. So this was necessary to publicize.

oSumAtrIX (approved):

PR #3018, opened 2023-09-23
refactor(Tumblr): Use a common filter patch

leumasme:
Depends on https://github.com/ReVanced/revanced-integrations/pull/479

Previously, we blocked tumblr timeline objects by replacing their expected "object_type" string that is returned from the api.  
This was suboptimal for a few reasons:
- It created a lot of boilerplate with finding and replacing strings
- It didn't completely remove the objects from the timeline, they just failed to parse so they were <null> objects. These are not visible, but still reserve space when the timeline is initially created, leading to layout shifting which was visible if the blocked timeline object was in view right away.

This PR adds a Timeline Filter Patch, which
- Injects a call to the integration filter method in `Timeline` and `PostsResponse` which are used for the front page and the blog view (clicking on a blog to see only posts from this block) respectively.
- Presents a `addObjectTypeFilter(typename: String)` method which injects the equivalent of  
`if ("$typename".equals(elementType)) iterator.remove()`  
into the integration filter function.

The two patches `DisableTumblrLivePatch` and `DisableDashboardAdsPatch` patch have also been updated to use the new timeline filtering.

*Additionally, more blocked ad post types were added to `DisableDashboardAds` since some ads were still visible*
*Wait, why are some patch file names suffixed with `Patch` and some are not? I guess I will just leave it like this.*


src/main/kotlin/app/revanced/patches/tumblr/ads/DisableDashboardAds.kt
```diff
@@ -1,33 +1,40 @@
...
+        // Called "client_side_ad" in api response
+        TimelineFilterPatch.addObjectTypeFilter("CLIENT_SIDE_AD")
+        // Called "display_io_interscroller" in api response
+        TimelineFilterPatch.addObjectTypeFilter("DISPLAY_IO_INTERSCROLLER_AD")
+        // Called "display_io_headline_video" in api response
+        TimelineFilterPatch.addObjectTypeFilter("DISPLAY_IO_HEADLINE_VIDEO_AD")
+        // Called "facebook_biddable_sdk_ad" in api response
+        TimelineFilterPatch.addObjectTypeFilter("FACEBOOK_BIDDAABLE")
+        // Called "google_native_ad" in api response
+        TimelineFilterPatch.addObjectTypeFilter("GOOGLE_NATIVE")
```
oSumAtrIX, line 38: Reduce duplicate code:

arrayOf("strings").forEach { TimelineFilterPatch.addObjectTypeFilter(it) }

src/main/kotlin/app/revanced/patches/tumblr/timelinefilter/TimelineFilterPatch.kt
```diff
@@ -0,0 +1,68 @@
...
+                    )
+                }
+            }
+        } ?: throw TimelineFilterIntegrationFingerprint.exception
+
+        TimelineConstructorFingerprint.result?.mutableMethod?.addInstructions(
+            0, """
+                    invoke-static {p1}, Lapp/revanced/tumblr/patches/TimelineFilterPatch;->filterTimeline(Ljava/util/List;)V
+                """
+        ) ?: throw TimelineConstructorFingerprint.exception
```
oSumAtrIX, line 61:
```suggestion
        TimelineConstructorFingerprint.result?.mutableMethod?.addInstructions(
            0,
            """
                invoke-static {p1}, Lapp/revanced/tumblr/patches/TimelineFilterPatch;->filterTimeline(Ljava/util/List;)V
            """
        ) ?: throw TimelineConstructorFingerprint.exception
```

src/main/kotlin/app/revanced/patches/tumblr/timelinefilter/TimelineFilterPatch.kt
```diff
@@ -0,0 +1,68 @@
...
+        PostsResponseConstructorFingerprint.result?.mutableMethod?.addInstructions(
+            0, """
+                    invoke-static {p2}, Lapp/revanced/tumblr/patches/TimelineFilterPatch;->filterTimeline(Ljava/util/List;)V
+                """
```
oSumAtrIX, line 65:
```suggestion
        PostsResponseConstructorFingerprint.result?.mutableMethod?.addInstructions(
            0,
            """
                invoke-static {p2}, Lapp/revanced/tumblr/patches/TimelineFilterPatch;->filterTimeline(Ljava/util/List;)V
            """
```

src/main/kotlin/app/revanced/patches/tumblr/timelinefilter/TimelineFilterPatch.kt
```diff
@@ -0,0 +1,68 @@
...
+                            nop
+                        """
+                    )
+                }
+            }
+        } ?: throw TimelineFilterIntegrationFingerprint.exception
+
+        TimelineConstructorFingerprint.result?.mutableMethod?.addInstructions(
+            0, """
+                    invoke-static {p1}, Lapp/revanced/tumblr/patches/TimelineFilterPatch;->filterTimeline(Ljava/util/List;)V
```
oSumAtrIX, line 59: Are you sure p1 won't change?

src/main/kotlin/app/revanced/patches/tumblr/timelinefilter/TimelineFilterPatch.kt
```diff
@@ -0,0 +1,68 @@
...
+import app.revanced.patcher.extensions.InstructionExtensions.addInstructions
+import app.revanced.patcher.extensions.InstructionExtensions.addInstructionsWithLabels
+import app.revanced.patcher.extensions.InstructionExtensions.removeInstructions
+import app.revanced.patcher.patch.BytecodePatch
+import app.revanced.patcher.patch.annotation.Patch
+import app.revanced.patches.tumblr.timelinefilter.fingerprints.PostsResponseConstructorFingerprint
+import app.revanced.patches.tumblr.timelinefilter.fingerprints.TimelineConstructorFingerprint
+import app.revanced.patches.tumblr.timelinefilter.fingerprints.TimelineFilterIntegrationFingerprint
+
+@Patch(description = "Filter what will be shown in the timeline.", requiresIntegrations = true)
```
oSumAtrIX, line 14:
```suggestion
@Patch(description = "Filter timeline objects.", requiresIntegrations = true)
```

src/main/kotlin/app/revanced/patches/tumblr/timelinefilter/TimelineFilterPatch.kt
```diff
@@ -0,0 +1,68 @@
...
+     *
+     * @param typename The enum name of the timeline object type to hide.
+     */
+    @Suppress("KDocUnresolvedReference")
+    internal lateinit var addObjectTypeFilter: (typename: String) -> Unit private set
+
+    override fun execute(context: BytecodeContext) {
+
+        TimelineFilterIntegrationFingerprint.result?.let { integration ->
+            val startIndex = integration.scanResult.patternScanResult!!.startIndex
```
oSumAtrIX, line 31: Maybe give this a better variable name. Something that describes what the current index in the code is.

src/main/kotlin/app/revanced/patches/tumblr/timelinefilter/TimelineFilterPatch.kt
```diff
@@ -0,0 +1,68 @@
...
+                removeInstructions(startIndex, 5)
+
+                addObjectTypeFilter = { typename ->
+                    // It's too much of a pain to find the register numbers manually, so this will just have to be
+                    // updated if the Timeline Filter integration changes
+                    // The java equivalent of this is
+                    //   if ("BLOCKED_OBJECT_DUMMY".equals(elementType)) iterator.remove();
+                    addInstructionsWithLabels(
+                        startIndex, """
+                            const-string v1, "$typename"
```
oSumAtrIX, line 44:
```suggestion
                addObjectTypeFilter = { typeName ->
                    // It's too much of a pain to find the register numbers manually, so this will just have to be
                    // updated if the Timeline Filter integration changes
                    // The java equivalent of this is
                    //   if ("BLOCKED_OBJECT_DUMMY".equals(elementType)) iterator.remove();
                    addInstructionsWithLabels(
                        startIndex, """
                            const-string v1, "$typeName"
```

src/main/kotlin/app/revanced/patches/tumblr/timelinefilter/TimelineFilterPatch.kt
```diff
@@ -0,0 +1,68 @@
...
+    internal lateinit var addObjectTypeFilter: (typename: String) -> Unit private set
+
+    override fun execute(context: BytecodeContext) {
+
+        TimelineFilterIntegrationFingerprint.result?.let { integration ->
+            val startIndex = integration.scanResult.patternScanResult!!.startIndex
+
+            integration.mutableMethod.apply {
+                // Remove "BLOCKED_OBJECT_DUMMY" object type filter
+                removeInstructions(startIndex, 5)
```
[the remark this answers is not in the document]
leumasme: I have tried the approach of modifying (class member) static array instead, but could not get the compiler to output the array instantiation in a way that can be easily modified. Maybe because it is created via constant string primitives, but the static array is simply being generated as a bunch of `const-string` and then a single `filled-new-array`. This is of course not extendable since it needs more registers the more elements are added.

If you can find a way to actually implement this, that would be great - otherwise I don't know how this could be done without wrapping the filters into classes or fully re-implemeting the array instantiation in smali.

src/main/kotlin/app/revanced/patches/tumblr/timelinefilter/TimelineFilterPatch.kt
leumasme: p1 here is actually the first argument as it is passed to the constructor (since we inject at the start of the function) - it could only change if the constructors argument order was switched.  
Although this could also be theoretically tracked since the argument we need is annotated with @Json(name="posts"), I'm not sure how to do this and don't think it's needed either.

src/main/kotlin/app/revanced/patches/tumblr/timelinefilter/TimelineFilterPatch.kt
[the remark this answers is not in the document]
leumasme: I'm not sure how that would help. If you would like to try to get it working in a more elegant way than currently, feel free to - I've already tested multiple ways for multiple hours and have enough of fighting the java compiler simply because the patch-integration separation makes this so painfully difficult.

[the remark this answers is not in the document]
leumasme: I've added a warning comment. In the future if/when there are multiple types of filters, it could be refactored into wrapper classes for the filters since they would then have a proper purpose.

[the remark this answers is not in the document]
oSumAtrIX: > If you can find a way to actually implement this, that would be great.

If you can not abuse the compiler to generate what you want, you can do it manually via a patch.

1. Add a field to your mutable class:
2. Inject these instructions in your class initializer:

   image
3. Modify your `addFilter` delegate, so that it appends "array add item" instructions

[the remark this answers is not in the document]
leumasme: Then I will either have to inject a method (initializer) or fight with increasing its register count again if it doesn't have enough.

Is this required to get this merged?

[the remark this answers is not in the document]
oSumAtrIX: The class initializer is present in every compiled class, you just need to inject the bytecode to fill the array. It would indeed be better to do this as it would be equal to the other filter patches patch wise.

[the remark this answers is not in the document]
oSumAtrIX: That's how the litho filters are also initialized.

leumasme (commented):
@oSumAtrIX This refactor does not work. The two added instructions used different registers (p1 / p2) because the argument we need is in different positions in the two fingerprinted methods.

oSumAtrIX:
@leumasme Instead of an array you can map the fingerprints to the correct register and then use it in the forEach loop.

src/main/kotlin/app/revanced/patches/tumblr/timelinefilter/TimelineFilterPatch.kt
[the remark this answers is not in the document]
leumasme: Updated to use a static class member list and inject List.add("{elem}") calls instead

leumasme:
Made it work again and updated the injection method as requested. Should be ready to merge (together with the integrations pr).

src/main/kotlin/app/revanced/patches/tumblr/timelinefilter/TimelineFilterPatch.kt
```diff
@@ -28,30 +28,25 @@ object TimelineFilterPatch : BytecodePatch(
...
                 removeInstructions(filterInsertIndex, 2)
 
                 addObjectTypeFilter = { typeName ->
-                    // The java equivalent of this is
-                    //   blockedObjectTypes.add({typeName})
+                    // blockedObjectTypes.add({typeName})
                     addInstructionsWithLabels(
                         filterInsertIndex, """
                             const-string v$stringRegister, "$typeName"
-                            invoke-interface { v$listRegister, v$stringRegister }, Ljava/util/List;->add(Ljava/lang/Object;)Z
```
leumasme: This does not work. HashSet is a class and not an interface, invoke-virtual should be used. Also because of your refactor, the dummy filter fingerprint is broken since it matches `const-string, invoke-interface` but HashSet.add gets compiled to `const-string, invoke-virtual`.

Either update the fingerprint and the injected smali to invoke-virtual, or update the integration to have variable type Set instead of HashSet (the interface that HashSet implements) and update the injected smali to `invoke-interface` Set.add

leumasme (commented):
@oSumAtrIX This PR does not work anymore with the changes you added before merging. It currently fails both in apply-patch stage and in app-launch stage.

PR #3028, opened 2023-09-25
feat(YouTube): Bump compatibility to `18.37.36`

oSumAtrIX:
Depends on https://github.com/ReVanced/revanced-integrations/pull/483

# TODO

- [x] `Remember video quality`: Fix class cast exception (likely offset issue)
- [x] `Return YouTube Dislike`: Failing fingerprint `TextComponentAtomicReferenceFingerprint`
- [x] `Return YouTube Dislike`: Add support for new shorts player UI (dislikes do not appear in shorts)
- [x] ~~`Return YouTube Dislike`: Fix shorts player not showing dislikes on first video loaded~~  Edit: problem is unrelated to 18.37.36
- [x] `Disable auto captions`: Fix patching null pointer exception
- [x] `Hide endscreen cards`: Failing fingerprint `LayoutIconFingerprint`
- [x] `Hide Shorts components`: Add support for new shorts UI components (patch does not function)


oSumAtrIX:
> shorts layout patch

Do you mean the `Hide Shorts components` patch? In case the buttons are now built by Litho, we should be able to filter them using the Litho filter patch. Do they appear in the logs?

> That means both RYD...

If the shorts dislike button is built by Litho, a TextSpec component should be used for the "Dislike" label. This means the same hook can be used as the normal player dislike button to filter the correct TextSpec component using the path builder object.

oSumAtrIX:
I have confirmed that the shorts components now show up in the logs and can be blocked, I'll push a fix for them but constrain the compatibility of this patch to the latest version to get rid of legacy code.

oSumAtrIX:
Alright, I'll try to keep it.

oSumAtrIX:
@LisoUseInAIKyrios  It may be possible to filter properly through protobufBufferArray. The abstract class Filter has two fields: 

image

But this class is missing a field for `ByteArrayFilterGroupList`. Is there a specific reason for this? 
Also, the inline docs say:

> If you have more than 1 filter patterns, then all instances of this class should filtered using {@link ByteArrayFilterGroupList#check(byte[])},

Do you mean filter patterns in the sense of `ByteArrayAsStringFilterGroup` or the actual filters inside that group? If I have two or more `ByteArrayAsStringFilterGroup`, should `ByteArrayFilterGroupList#check(byte[])` be used in this case, or when I have `ByteArrayAsStringFilterGroup` with multiple filters?

oSumAtrIX:
All todos are now ticked. I am not yet entirely sure about the spoof to old version issue as I was not able to reproduce it with the latest commits. Once you give me the green light with https://github.com/ReVanced/revanced-integrations/pull/483#issuecomment-1736397443 we should be able to merge.

oSumAtrIX:
Does the video desync with audio or sync after the request is finished? If it desyncs, it may be better to set a shorter timeout.

oSumAtrIX:
Unfortunately, we do not know any hook that would allow us to get a reference that we can pass via a callback to the future to edit the text, so this is gonna be a limitation.

oSumAtrIX:
Each patch is constrained to versions. The patches with issues can remain as is (as long as they are compatible with all the previous versions) while the other working patches can be bumped. This branch can then be merged to dev and main. ReVanced Manager will continue to suggest the current version as most patches support it, while being able to patch the latest version either including the patches that do not yet work reliably (toggling in the settings) or without them (default).

oSumAtrIX:
I've been also testing this version and it works fine, I can personally not reproduce blocking thread issue, but I suppose it occurs on slow networks. Once I'm home, I'll append the new compatible version to the annotation field and merge

oSumAtrIX:
RYD patch lost compatibility with the current supported version. Either we have to revert the changes for RYD patch in this PR or we have to constrain it to the currently working version (with the present issue).

cyberboh:
Missing something:
```
INFO: Decoding app manifest
WARNING: Player flyout menu is incompatible with version 18.37.36. This patch is only compatible with version com.google.android.youtube: 18.16.37, 18.19.35, 18.20.39, 18.23.35, 18.29.38, 18.32.39
```

PR #3029, opened 2023-09-25
feat(YouTube - Premium heading): Allow using default heading

aliernfrog:
Lets users use the non-premium heading even if they have premium.


oSumAtrIX (changes requested):

aliernfrog:
Applied requested changes, tested and works fine.

PR #3030, opened 2023-09-25
fix(YouTube - Client spoof): Removed unused code

LisoUseInAIKyrios:
Changed the patch point to pass in the video id.  The video id of `VideoInformation` is not yet updated and it's either null or it's the previous video loaded.

Also removed old unused code from the prior seekbar thumbnail partial fix.

[integration changes](https://github.com/ReVanced/revanced-integrations/pull/480)


src/main/kotlin/app/revanced/patches/youtube/misc/fix/playback/SpoofSignaturePatch.kt
```diff
@@ -88,40 +93,88 @@ object SpoofSignaturePatch : BytecodePatch(
...
                         invoke-static {}, $INTEGRATIONS_CLASS_DESCRIPTOR->getStoryboardRendererSpec()Ljava/lang/String;
                         move-result-object p$storyBoardUrlParams
                     """,
-                        ExternalLabel("ignore", getInstruction(0))
+                    ExternalLabel("ignore", getInstruction(0))
+                )
+            }
+        } ?: throw StoryboardRendererSpecFingerprint.exception
+
+        SettingsPatch.PreferenceScreen.MISC.addPreferences(
```
oSumAtrIX, line 132: For consistency reasons, move this to the top of the execute method.

src/main/kotlin/app/revanced/patches/youtube/misc/fix/playback/SpoofSignaturePatch.kt
```diff
@@ -45,22 +50,22 @@ object SpoofSignaturePatch : BytecodePatch(
...
-                val protobufParameterRegister = 3
+                val videoIdParameter = 1
+                val protobufParameter = 3
 
                 addInstructions(
                     0,
                     """
-                        invoke-static {p$protobufParameterRegister}, $INTEGRATIONS_CLASS_DESCRIPTOR->spoofParameter(Ljava/lang/String;)Ljava/lang/String;
-                        move-result-object p$protobufParameterRegister
+                        invoke-static {p$videoIdParameter, p$protobufParameter}, $INTEGRATIONS_CLASS_DESCRIPTOR->spoofParameter(Ljava/lang/String;Ljava/lang/String;)Ljava/lang/String;
```
oSumAtrIX, line 59: Instead of this, make `VideoIdPatch` hook `videoIdParameter` earlier. You can hook the same method in addition to the current hook of `VideoIdPatch` so that `VideoInformationPatch` can be used in this patch. I believe the earliest useful time for a video id hook is from the player model response. One of the getters that returns the video id can be hooked which would act as the earliest point in time where the video id would be useful.

oSumAtrIX (changes requested):

src/main/kotlin/app/revanced/patches/youtube/misc/fix/playback/SpoofSignaturePatch.kt
oSumAtrIX: Sorry, I am a little confused on the new commits. The only change that I had in mind was that the integrations field "videoId" is also set from the method that was previously used to retrieve the video id and pass to `spoofParameter`.

This way, the field would be set earlier in time, so that the patches such as the client spoof patch could use it.

src/main/kotlin/app/revanced/patches/youtube/misc/fix/playback/SpoofSignaturePatch.kt
```diff
@@ -38,29 +41,60 @@ object SpoofSignaturePatch : BytecodePatch(
...
+                        )
+                    )
                 )
-            }
-        } ?: throw ProtobufParameterBuilderFingerprint.exception
+            )
+        )
+
+        // Hook the player parameters.
+        VideoIdPatch.injectProtoBufferHook("$INTEGRATIONS_CLASS_DESCRIPTOR->spoofParameter(Ljava/lang/String;)Ljava/lang/String;")
```
oSumAtrIX, line 93: This does not make sense to me. Why does `VideoIdPatch` hook the player parameter? `VideoIdPatch` should only hook video ids

src/main/kotlin/app/revanced/patches/youtube/video/information/VideoInformationPatch.kt
```diff
@@ -111,9 +111,7 @@ object VideoInformationPatch : BytecodePatch(
         /*
          * Inject call for video id
          */
-        val videoIdMethodDescriptor = "$INTEGRATIONS_CLASS_DESCRIPTOR->setVideoId(Ljava/lang/String;)V"
-        VideoIdPatch.injectCall(videoIdMethodDescriptor)
-        VideoIdPatch.injectCallBackgroundPlay(videoIdMethodDescriptor)
```
oSumAtrIX, line 116: Can you explain why `injectCallBackgroundPlay` is not necessary here anymore?

src/main/kotlin/app/revanced/patches/youtube/misc/fix/playback/SpoofSignaturePatch.kt
oSumAtrIX: > The problem is the video id needs to be set before the protobuffer hook, and both of hooks are in the same method.

In this case a third patch can be used which both patches depend on. The third patch would manage this specific method. It could be called a generic name such as "PlayerResponseMethodHookPatch". The patch would offer two methods. One method to hook the video id string, and one method to hook the player parameter string. Then, both the video id patch as well as the spoof patch could make use of it. The video id patch would simply delegate its inject function to that of PlayerResponseMethodHookPatch and  the spoof patch would call the PlayerResponseMethodHookPatch to hook the player parameter. This way the single-responsibility OOP principle is retained.

oSumAtrIX (approved):

PR #3038, opened 2023-09-27
fix(YouTube - Video Id): Fix video id not showing the currently playing video

LisoUseInAIKyrios:
Fix for RYD shorts sometimes not showing dislikes, which was caused by the new video id hook.

[integration changes](https://github.com/ReVanced/revanced-integrations/pull/484)


src/main/kotlin/app/revanced/patches/youtube/video/playerresponse/PlayerResponseMethodHookPatch.kt
```diff
@@ -32,10 +33,16 @@ object PlayerResponseMethodHookPatch : BytecodePatch(
...
-            ++videoIdHookInsertIndex, "invoke-static {p$VIDEO_ID_PARAMETER}, $methodDescriptor"
+            videoIdHookInsertIndex++, "invoke-static {p$VIDEO_ID_PARAMETER}, $methodDescriptor"
         )
 
-        // TODO: Explain why any video id hook has to precede any protoBufferParameterHook.
+        /**
+         * Adjust the buffer hook insert index since this video id hook was added before it.
+         *
+         * This ensures all video id hooks always run before the buffer hooks,
+         * so if the buffer hook calls into [VideoInformationPatch]
```
oSumAtrIX, line 43: `PlayerResponseMethodHookPatch` is a standalone patch that does not depend on or know about `VideoInformationPatch`. As an example, where this can be a problem is when another patch in the future needs the opposite order. Instead this patch can offer a method to hook before or after the video id as an example solution.

oSumAtrIX: > I don't see how any future patch could need the opposite order

I understand, but from the design perspective, this patch should not behave in a certain way for a patch it has no relation with.

 > can add hooks for both the player response video ID and the buffer

When I was refactoring the code, I also wanted to do this but the buffer hook explicitly requires a return value which is why designing it this way would be odd.

oSumAtrIX: I have now refactored the code. The strategy is to collect a set of hooks differentiated by the types "VideoId", "BeforeVideoId" and "ProtoBufferParameter". On close the hooks are inserted at the beginning of the method so that the "BeforeVideoId" hooks are inserted before and "ProtoBufferParameter" hooks are inserted after the "VideoId" hooks.

Order of insertion is kept by inverting the individual set of hooks.

oSumAtrIX (approved):

PR #3042, opened 2023-09-28
fix(YouTube - Premium heading): Correct inverted logic

aliernfrog:
Fixes an inverted logic caused by the [last commit](https://github.com/ReVanced/revanced-patches/pull/3029/commits/5f2f9676017f524d8315f095b18f1e74e85603f6) of #3029.

### DIFF
--- a/src/main/kotlin/app/revanced/patches/youtube/layout/branding/header/PremiumHeadingPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/youtube/layout/branding/header/PremiumHeadingPatch.kt
@@ -32,9 +32,9 @@ object PremiumHeadingPatch : ResourcePatch() {
         val resDirectory = context["res"]
 
         val (original, replacement) = if (usePremiumHeading!!)
-            DEFAULT_HEADING_RES to PREMIUM_HEADING_RES
-        else
             PREMIUM_HEADING_RES to DEFAULT_HEADING_RES
+        else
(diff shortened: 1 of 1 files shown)


src/main/kotlin/app/revanced/patches/youtube/layout/branding/header/PremiumHeadingPatch.kt
```diff
@@ -32,9 +32,9 @@ object PremiumHeadingPatch : ResourcePatch() {
         val resDirectory = context["res"]
 
         val (original, replacement) = if (usePremiumHeading!!)
-            DEFAULT_HEADING_RES to PREMIUM_HEADING_RES
-        else
             PREMIUM_HEADING_RES to DEFAULT_HEADING_RES
+        else
+            DEFAULT_HEADING_RES to PREMIUM_HEADING_RES
```
oSumAtrIX, line 37: I do not quite understand, if usePremiumHeading is false, then it should be premium (original) -> default (replacement), but this PR inverts this logic?

aliernfrog: `original` is the source resource, and `replacement` is the one getting replaced. Maybe those can be renamed to `source` and `target` to avoid confusions?

oSumAtrIX (approved):

PR #3061, opened 2023-10-02
feat(YouTube - Return YouTube Dislike): Support version `18.37.36`

oSumAtrIX:
# About

This PR brings back support of Return YouTube Dislike for version  18.37.36.
It fixes the issue with the patch showing the wrong dislikes for Shorts by using a different hook: https://github.com/ReVanced/revanced-patches/commit/e5be42bba237ffcde5fc57b04981b715a7a97c6b.

Logs showed, that this hook is being called with a video id, every time a Short's `TextComponent` was hooked.

# Issues that may occur

Because these are two separate hooks, it is unclear, if synchronization issues could occur. It is also unclear how this hook affects the current patch in detail.


oSumAtrIX:
Thanks, informing sounds sensible as clearing the data would be destructive and not worth a simple restart. Instead of clearing the data, why not force a restart?

PR #3071, opened 2023-10-04
feat(YouTube): Bump compatibility to `18.38.44`

LisoUseInAIKyrios:
Only a single fingerprint needed updating (and it still works with `18.37.36`)


src/main/kotlin/app/revanced/patches/youtube/layout/hide/shorts/fingerprints/BottomNavigationBarFingerprint.kt
```diff
@@ -1,18 +1,24 @@
...
 
 object BottomNavigationBarFingerprint : MethodFingerprint(
+    returnType = "V",
+    accessFlags = AccessFlags.PUBLIC or AccessFlags.FINAL,
+    parameters = listOf("Landroid/view/View;", "Landroid/os/Bundle;"),
     opcodes = listOf(
-        Opcode.MOVE_RESULT_OBJECT, // Refers to bottom navigation bar
+        Opcode.CONST, // R.id.app_engagement_panel_wrapper
+        Opcode.INVOKE_VIRTUAL,
+        Opcode.MOVE_RESULT_OBJECT,
```
oSumAtrIX, line 15: I assume this breaks compatibility with old versions, so it is important to not forget dropping support for it

PR #3098, opened 2023-10-06
fix(YouTube - Hide get YouTube Premium advertisements): move patch to `Hide ads`

LisoUseInAIKyrios:
Patch is already part of `Hide ads`, but during the recent refactoring it got a name added by mistake.

Changed "advertisement" to promo, to better indicates it's not advertisements Premium users will see, but instead a promotion to signup for YT Premium.

### DIFF
--- a/src/main/kotlin/app/revanced/patches/youtube/ad/getpremium/HideGetPremiumPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/youtube/ad/getpremium/HideGetPremiumPatch.kt
@@ -15,7 +15,7 @@ import app.revanced.patches.youtube.misc.settings.SettingsPatch
 import com.android.tools.smali.dexlib2.iface.instruction.TwoRegisterInstruction
 
 @Patch(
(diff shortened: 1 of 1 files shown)


src/main/kotlin/app/revanced/patches/youtube/ad/getpremium/HideGetPremiumPatch.kt
```diff
@@ -15,8 +15,7 @@ import app.revanced.patches.youtube.misc.settings.SettingsPatch
 import com.android.tools.smali.dexlib2.iface.instruction.TwoRegisterInstruction
 
 @Patch(
-    name = "Hide \"Get YouTube Premium\" advertisements",
-    description = "Hides YouTube Premium advertisements under video player.",
+    description = "Hides YouTube Premium signup promotion under the video player.",
```
oSumAtrIX, line 18:
```suggestion
    description = "Hides YouTube Premium signup promotions under the video player.",
```

PR #3116, opened 2023-10-08
feat: Add `Spoof SIM country` patch

scria1000:
At the request of governments in certain countries, particularly the one sharing a name with the bird traditionally consumed on Thanksgiving Day, Google has been imposing a countrywide restriction on satellite imagery quality for users accessing the app from those regions for more than a decade. This patch does what a VPN can't; however, it won't reveal any secret military bases.

~~I'm unsure about the patch name, I don't want it to give the impression that it can de-blur specific locations such as airports, military bases, or re-education camps, which it can't. It can only evade territory-wide blurring of certain countries.~~

| Before| After|
| -------- | ------- |
|||


oSumAtrIX:
> This patch does what a VPN can't

Why?

oSumAtrIX:
Since this patch modifies Google Maps, installing it on devices is impossible if it is preinstalled due to signature mismatch. I suppose this is root only.

scria1000:
> Does this require root?

Will require root for patching if it's a split APK.
Patched APK won't install if it came bundled as a system app, due to signature mismatch.
Otherwise no (at least on my LineageOS). Perhaps a microG integration is in order?

src/main/kotlin/app/revanced/patches/googlemaps/censorship/UncensoredSatelliteViewPatch.kt
```diff
@@ -0,0 +1,45 @@
...
+        classDef: ClassDef, method: Method, instruction: Instruction, instructionIndex: Int
+    ): Int? {
+        if (instruction.opcode != Opcode.INVOKE_VIRTUAL) return null
+        if (instruction !is ReferenceInstruction) return null
+        val reference = instruction.reference as? MethodReference ?: return null
+
+        if (reference.definingClass != "Landroid/telephony/TelephonyManager;") return null
+        if (reference.name != "getSimCountryIso" && reference.name != "getNetworkCountryIso") return null
+
+        return instructionIndex
```
oSumAtrIX, line 34: This is very similar to another codeblock, maybe extract this into a function?

oSumAtrIX (changes requested):

scria1000:
> > This patch does what a VPN can't
> 
> Why?

Speaking for the app; it doesn't take your IP address location into account when deciding whether to display the censored map. It appears to rely solely on info returned by telephony services.

src/main/kotlin/app/revanced/patches/all/telephony/sim/spoof/SpoofSimCountryPatch.kt
```diff
@@ -0,0 +1,76 @@
...
+
+@Patch(
+    name = "Spoof SIM country",
+    description = "Spoofs country information returned by the SIM card provider.",
+    use = false,
+)
+@Suppress("unused")
+object SpoofSimCountryPatch : AbstractTransformInstructionsPatch<Pair<Int, String>>() {
+    private val networkCountryIso by stringPatchOption(
+        "getNetworkCountryIso",
```
oSumAtrIX, line 27: The option key should not be abused to check for the reference name in `filterMap`.

src/main/kotlin/app/revanced/patches/all/telephony/sim/spoof/SpoofSimCountryPatch.kt
```diff
@@ -0,0 +1,76 @@
...
+        entry: Pair<Int, String>
+    ) = transformMethodCall(entry, mutableMethod)
+
+    private fun transformMethodCall(
+        entry: Pair<Int, String>,
+        mutableMethod: MutableMethod
+    ) {
+        val (instructionIndex, methodCallValue) = entry
+
+        // Get register of the next MOVE_RESULT_OBJECT instruction that accompanies INVOKE_VIRTUAL.
```
oSumAtrIX, line 69: This commend is a duplicate of the code below.

src/main/kotlin/app/revanced/patches/all/telephony/sim/spoof/SpoofSimCountryPatch.kt
```diff
@@ -0,0 +1,76 @@
...
+        val instruction = mutableMethod.getInstruction<BuilderInstruction11x>(instructionIndex + 1)
```
oSumAtrIX, line 70: MOVE_RESULT_OBJECT is a OneRegisterInstruction, you should use that as the generic type parameter.

src/main/kotlin/app/revanced/patches/all/telephony/sim/spoof/SpoofSimCountryPatch.kt
```diff
@@ -0,0 +1,76 @@
...
+        val register = instruction.registerA
+
+        mutableMethod.removeInstruction(instructionIndex + 1)
+        mutableMethod.replaceInstruction(instructionIndex, "const-string v$register, \"$methodCallValue\"")
```
oSumAtrIX, line 74: You can simply replace the MOVE_RESULT_OBJECT instruction for simplicity.

src/main/kotlin/app/revanced/patches/all/telephony/sim/spoof/SpoofSimCountryPatch.kt
```diff
@@ -0,0 +1,76 @@
...
+        instruction: Instruction,
+        instructionIndex: Int
+    ): Pair<Int, String>? {
+        if (instruction !is ReferenceInstruction) return null
+
+        val reference = instruction.reference as? MethodReference ?: return null
+        if (reference.definingClass != "Landroid/telephony/TelephonyManager;") return null
+
+        if (!options.contains(reference.name)) return null
+        return (options[reference.name].value as? String)?.let { instructionIndex to it.trim() }
```
oSumAtrIX, line 55: This is not ideal. You should not use the option key for this purpose. Create an enum; if the reference matches, then you get the replacement value from the option.

oSumAtrIX (changes requested):

scria1000:
> It would be nice if the patch logged if no changes were made (or alternatively log how many byte code locations were changed), just for the sake of trying this on random apps and to know if the app checks the sim code.
> 
> But currently there is no way to turn on verbose logging from CLI patcher.

Like this?

 

Yeah, a verbosity setting would be nice.

oSumAtrIX:
`logger.fine` can be added so that the logging level decreased to display them via environment variables, although offtopic to this issue for now.

oSumAtrIX (changes requested):
Looks good to me. Please look into https://github.com/ReVanced/revanced-patches/pull/3116#discussion_r1357232789.

src/main/kotlin/app/revanced/patches/all/telephony/sim/spoof/SpoofSimCountryPatch.kt
scria1000: I need input on this, how else would you do it:

1.
```
    private val networkCountryIso by stringPatchOption()

    private val simCountryIso by stringPatchOption()

    override fun filterMap(
        /* ... */
    ): Pair<Int, String>? {
        /* ... */

        val methodMatch = MethodCall.entries.firstOrNull {
            /* ... */
        }

        val optionValue = options[methodMatch.optionKey].value as? String ?: return null
        return instructionIndex to optionValue.lowercase()
    }
 
    private enum class MethodCall(
        /* ... */
        val optionKey: String
    ) {
        NetworkCountryIso(
            "networkCountryIso"
        ),
        SimCountryIso(
            "simCountryIso"
        )
    }

```

2.
```
    private val networkCountryIso by stringPatchOption()

    private val simCountryIso by stringPatchOption()

    override fun filterMap(
        /* ... */
    ): Pair<Int, String>? {
        /* ... */

        val methodMatch = MethodCall.entries.firstOrNull {
            /* ... */
        }

        return methodMatch.optionValue()?.let { instructionIndex to it.lowercase() }
    }
 
    // Lambda because we don't want the previous option value to persist if the patch is cancelled mid-execution
    private enum class MethodCall(
        /* ... */
        val optionValue: () -> String?
    ) {
        NetworkCountryIso(
            optionValue = { networkCountryIso }
        ),
        SimCountryIso(
            optionValue = { simCountryIso }
        )
    }

```

oSumAtrIX: `MethodCall` should be an enum of methods you want to transform and not contain the options key. See 
https://github.com/revanced/revanced-patches/blob/aeb5299ca50cdeef1c6db273d312b2afea82af7b/src/main/kotlin/app/revanced/patches/all/connectivity/wifi/spoof/SpoofWifiPatch.kt#L44-L47

Then, you can check if the current instruction matches the MethodCall enum and replace it with the value of the respective option.

scria1000:
```kotlin

        val optionValue = when(methodMatch) {
            MethodCall.NetworkCountryIso -> networkCountryIso
            MethodCall.SimCountryIso -> simCountryIso
        }
        return optionValue?.let { instructionIndex to it.lowercase() }

``` 

Is this fine?

oSumAtrIX: Yeah, works:

```kotlin

        val iso = when(methodMatch) {
            MethodCall.NetworkCountryIso -> networkCountryIso
            MethodCall.SimCountryIso -> simCountryIso
        }.toLowerCase()

        return optionValue?.let { instructionIndex to iso }

```

scria1000: Thank you for the suggestions.

oSumAtrIX: Whoops the null check on optionValue may not be correct

oSumAtrIX:
I have made some refactoring changes, please check if they look good to you.

oSumAtrIX (approved):

src/main/kotlin/app/revanced/patches/all/telephony/sim/spoof/SpoofSimCountryPatch.kt
```diff
@@ -0,0 +1,111 @@
...
+            "getNetworkCountryIso",
+                emptyList(),
+            "Ljava/lang/String;"
+            )
+        ),
+        SimCountryIso(
+            ImmutableMethodReference(
+            "Landroid/telephony/TelephonyManager;",
+            "getSimCountryIso",
+                emptyList(),
```
oSumAtrIX, line 106:
```suggestion
            emptyList(),
            "Ljava/lang/String;"
            )
        ),
        SimCountryIso(
            ImmutableMethodReference(
            "Landroid/telephony/TelephonyManager;",
            "getSimCountryIso",
            emptyList(),
```

scria1000:
It's great, thanks.

oSumAtrIX:
Have you tested it? I can not test because there is no censoring for me.

scria1000:
Yes, the patch still works fine (Btw, the app immediately returns to censored state when it is unpatched so there doesn't seem to caching involved).

nightznero:
could somebody explain how to use `Network ISO Country Code` cuz adding 311160 (for TMobile for example) doesnt work but `SIM ISO' With any code of country work flawless but Network ISO doesnt,am i missing something? Or just both option cannot be enabled?

oSumAtrIX:
Open a detailed issue

PR #3147, opened 2023-10-15
feat(FX File Explorer): Add `Unlock Plus` patch

ungeskriptet:
Add a patch that makes FX File Explorer think the unlock key app is installed.


ungeskriptet:
@LisoUseInAIKyrios Ah sorry, no I did not. So is the patch not allowed because it goes under the 'Payment circumvention' category or are you saying that I should change the main branch to dev?

oSumAtrIX:
Because we are focusing on decentralizing patches, do not delete it as in the future you will be able to host and distribute it conveniently yourself.

PR #3196, opened 2023-10-25
chore(id-austria): Set fixed version for ID Austria

1fexd:
The new version `3.0.0` is not supported yet. Not sure when I will have time to check it out, therefore I marked only `2.7.1` as compatible.


oSumAtrIX (approved):

1fexd:
Thanks :)

PR #3219, opened 2023-11-02
feat(lightroom): Constrain compatibility to last working version

taku-nm:
Patching older versions of Lightroom than the one I am constraining to, doesn't cause the patch itself to fail, however it does not unlock any of the premium functions, therefore making the patch itself useless. Hence the constraint to `9.0.0`


oSumAtrIX:
This would suggest against patching newer versions, older versions are not to be considered. Using no constraint should be treated as working on now and newer.

PR #1030, opened 2022-11-13
feat(twitch/show-deleted-messages): Show deleted livestream chat messages

timschneeb:
This patch enables a feature that is normally only accessible to Twitch channel moderators.
It allows all viewser to read deleted chat messages (hidden behind a clickable spoiler).

PR #3246, opened 2023-11-09
feat(Twitch): Support version `17.0.0`

LizenzFass78851:
Continued from #3243

Now supports version `17.0.0` of Twitch.

- the ad patches and “Auto Claim Channel Points” work.
- "shop deleted messages" As I said, I can't test it, I tried but I couldn't test it myself. But the patch is included in this pr

I ask that a maintainer counter-test this because I actually have no idea about Java.
It's better if someone counter-tests it so that there aren't any unidentified errors that I couldn't find

### DIFF
--- a/src/main/kotlin/app/revanced/patches/twitch/ad/audio/AudioAdsPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/twitch/ad/audio/AudioAdsPatch.kt
@@ -17,7 +17,7 @@ import app.revanced.patches.twitch.misc.settings.SettingsPatch
     name = "Block audio ads",
     description = "Blocks audio ads in streams and VODs.",
     dependencies = [IntegrationsPatch::class, SettingsPatch::class],
-    compatiblePackages = [CompatiblePackage("tv.twitch.android.app", ["15.4.1", "16.1.0"])],
+    compatiblePackages = [CompatiblePackage("tv.twitch.android.app", ["15.4.1", "16.1.0", "17.0.0"])],
 )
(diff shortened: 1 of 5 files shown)


oSumAtrIX:
Any reason the PR has been closed?

LizenzFass78851:
> Any reason the PR has been closed?

because I don't know whether the pull request will be merged. I am aware that it may take some time until it is merged. I will open a new pull request for this because I can no longer open the previous pull request. As mentioned, the promotional patches work but unfortunately I can't test beyond that.
Even if it looks like I've added a new version number of Twich to a Kotlin code, that doesn't mean that I'm a full-fledged Java developer and if the patches don't work properly due to an unnoticed error, I can don't fix it.

LizenzFass78851:
> You said everything works except you haven't verified show deleted messages. Since all other patches work then the deleted messages patch also probably still works.
> 
> You can try asking someone in Twitch chat to delete a message and they'll probably comply.

I tested this with the "show deleted messages".
The following result:
Screenshot_20231111-093831


Is the result correct enough to ensure that the "show deleted messages" patch works?


-----


As mentioned, I could no longer open this pull request so I had to open a new pull request #3250.

LizenzFass78851:
@LisoUseInAIKyrios 

I did some research and according to the pull request #1030 that was introduced it says word verbatim:

> This patch enables a feature that is normally only accessible to Twitch channel moderators.
> **It allows all viewser to read deleted chat messages (hidden behind a clickable spoiler).**

According to the screenshot and the description mentioned, the patch should work

PR #3345, opened 2023-11-29
feat(YouTube): Support version `18.45.43`

LisoUseInAIKyrios:
- added version `18.45.43`
- removed very old targets that nobody should be using.  Kept `18.29.38` as the minimum, as that's the minimum supported by `Hide layout components`

The remove old targets could be omitted from this PR, but I think it keeps the website cleaner when users view the supported versions (nobody should be patching `18.16.37`)

~~On another note, it appears the Shorts player is back to using the old non litho code.  At least it is for me.~~ Edit: Version spoofing was accidentally left on.  Shorts player is still using litho.


oSumAtrIX:
> On another note, it appears the Shorts player is back to using the old non litho code

I was not able to reproduce this. On my end, the new player is being used. Have you considered A/B tests considered?

PR #3356, opened 2023-12-02
feat: Publish as library

oSumAtrIX:
# About

This PR focuses on allowing patches from other future bundles to interface with each other. For that this PR fixes visibility modifiers as well as prepares the project to be published as a library to link to.


Aunali321:
Alright

PR #3377, opened 2023-12-05
fix(Twitch - Settings): Support missing version `16.1.0` and `15.4.1`

LizenzFass78851:
This will add back the missing versions for Twitch Settings Patch.

Tested successfully with the Twitch version `16.1.0` and `15.4.1`

### DIFF
--- a/src/main/kotlin/app/revanced/patches/twitch/misc/settings/SettingsPatch.kt
+++ b/src/main/kotlin/app/revanced/patches/twitch/misc/settings/SettingsPatch.kt
@@ -30,7 +30,7 @@ import java.io.Closeable
     description = "Adds settings menu to Twitch.",
     dependencies = [IntegrationsPatch::class, SettingsResourcePatch::class],
     compatiblePackages = [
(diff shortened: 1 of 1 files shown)


oSumAtrIX:
What point does it have?

LizenzFass78851:
> What point does it have?

When it comes to Twitch patches, version `16.9.1` `16.1.0` and `15.4.1` are supported for all advertising and chat patches.

In the settings patch, a past change set this to version `16.9.1` as this is the last version with which the settings patch works.

As mentioned above, version `16.1.0` and `15.4.1` are also supported for the promotional patches and the settings patch should also be supported for the versions mentioned because I also think it works on the versions mentioned.

I also counter-tested it

oSumAtrIX:
Patches work with the versions as advertised, but may not work when not. This does not mean they won't. Patches are separate of each other, so adding a version to one for the sake of another patch supporting it is not the correct approach. Even in the case, that multiple patches support distinct versions only. In such a case you'd chose either version and either not use one patch or force patch on that version. The versions are not missing as assumed in the PR title, the versions simply were not necessary for the Settings patch. Coincidentally the patches support one common so you can use all patches as advertised on that version

PR #3378, opened 2023-12-05
feat(YouTube): add DeArrow option for `Alternative Thumbnails` Patch

shadow578:
adds the option to use thumbnails provided by [DeArrow](https://dearrow.ajay.app/) to the existing alternative thumbnails patch


oSumAtrIX:
Run the Gradle task API dump. The task is ensuring the public API ist being changed unintentionally as soon other patch repositories can use it. Breaking changes should be avoided by deprecation. If the API check fails, it may indicate a breaking change, but not necessarily. In this case you just dump the API and commit it with the change. If it is a breaking change you commit it as one using the `BREAKING CHANGE` footer.

oSumAtrIX (approved):

PR #3466, opened 2025-07-02
ci: Use default fetch-depth of 1 for checkout step

validcube:
Why fetch *all* commit history when we are not going to use them. 

##### To be updated, but for now--for my own information--in case I forgot it: https://trello.com/c/xWPgKGaO/7-revanced-ci-template-01-07-2025


oSumAtrIX (approved):

oSumAtrIX:
npm install should be changed the same way it was changed in the counterfeits repo

